PR #1, opened 2024-02-09
Add support for ECC and fix some issues along the way

Jakuje:


build.rs
```diff
@@ -102,6 +102,15 @@ fn build_fips() {
...
         .allowlist_item("BN_.*")
+        .allowlist_item("d2i_ECPKParameters")
+        .allowlist_item("d2i_ECDSA_SIG")
+        .allowlist_item("i2d_ECDSA_SIG")
+        .allowlist_item("d2i_ASN1_OCTET_STRING")
+        .allowlist_item("i2d_ASN1_OCTET_STRING")
+        .allowlist_item("EC_GROUP_.*")
+        .allowlist_item("ECDSA_SIG_.*")
+        .allowlist_item("ASN1_STRING_.*")
+        .allowlist_item("NID_undef")
```
simo5, line 113: You can probably allowlist "d2i_.\*" "i2d_.\*", is NID_undef really needed?

src/ossl/ecc.rs
```diff
@@ -0,0 +1,905 @@
...
+        }
+    }}
+}
+
+macro_rules! make_ec_param_ossl_param_utf8_string {
+    ($name:expr; $vin:expr; $vout:expr) => {{
+        let ec_group = make_ec_group_from_ec_params!($vin);
+        if ec_group.is_null() {
+            unsafe { EC_GROUP_free(ec_group); }
+            return err_rv!(CKR_DEVICE_ERROR)
```
simo5, line 89: should probably be CKR_GENERAL_ERROR

src/ossl/ecc.rs
```diff
@@ -0,0 +1,905 @@
...
+}
+
+macro_rules! param_bn_to_obj_attr {
+    ($params:expr; $name:expr; $obj:expr; $attr:expr) => {{
+        let mut vec = Vec::<u8>::new();
+        param_bn_to_vec!($params; $name; vec);
+        $obj.set_attr(attribute::from_bytes($attr, vec))?;
+    }};
+}
+
```
simo5, line 381: This is a comment for all of the above:
Looks like a ton of macro rules here, any reason why these can't be just regular functions?
Debugging with macro_rules is extremely painful, so they should be used only when really needed, and for very small compact actions that will unlikely need to be debugged.

build.rs
Jakuje: Not necessarily. It can be removed and I can use just the number 0 or something.

README.md
```diff
@@ -1,4 +1,27 @@
...
+# Setup
+
+First after cloning, we need to pull and update openssl submodule:
+
+$ git submodule init
+$ git submodule update
+
+Build the rust project:
+
+$ cargo build
```
simo5, line 17: we may want to specify that --features fips is needed if the fips version is to be built ?

src/ecc.rs
```diff
@@ -0,0 +1,425 @@
...
+    */
+}
+
+static PUBLIC_KEY_FACTORY: Lazy<Box<dyn ObjectFactory>> =
+    Lazy::new(|| Box::new(ECCPubFactory::new()));
+
+static PRIVATE_KEY_FACTORY: Lazy<Box<dyn ObjectFactory>> =
+    Lazy::new(|| Box::new(ECCPrivFactory::new()));
+
+fn check_key_object(key: &Object, public: bool, op: CK_ULONG) -> KResult<()> {
```
simo5, line 219: should I make this stuff available in a util module so you can simply reuse it here instead of copy/paste?

Jakuje: Yes, if we add a key type argument, I think it should do the generic enough for all key types we will need.

README.md
Jakuje: I did not touch it after we talked this week that it might make sense to use FIPS by default. If we will not go that direction, the fips feature should be indeed separately documented.

src/ecc.rs
simo5: I pushed a patch that provides an Object function for that.

Jakuje: Rebased and resolved.

Jakuje:
Rebased . I think the last remaining issue with the FIPS version is that the ECDSA siganure in DER format (as returned by OpenSSL does not have strict length so we need to get the length from the FIPS submodule somehow (see the top commit). The current version does not build, but I hope I will be able to tackle this somehow. Insights welcomed.

src/fips.rs
```diff
@@ -752,12 +752,19 @@ impl ProviderSignatureCtx {
...
                         siglen_ptr,
                         signature.len()
-                    ))
+                    );
+                    if res != 1 {
+                        err_rv!(CKR_DEVICE_ERROR);
+                    }
+                    /* update to match the real size -- for example ECDSA signatures may vary in
+                     * length */
+                    signature.resize(siglen, 0);
```
simo5, line 766: unsafe { signature.set_len(siglen) }; ?

simo5: ah we pass in a slice ... I guess we may have to pass in a vector or separate CK_BYTE_PTR and CK_ULONG_PTR

Jakuje: Resolved somehow. But there might be better way.

Jakuje:
Pushed the merge of the fips/ossl code into one file (should be probably squashed later). Keeping now for review. Tests pass. There are couple of todos and there might be some things to fix before merging so let me know if you have something in mind.

simo5 (changes requested):
Thanks Jakub,
this looks rally good and close to the finishing line.

I left a few comments and just some minor tweaks requests.

src/ecc.rs
```diff
@@ -0,0 +1,392 @@
...
+        data
+    }
+}
+
+impl ObjectFactory for ECCPubFactory {
+    fn create(&self, template: &[CK_ATTRIBUTE]) -> KResult<Object> {
+        let obj = self.default_object_create(template)?;
+
+        bytes_attr_not_empty!(obj; CKA_EC_PARAMS);
+        bytes_attr_not_empty!(obj; CKA_EC_POINT);
```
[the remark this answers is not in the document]
Jakuje: Either of these would require parsing the ec params and also the ec point using the asn1 library, which is now done in the `src/ossl/ecc.c`. I think it can be moved, but OpenSSL does this anyway when the key is imported. The only advantage might be that the debugging might  be easier for bad keys, but we would have to reimplement quite some stuff from openssl.
Checking the length of point based on the parameters might make sense though.

src/fips.rs
```diff
@@ -746,18 +746,29 @@ impl ProviderSignatureCtx {
         }
     }
 
-    pub fn digest_sign_final(&mut self, signature: &mut [u8]) -> KResult<()> {
+    pub fn digest_sign_final(
+        &mut self,
+        signature: &mut [u8],
+        signature_len: &mut Option<&mut usize>,
```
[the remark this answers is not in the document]
Jakuje: Right. This was C-coder attempt to resolve the issue :) Using return value is much better. Done.

src/ossl/ecc.rs
```diff
@@ -0,0 +1,953 @@
...
+            slice::from_raw_parts(
+                $name.as_ptr() as *const std::os::raw::c_char,
+                $name.len(),
+            )
+            .to_vec()
+        }
+    };
+}
+
+fn get_digest_name(
```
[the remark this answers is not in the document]
Jakuje: Yeah, that would make sense. I will see if I can reorg these to common

src/ossl/ecc.rs
```diff
@@ -0,0 +1,953 @@
...
+            OSSL_PKEY_PARAM_PUB_KEY as *const u8 as *const i8,
+        )
+    };
+    if p.is_null() {
+        return err_rv!(CKR_DEVICE_ERROR);
+    }
+    /* get length */
+    let mut str_len = 0;
+    let res = unsafe {
+        OSSL_PARAM_get_octet_string(
```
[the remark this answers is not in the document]
Jakuje: Thanks for suggestion. Changed.

src/ossl/ecc.rs
```diff
@@ -0,0 +1,953 @@
...
+    let mut s_bytes = sig.s.as_bytes();
+    // skip leading zero
+    if s_bytes[0] == 0 {
+        s_bytes = &s_bytes[1..];
+    }
+    signature[bn_len..(bn_len + s_bytes.len())].copy_from_slice(s_bytes);
+    /* pad with zeroes */
+    if s_bytes.len() < bn_len {
+        /* pad with zeroes */
+        signature[(bn_len + s_bytes.len())..(2 * bn_len)].fill(0);
```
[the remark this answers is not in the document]
Jakuje: Applied. Thanks

src/ossl/ecc.rs
```diff
@@ -0,0 +1,953 @@
...
+    let s_buf = &signature[bn_len..];
+    let mut s_vec = Vec::with_capacity(bn_len);
+    // if the buffer starts with bit 1, it would be interpreted as negative integer.
+    // Prepend one more null byte.
+    if s_buf[0] & 0x80 != 0 {
+        s_vec.push(0);
+    }
+    s_vec.extend_from_slice(s_buf);
+    let sig = EcdsaSignature {
+        r: asn1::BigUint::new(&r_vec.as_slice()).expect("bad r value"),
```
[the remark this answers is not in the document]
Jakuje: Added a comment as it was not completely clear to me how to make a use of these.

src/ossl/ecc.rs
```diff
@@ -0,0 +1,953 @@
...
+        Ok(NAME_SECP521R1)
+    } else {
+        err_rv!(CKR_GENERAL_ERROR)
+    }
+}
+
+fn oid_to_bits(
+    oid: asn1::ObjectIdentifier,
+) -> KResult<usize> {
+    // the match syntax does not allow me to construct OIDs on left side
```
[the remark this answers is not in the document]
Jakuje: I did not address this one as it is not completely clear how I should approach this. You mean the lazy_static crate to initialize these?

src/ecc.rs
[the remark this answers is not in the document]
simo5: If the checks are already made elsewhere I am fine, but I am concerned about importing a key that will never work later.
That said I also do very little checking for RSA, so we can probably just defer better checking to later.

I do not have an issue with generate because that one is self checked in openssl, but create is just "import", unfortunately PKCS#11 calls import "create" which can be misleading.

src/ossl/ecc.rs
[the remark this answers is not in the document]
simo5: yes just like it is done for the FACTORY static variables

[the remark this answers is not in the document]
simo5: Just to be clear, We do nto need to do it now, I can refactor this later

src/ossl/ecc.rs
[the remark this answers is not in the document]
simo5: I realized there is unsafe behavior with some of these macros that leaks through.
So leave them as copies for now, I will commit some more robust code in ossl/common.rs shortly, and then we can refactor the code after your PR is merged.

Jakuje:
Rebased again to use the new OsslParam API.

simo5 (approved):

simo5:
LGTM!

PR #4, opened 2024-02-27
Fix CI and add codespell checks

simo5:
Add codespell


simo5:
what do you think about this CI jobs for starters?

simo5:
Pretty happy about this start, and given it found some code fixes that would probably get in the way of pushing code later, I am going to merge this one w/o review, nothing earth shattering anyway, and anything can be easily changed in a future PR.

PR #5, opened 2024-02-28
Support for wrapping EC keys + RSA wrapping fixes/comments/questions

Jakuje:
Fixes also unwrapping for RSA keys, but we might need better solution.

I think the `DerEncBigUint` ensures the BN is unsigned and starts with zero byte, which is something we do not want when we import the BN as a modulus as you are using the modulus size to enforce the buffer size. Without stripping the null byte from the modulus, the operation requires 257 B buffer for signature and verification of this signature does not work.

The RSA wrapping is also using the `RSAPrivateKey` ASN1 directly, but from my reading of the PKCS#11 specification, we should use `PrivateKeyInfo`. This not obvious for RSA keys as everything needed is in the inner object, but EC keys needs the `privateKeyAlgorithm` field to learn what curve is used, if I read it right. 

Keeping it as a draft as there is more quesitons to resolve before merging.

### DIFF
--- a/src/ecc.rs
+++ b/src/ecc.rs
@@ -4,6 +4,7 @@
 use super::attribute;
 use super::error;
 use super::interface;
+use super::kasn1;
 use super::object;
 use super::{attr_element, bytes_attr_not_empty, err_rv};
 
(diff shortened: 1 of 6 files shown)


src/rsa.rs
```diff
@@ -238,9 +238,19 @@ impl PrivKeyFactory for RSAPrivFactory {
...
+        /* Skip leading zeroes as they are not significant in the BN */
+        let modulus = rsapkey.modulus.as_bytes();
+        let mut skip = 0;
+        for i in 0..modulus.len() {
+            if modulus[i] != 0 {
+                break;
+            }
+            skip += 1;
+        }
+        let modulus = &modulus[skip..];
```
simo5, line 252: I think this should be code in DerEncBigUint, in a function named something like: as_nopad_bytes() or similar. We do not want to repeat this code in multiple places.

src/ecc.rs
```diff
@@ -62,6 +144,66 @@ impl CommonKeyFactory for ECCPubFactory {}
...
+}
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+struct PrivateKeyInfo<'a> {
+    version: Version,
+    private_key_algorithm: asn1::ObjectIdentifier,
+    // private_key: &'a [u8],
+    private_key: asn1::OctetStringEncoded<ECPrivateKey<'a>>,
+    // likely not needed
+    //#[explicit(1)]
```
simo5, line 187: needed if we ever want to parse data coming from outside our module.

simo5: we can simply ignore the content of course, but the asn.1 structure needs to be complete or the parser will just fail

src/ecc.rs
```diff
@@ -62,6 +144,66 @@ impl CommonKeyFactory for ECCPubFactory {}
...
+    // ecParametdders   ECParameters,
+    OId(asn1::ObjectIdentifier),
+    // implicitlyCA asn1::Null,
+    CurveName(asn1::PrintableString<'a>),
+}
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+struct ECPrivateKey<'a> {
+    version: Version,
+    private_key: &'a [u8],
```
simo5, line 160: is this not an INTEGER in ASN.1 ?
can you add comment that references the canonical ASN.1 definition (RC or whatever doc).

simo5: having a Cow using abstraction here like DerEncBigUint, would avoid the need for the new_owned() functions and save memcopies, and allow us to properly zeroize memory

src/rsa.rs
```diff
@@ -89,6 +89,8 @@ struct OtherPrimeInfo<'a> {
     coefficient: DerEncBigUint<'a>,
 }
 
+// TODO Is this correct? According to the PKCS#11 specs, for wrapping we should use
+// BER-encoded according to PKCS #8’s PrivateKeyInfo ASN.1 type
```
simo5, line 93: DER is a strict subset of BER encoding, so when we produce the PrivateKeyInfo we are BER compliant.
Unfortunately we may fail to parse a structure that actually uses BER encodings that are not conformant to the stricter DER, however I am not aware of any implementation that uses non-DER for encoding this structure. If that happens and we need to become compatible we'll have to figure out then what we want to do.

simo5 (commented):
I left some suggestions/comments, the general changes look good to me.

src/rsa.rs
Jakuje: the point here is the PrivateKeyInfo x RSAPrivateKey

src/rsa.rs
Jakuje: Indeed, this was mostly PoC to get the test working for now. Will do something like that.

src/rsa.rs
simo5: https://www.ietf.org/rfc/rfc5208.html#section-5
I see now, this is incomplete

src/rsa.rs
Jakuje: Moved to the separate function. Done so far only for the modulus as it has obvious conseqences, but might be worth doing also for other parameters.

src/ecc.rs
Jakuje: This comes from https://www.secg.org/sec1-v2.pdf, section C.4  which is referenced by the PKCS#11 specs. It is also written this way in "2.7 Wrapping/unwrapping private keys" of PKCS#11 3.0 to be confusing enough. I can add references or original structures to the comments if needed.

I will check the CoW. We should zeroize this one indeed.

src/rsa.rs
simo5: Yeah, should probably be done for all of them I guess.

src/ecc.rs
simo5: Ok it is an OCTET STRING, so basically we need the same code as BigUint except for the Tag which will have to identify an Octet string instead of an INTEGER.
Of course with octet string there is no leading zeros/sign problem, so that helps keeping the code simple.

src/rsa.rs
Jakuje: OK. Will add.

src/ecc.rs
Jakuje: Added them, but hope we will not have to use them any time soon :)

src/rsa.rs
Jakuje: Tried to wrap the RSA with the PrivateKeyInfo structure, but it needs some more sharing between algorithms, which I did not do yet. And integration testing, which will be possible likely as soon as we will be able to use the module without the pReserved pointer. Resolving for now.

src/ecc.rs
Jakuje: Tried to do that. I am still slightly lost here so let me know if what I did makes sense or I am off ...

src/rsa.rs
```diff
@@ -129,6 +129,45 @@ impl RSAPrivateKey<'_> {
...
+    attribute_value: asn1::SetOf<'a, asn1::Tlv<'a>>, // ANY
+}
+
+type Attributes<'a> = asn1::SetOf<'a, Attribute<'a>>;
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+struct PrivateKeyInfo<'a> {
+    version: Version,
+    private_key_algorithm: asn1::ObjectIdentifier,
+    private_key: asn1::OctetStringEncoded<RSAPrivateKey<'a>>,
```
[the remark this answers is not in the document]
simo5: we will need to have an enum of OIDs plus specialized structures:

```
enum PrivateKey {
   #defined_by(OID_RSA_ENCRYPTION)]
   Rsa(asn1::OctetStringEncoded<RSAPrivateKey<'a>>),
   ...
}

```
then later on you assign:
`private_key: PrivateKey::Rsa(asn1::OctetStringEncoded::new(rsa_private_key))`

src/rsa.rs
```diff
@@ -129,6 +129,45 @@ impl RSAPrivateKey<'_> {
...
+    #[explicit(1)]
+    attributes: Option<Attributes<'a>>,
+}
+
+impl PrivateKeyInfo<'_> {
+    pub fn new_owned<'a>(
```
simo5, line 155: why new_owned() and not just new() ?

src/rsa.rs
```diff
@@ -129,6 +129,45 @@ impl RSAPrivateKey<'_> {
     }
 }
 
+const OID_RSA_ENCRYPTION: &[u8] = &[
+    0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x0D, 0x01, 0x01, 0x01,
+];
```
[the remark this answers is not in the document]
Jakuje: Thanks! Good idea. This sounds like it actually works. But I just need to use the numberic OID, not the encoded bits:
```
+const OID_SECP384R1: asn1::ObjectIdentifier = asn1::oid!(1, 3, 132, 0, 34);
```

src/kasn1.rs
```diff
@@ -40,6 +40,18 @@ impl<'a> DerEncBigUint<'a> {
     pub fn as_bytes(&self) -> &[u8] {
         &self.data
     }
+
+    /// Get the BN bytes without possible leading NULL bytes.
+    pub fn as_nopad_bytes(&self) -> &[u8] {
+        let mut skip = 0;
+        for i in 0..self.data.len() {
```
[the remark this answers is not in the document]
Jakuje: The rust complains here that the Cow is not an iterator with your proposed solution:
```
error[E0277]: `&Cow<'_, [u8]>` is not an iterator
  --> src/kasn1.rs:47:20
   |
47 |         for val in &self.data {
   |                    ^^^^^^^^^^ `&Cow<'_, [u8]>` is not an iterator
   |
   = help: the trait `Iterator` is not implemented for `&Cow<'_, [u8]>`
   = note: required for `&Cow<'_, [u8]>` to implement `IntoIterator`

For more information about this error, try `rustc --explain E0277`.
```

src/rsa.rs
Jakuje: changed.

simo5 (approved):

PR #7, opened 2024-02-29
Add SHA3 signature mechanisms for RSA and ECC

Jakuje:
should be straight-forward. But might be open to some simplification by adding this in a loop ... WDYT?


simo5:
good enough, in my rsa pss/oaep PR I have an example of how to loop, but let's get this merged now, and then we'll add the looping when we move the regisration following the RSA example, after we are happy with it.

simo5 (approved):
LGTM

PR #8, opened 2024-02-29
ci: Attempt to cache build to speed up CI

Jakuje:


Jakuje:
This sounds like does the job now:
 * The caches are created in https://github.com/latchset/kryoptic/actions/caches (~500MB for OpenSSL and ~100MB for Rust dependencies)
 * The build time gets to ~1 minute from ~6 minutes previously

The ugly thing is that we need to write the openssl cache only from the fips build (which is usable also in the non-fips one).

simo5 (approved):
A W E S O M E

PR #12, opened 2024-03-01
Share the PrivateKeyInfo structure among RSA and ECC implementation

### LINKED ISSUE #10: Use generic PrivateKeyInfo
Currently, there is separate copy of PrivateKeyInfo in RSA and ECC code, handling the wrapping and unwrapping of private keys.

Simo suggested that this would be something that could be used with `#defined_by(OID)` of rust-asn1 library to provide generic structure. The WIP branch implementing this is available here (but still has some memory ownership issues to resolve):

https://github.com/Jakuje/kryoptic/commits/defined-by/

              sounds like this is a place where to use:
#[defined_by(private_key_algorithm)]

See: https://docs.rs/asn1/latest/asn1/#defined-by

_Originally posted by @simo5 in https://github.com/latchset/kryoptic/pull/5#discussion_r1507927653_

Jakuje:
~Using the `defined-by` magic.~

Using opaque OCTET STRING

Fixes #10.

### DIFF
--- a/src/kasn1.rs
+++ b/src/kasn1.rs
@@ -123,3 +123,44 @@ impl<'a> asn1::SimpleAsn1Writable for DerEncOctetString<'a> {
         dest.push_slice(self.as_bytes())
     }
 }
+
+type Version = u64;
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+struct Attribute<'a> {
+    attribute_type: asn1::ObjectIdentifier,
+    attribute_value: asn1::SetOf<'a, asn1::Tlv<'a>>, // ANY
+}
+
+type Attributes<'a> = asn1::SetOf<'a, Attribute<'a>>;
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+pub struct PrivateKeyInfo<'a> {
+    version: Version,
+    private_key_algorithm: asn1::ObjectIdentifier,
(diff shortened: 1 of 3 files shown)


Jakuje:
Rewritten to use opaque OCTET STRING in the shared code as the DEFINED by would need us to share all algorithm structures. Should be ready.

src/kasn1.rs
```diff
@@ -123,3 +123,44 @@ impl<'a> asn1::SimpleAsn1Writable for DerEncOctetString<'a> {
...
+    version: Version,
+    private_key_algorithm: asn1::ObjectIdentifier,
+    private_key: DerEncOctetString<'a>,
+    #[explicit(1)]
+    attributes: Option<Attributes<'a>>,
+}
+
+impl PrivateKeyInfo<'_> {
+    pub fn new<'a>(
+        private_key_asn1: &'a Vec<u8>,
```
simo5, line 148: Why not pass directly a slice here?

simo5 (approved):
Just one nit but otherwise this looks good and simple enough!

src/kasn1.rs
Jakuje: I think I got confused with the lifetimes. Changed to use the `&[u8]`. Its indeed simpler.

simo5 (approved):
LGTM

PR #18, opened 2024-03-14
Various fixes for RSA PSS and ECC

Jakuje:
This covers various corner cases of RSA PSS mechanisms and ECC keys found during running the p11test against the kryoptic (CI integration in progress).

### DIFF
--- a/src/kasn1.rs
+++ b/src/kasn1.rs
@@ -29,6 +29,18 @@ impl<'a> DerEncBigUint<'a> {
             de = DerEncBigUint {
                 data: Cow::Owned(v),
             };
+        } else {
+            // Skip possible leading zeroes that do not affect sign of the resulting integer
+            let mut skip = 0;
+            while de.data[skip] == 0
+                && skip + 1 < de.data.len()
+                && de.data[skip + 1] & 0x80 == 0
+            {
+                skip += 1;
+            }
+            de = DerEncBigUint {
+                data: Cow::from(&data[skip..]),
+            };
         }
         /* check it works */
         match asn1::BigUint::new(&de.data) {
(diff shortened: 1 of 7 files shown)


src/kasn1.rs
```diff
@@ -37,6 +37,18 @@ impl<'a> DerEncBigUint<'a> {
         }
     }
 
+    /// Skip possible leading zeroes that do not affect sign of the resulting integer
+    pub fn new_unpad(data: &'a [u8]) -> KResult<Self> {
```
simo5, line 41: I do not think we should have a different new method.
It should be all done within new() because this is always relevant to create proper Strict DER encoding.

src/ossl/rsa.rs
```diff
@@ -1014,6 +1022,10 @@ impl Sign for RsaPKCSOperation {
                 if data.len() > self.max_input {
                     return err_rv!(CKR_DATA_LEN_RANGE);
                 }
+                if self.mech == CKM_RSA_PKCS_PSS && data.len() < self.max_input
```
simo5, line 1025: should we just have something like:
```
if match self.mech {
    CKM_RSA_PKCS => data.len() > self.max_input,
    CKM_RSA_PKCS_PSS => data.len() != self.max_input,
    _ => return err_rv!(CKR_GENERAL_ERROR),
    } { 
    return err_rv!(CKR_DATA_LEN_RANGE);
}
```

simo5 (commented):
Mostly looks good to me, just some minor nits

src/kasn1.rs
Jakuje: Sure. I will move it to normal new. I was thinking if it could break something, but it is unlikely.

src/ossl/rsa.rs
Jakuje: Sounds nicer. Let me change it.

simo5 (approved):
LGTM

PR #19, opened 2024-03-22
Add macro to deal with unsafe ptr to Vec

simo5:
Avoid lots of open coded convrsions and use a safer method to clone a pointer into an onwed vector.

Also remove from_raw_parts() call whre possible, replacing with safer and/or better alternatives.

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -7,6 +7,8 @@ use {super::fips, fips::*};
 #[cfg(not(feature = "fips"))]
 use {super::ossl, ossl::*};
 
+use super::bytes_to_vec;
+
 use std::ffi::{c_char, c_int, c_void};
 use zeroize::Zeroize;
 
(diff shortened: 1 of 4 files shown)


src/ossl/aes.rs
```diff
@@ -224,26 +217,18 @@ impl AesOperation {
...
-                                (*gcm_params).pAAD,
-                                (*gcm_params).ulAADLen as usize,
-                            )
-                            .to_vec()
-                        } else {
-                            Vec::new()
-                        }
+                    aad: if unsafe { (*gcm_params).ulAADLen } > 0 {
+                        bytes_to_vec!(unsafe { (*gcm_params).pAAD }, unsafe {
+                            (*gcm_params).ulAADLen
```
Jakuje, line 228: just a nit: Using temporary variable  for aadlen, similarly as done in the other cases, would make this block slightly more readable.

Jakuje (approved):
solves the issue of crashes when freeing the AesParams in  p11test.

src/ossl/aes.rs
simo5: Good point, fixed and updated

PR #21, opened 2024-03-25
Fix AES `C_Encrypt` with NULL result to query result size

Jakuje:
The `aes` module did not implement the `encryption_len()` which caused every operation with NULL result querying for the result size to fail.

There are some attempts to implement this inside of the encrypt() function, which became pointless and could be deleted (all within `if cipher.is_null` block:

https://github.com/latchset/kryoptic/blob/1c13e98edc766584281ac3dff7664b40a0afda35/src/ossl/aes.rs#L841-L851

But these are now used by wrapping, which does not have this handling separately and which I properly did not review yet so feel free to wait before merging this to figure out wrapping.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -855,7 +855,7 @@ extern "C" fn fn_encrypt(
         return CKR_OPERATION_NOT_INITIALIZED;
     }
     if encrypted_data.is_null() {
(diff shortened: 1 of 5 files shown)


simo5 (approved):
LGTM

PR #27, opened 2024-03-27
Use stricter data conversion to store data in sql

### LINKED ISSUE #20: Make sqlite storage file format platform neutral
CK_ULONG type attributes are stored as byte sequences, this makes them incompatible with reading the db on platform with different endianess or unsigned long size.

For maximum compatibility we should turn CK_ULONG values to a specific endian u32 value, before storage. (the PKCS#11 doesn't say it very explicitly but CK_ULONG values are supposed to be u32 compatible).

NULL values should also be supported, currently the code fails to read and convert a NULL value to something sensible, but some attributes may need to store a 0 length vector, to indicate the attribute is present, just intentionally empty.

simo5:
This allows us to be:
a) more portable files
b) stricter in validating data
c) support Null values which are needed in some rare case to indicate the attribute is present but empty (ulValueLen = 0).

Fixes #20

### DIFF
--- a/src/storage/sqlite.rs
+++ b/src/storage/sqlite.rs
@@ -1,10 +1,11 @@
 // Copyright 2024 Simo Sorce
 // See LICENSE.txt file for terms
 
-use rusqlite::{params, Connection, Rows, ToSql, Transaction};
+use rusqlite::{params, types::Value, Connection, Rows, Transaction};
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
+use super::super::attribute;
 use super::super::error;
 use super::super::interface;
 use super::super::object;
@@ -12,6 +13,7 @@ use super::super::{err_not_found, err_rv};
 
 use super::Storage;
 
+use attribute::AttrType;
 use error::{KError, KResult};
 use interface::*;
 use object::Object;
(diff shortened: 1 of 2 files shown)


src/storage/sqlite.rs
```diff
@@ -91,24 +93,59 @@ impl SqliteStorage {
...
+                            attribute::string_to_ck_date(s)?,
+                        ),
+                        None => return err_rv!(CKR_ATTRIBUTE_VALUE_INVALID),
+                    },
+                    AttrType::DenyType => {
+                        return err_rv!(CKR_ATTRIBUTE_TYPE_INVALID)
+                    }
+                    AttrType::IgnoreType => {
+                        return err_rv!(CKR_ATTRIBUTE_TYPE_INVALID)
+                    }
```
Jakuje, line 147: These two branches have the same content so it might be cleaner to merge them (not sure about the exact syntax now)
```suggestion
                    AttrType::DenyType,
                    AttrType::IgnoreType => {
                        return err_rv!(CKR_ATTRIBUTE_TYPE_INVALID)
                    }
```
Or catch them as `_`

src/storage/sqlite.rs
```diff
@@ -149,31 +181,37 @@ impl SqliteStorage {
...
+                    AttrType::BytesType => Value::from(a.to_buf()?),
+                    AttrType::DateType => {
+                        Value::from(a.to_attribute()?.to_date_string()?)
+                    }
+                    AttrType::DenyType => {
+                        return err_rv!(CKR_ATTRIBUTE_TYPE_INVALID)
+                    }
+                    AttrType::IgnoreType => {
+                        return err_rv!(CKR_ATTRIBUTE_TYPE_INVALID)
+                    }
```
Jakuje, line 199: same here

src/storage/sqlite.rs
```diff
@@ -197,8 +235,20 @@ impl SqliteStorage {
...
+            let col_id = Value::from(objid as i32);
+            let col_attr = Value::from(a.get_type() as u32);
+            let col_val = match a.get_attrtype() {
+                AttrType::BoolType => Value::from(a.to_bool()?),
+                AttrType::NumType => Value::from(a.to_ulong()? as u32),
+                AttrType::StringType => Value::from(a.to_string()?),
+                AttrType::BytesType => Value::from(a.to_bytes()?.clone()),
+                AttrType::DateType => Value::from(a.to_date_string()?),
+                AttrType::DenyType => continue,
+                AttrType::IgnoreType => continue,
```
Jakuje, line 247: and here

Jakuje (approved):
minor comments, but looks good enough.

src/storage/sqlite.rs
simo5: Fixed all of them

Jakuje (approved):
lgtm

PR #31, opened 2024-04-05
Fix Key unwrap

### LINKED ISSUE #30: Unwrapped AES keys do not work
The unwrapped AES keys have empty `CKA_VALUE`, which can be reproduced with the following test (it fails while importing the key as the `CKA_VALUE` in storage is empty):

https://github.com/Jakuje/kryoptic/commit/23d2a43419624644dcb2e8501491cda80b5bf67e

I was not completely sure how to fix it as I am slightly lost in the attributes flags. There is `OAFlags::UnsettableOnUnwrap`, which is used only on the secret keys `CKA_VALUE`, according to the name it makes sense, but the function that is checking this flag now operates only on template (instead of on the created `attrs` vec) here:
https://github.com/latchset/kryoptic/blob/b9dc257d2912042a0f28b5300d28d36667f4a51c/src/object.rs#L769

I think this should operate on `&atrs` but then we need to change the flags for `CKA_VALUE` so it  can be set from this function (now it would fail).

simo5:
The code was simply doing the wrong thing,
correctly use the template to generate the object and then add the CKA_VALUE attribute to it.

Fixes #30

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -751,24 +751,25 @@ pub trait SecretKeyFactory {
         mut data: Vec<u8>,
         template: &[CK_ATTRIBUTE],
     ) -> KResult<Object> {
-        let mut attrs = template.to_vec();
-        match attrs.iter().position(|x| x.type_ == CKA_VALUE_LEN) {
-            Some(idx) => {
-                let len = attrs[idx].to_ulong()?;
-                if (len as usize) < data.len() {
-                    unsafe { data.set_len(len as usize) };
+        let mut obj =
+            match template.iter().position(|x| x.type_ == CKA_VALUE_LEN) {
+                Some(idx) => {
+                    let len = template[idx].to_ulong()?;
+                    if (len as usize) < data.len() {
+                        unsafe { data.set_len(len as usize) };
+                    }
+                    if (len as usize) > data.len() {
+                        data.zeroize();
+                        return err_rv!(CKR_KEY_SIZE_RANGE);
+                    }
+                    let mut tmpl = template.to_vec();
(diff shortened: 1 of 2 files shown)


src/object.rs
```diff
@@ -751,24 +751,25 @@ pub trait SecretKeyFactory {
...
+                    let len = template[idx].to_ulong()?;
+                    if (len as usize) < data.len() {
+                        unsafe { data.set_len(len as usize) };
+                    }
+                    if (len as usize) > data.len() {
+                        data.zeroize();
+                        return err_rv!(CKR_KEY_SIZE_RANGE);
+                    }
+                    let mut tmpl = template.to_vec();
+                    let _ = tmpl.swap_remove(idx);
```
Jakuje, line 766: This removes the `CKA_VALUE_LEN` from the resulting object. But I would argue that it could be useful at times. It duplicates the information in the `CKA_VALUE`, if the value is readable, but otherwise there is no way for the caller to know if it is aes128, aes256 or key of different size.
So I think it should be kept in the object.

simo5: Uhmmm I had the impression that you could not keep it, but it turns out you only can't specify it on object creation (key import), nothing is said about the other cases.
However I think we remove it also in case of key generation.
The idea was to have a consistent representation of all keys with CKA_VALUE_LEN always omitted.
But you make a good argument about knowing the key length.

I will create a separate PR where I make sure it is always added to the key objects, together with some other stuff I found.

simo5: Followup bug in #32

Jakuje (approved):

PR #34, opened 2024-04-08
Improve secret key objects behavior

### LINKED ISSUE #32: store CKA_VALUE_LEN on all key objects
For asymmetric key types the size of the keys is obvious and encoded (one way or another) in public parameters.
However for symmetric keys, the only way to know the key length is to expose the CKA_VALUE_LEN parameter, because CKA_VALUE is never returned.
Ensure we always store CKA_VALUE_LEN (derived from CKA_VALUE or retained during generation/unwrap) on secret objects.

simo5:
The only way to get the key size for a secret key is if it store the CKA_VALUE_LEN as a separate attribute, given CKA_VALUE can never be returned.

So make sure we retain it or explicitly set for operations that forbid providing it in the input template.

Also make Generic Secrets not special, they are not and actually usually can't perform any encryption/decryption/etc.. operations.

Fixes #32

### DIFF
--- a/src/hmac.rs
+++ b/src/hmac.rs
@@ -198,7 +198,7 @@ pub fn register(mechs: &mut Mechanisms, ot: &mut ObjectFactories) {
                     ulMaxKeySize: 0,
                     flags: CKF_SIGN | CKF_VERIFY,
                 },
-                keytype: rs.1,
+                keytype: rs.4,
                 minlen: hashlen,
                 maxlen: hashlen,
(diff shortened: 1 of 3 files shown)


simo5:
Aannd ad a fix for HMAC stuff, as I found that it wasn't really working correctly if you used anythign but CKK_GENERIC_SECRET type keys ...

Jakuje (approved):

PR #35, opened 2024-04-19
Support configuration through environment variable

Jakuje:
This is rough PoC of allowing the kryoptic to work without the pReserved pointer and without the init args altogether (as they are not mandatory). The code can probably be simplified, but this worked for me and allowed me to run the general pkcs11 applications against kryoptic without issues.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -3,6 +3,7 @@
 
 use super::*;
 use hex;
+use std::env;
 use std::ffi::CString;
 use std::sync::Once;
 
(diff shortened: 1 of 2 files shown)


src/tests.rs
```diff
@@ -162,14 +178,66 @@ fn test_token(name: &str) {
...
+    testdata.setup_db(None);
+
+    let mut plist: *mut CK_FUNCTION_LIST = std::ptr::null_mut();
+    let pplist = &mut plist;
+    let result = C_GetFunctionList(&mut *pplist);
+    assert_eq!(result, 0);
+    unsafe {
+        let list: CK_FUNCTION_LIST = *plist;
+        match list.C_Initialize {
+            Some(value) => {
```
simo5, line 192: value -> init_fn  please

src/tests.rs
```diff
@@ -162,14 +178,66 @@ fn test_token(name: &str) {
...
```
simo5, line 217: same here

simo5 (approved):
LGTM

simo5:
Should we add a default place to look at when no env var is set ?
(perhaps in a separate PR)

simo5 (changes requested):
Sorry resetting status as I forgot I asked you to change a name

simo5 (approved):

simo5:
Thanks!

PR #38, opened 2024-05-01
Add initiall support for Key Derivation Functions

simo5:
This creates the necessary infrastructure.
Also gives an initial implementation of the SP800-108 KDF, which is one of the most complicated in the PCKCS#11 spec, and therefore was chosen to validate that the API would be able to handle all the quirks of this derivation API.
The current SP800-108 implementation limits quite heavily how parameters can be used as it uses the OpenSSL interpretation of it which has substantial limitations compared to the what the NIST SPEC and PKCS#11 allow.


src/kdf.rs
```diff
@@ -0,0 +1,351 @@
...
+        Ok(Sp800CounterFormat {
+            defined: true,
+            le: match cf.bLittleEndian {
+                0 => false,
+                1 => true,
+                _ => return err_rv!(CKR_MECHANISM_PARAM_INVALID),
+            },
+            bits: match cf.ulWidthInBits {
+                8 | 16 | 24 | 32 => cf.ulWidthInBits as usize,
+                _ => return err_rv!(CKR_MECHANISM_PARAM_INVALID),
```
rjrelyea, line 166: This restricts the max size of the format to 32 bits, which is probably fine, but we should note that. So every counter has to be 32 bits or less.

src/kdf.rs
```diff
@@ -0,0 +1,351 @@
...
+            },
+            le: match dkm.bLittleEndian {
+                0 => false,
+                1 => true,
+                _ => return err_rv!(CKR_MECHANISM_PARAM_INVALID),
+            },
+            bits: match dkm.ulWidthInBits {
+                8 | 16 | 24 | 32 => dkm.ulWidthInBits as usize,
+                _ => return err_rv!(CKR_MECHANISM_PARAM_INVALID),
```
rjrelyea, line 203: Same comment for counter... our encoded length has to be 32 bits or less. That doesn't mean length can't be more, just that longer lengths will wrap (which is now this is defined anyway). I don't know if SP 800 specifies that lengths can't be greater than 32 bits. NSS max is 64 bits because thats how big the underlying counter is. I don't know if any real world use would actually have more than 32 bits here.

src/sp800_108.rs
```diff
@@ -0,0 +1,396 @@
...
+) -> KResult<()> {
+    let len = match param.method {
+        CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS => slen,
+        CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS => klen,
+        _ => return err_rv!(CKR_MECHANISM_PARAM_INVALID),
+    };
+    /* up to 64 bits */
+    match param.bits {
+        8 => {
+            if len > MAX8 {
```
rjrelyea, line 83: I believe len is allowed to be > MAX8. you need to take the low 8 bits in that case. Same with 16, 24, and 32, so you just need to mask off the len before you convert it to data (len & 0xff for 8, len & 0xffff for 15, etc).

src/sp800_108.rs
```diff
@@ -0,0 +1,396 @@
...
+    op: &mut Box<dyn Sign>,
+) -> KResult<()> {
+    if !param.defined {
+        return err_rv!(CKR_MECHANISM_PARAM_INVALID);
+    }
+    match param.bits {
+        8 => {
+            if ctr > MAX8 {
+                return err_rv!(CKR_MECHANISM_PARAM_INVALID);
+            }
```
rjrelyea, line 29: See comment for DKLEN below... but for counters, its more likely that we don't want to overflow the counter because that could result in repeat blocks, so I wouldn't change the code here.

src/kdf.rs
```diff
@@ -0,0 +1,351 @@
...
+                {
+                    return err_rv!(CKR_ARGUMENTS_BAD);
+                }
+                let kdf_params = unsafe {
+                    *(mech.pParameter
+                        as *const CK_SP800_108_FEEDBACK_KDF_PARAMS)
+                };
+                Sp800Operation::feedback_kdf_new(kdf_params)?
+            }
+            CKM_SP800_108_DOUBLE_PIPELINE_KDF => {
```
rjrelyea, line 65: Ah, no double pipeline support, that's why i didn't see it in sp800_108. That's fine.

rjrelyea (commented):
I just reviewed src/kdf.rx and sp800_108.rs. I'ved added several comments, most are just notes and need no action. The only thing I saw that was worth changing was the dkm_update function. I believe length is allowed to overflow the format, in which case you just include the lower bits of the length. Counter is not allowed to overflow (as it would generate new copies of the same block, at least in counter mode). I've verified that NSS enforces this (through limiting the max counter) and that it's required bit FiPS ans SP-800 (based on the comment).

Otherwise I didn't see anything that was a problem. (I did not differences between NSS and this implementation, but they are neither spec breaking nor likely relevent).

src/object.rs
```diff
@@ -346,6 +346,47 @@ pub trait ObjectFactory: Debug + Send + Sync {
...
+    fn default_object_derive(
+        &self,
+        template: &[CK_ATTRIBUTE],
+        origin: &Object,
+    ) -> KResult<Object> {
+        /* FIXME: handle CKA_DERIVE_TEMPLATE */
+
+        let mut obj = self.internal_object_create(
+            template,
+            OAFlags::UnsettableOnUnwrap,
```
Jakuje, line 358: I see that unwrapping and deriving operations have some similar characteristics but reusing the "upwrap" here might sounds confusing so it should get at least a comment of we can not figure out a common name for these two operations and name the flag better `UnsettableOnUnwrapOrDerive`?

src/ossl/common.rs
```diff
@@ -261,6 +286,23 @@ impl OsslParam {
...
+    ) -> KResult<OsslParam> {
+        if self.finalized {
+            return err_rv!(CKR_GENERAL_ERROR);
+        }
+
+        let container = (val as c_int).to_ne_bytes().to_vec();
+        unsafe {
+            OSSL_PARAM_construct_int(key, container.as_ptr() as *mut c_int);
+        }
+        self.v.push(container);
```
Jakuje, line 302: Here is the same bug I am fixing in https://github.com/latchset/kryoptic/pull/39/commits/3baad23f27414d0a43f2574f289bd768ad9746b0
```suggestion
        let param = unsafe {
            OSSL_PARAM_construct_int(key, container.as_ptr() as *mut c_int)
        };
        self.v.push(container);
        self.v.push(param);
```
This compiles ok, but just throws away the parameter we created.

The above mentioned PR has fix for `size_t` case already.

src/tests.rs
```diff
@@ -4217,6 +4218,198 @@ fn test_key() {
     );
     assert_eq!(ret, CKR_OK);
 
+    /* Test key derivation */
```
Jakuje, line 4221: Can we have it in separate function? This one is terribly long already:
 * when it fails one needs to dig deeper into the code whats the issue
 * these tests cant be executed in parallel if it is infinite spaghetti code

Jakuje (commented):
mostly generic comments to the code as general. I did not dive into the KDF itself too much yet.

src/kdf.rs
```diff
@@ -0,0 +1,351 @@
...
+        Ok(Operation::Derive(Box::new(kdf)))
+    }
+}
+
+pub fn register(mechs: &mut Mechanisms, _: &mut ObjectFactories) {
+    Sp800Operation::register_mechanisms(mechs);
+}
+
+#[derive(Debug)]
+struct Sp800CounterFormat {
```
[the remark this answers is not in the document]
simo5: This is the name of the structure as derived from the Spec (CK_SP800_108_COUNTER_FORMAT), I know it is a bit confusing, but changing it from the spec would make it even harder to follow what is going on.

src/kdf.rs
simo5: This is directly from the sp800-108 spec:
```
10) r – An integer, smaller or equal to 32, that indicates the length of the binary
representation of the counter i.

```and NSS does the same.

File: src/kdf.rs (hunk 2 above)
Comment (simo5) [reply src/kdf.rs:203]: Uhm this one here is kinda wrong, I need to make it go up to 64bit, just because the spec puts no limits on it, but probably nothing will ever come close to 32 bit even ... yet there is no point to restrict to 32 bit, 64bit values may be used in some KDF that wants to take in a 64bit usize/size_t parameters for lengths.

File: src/kdf.rs [hunk 10]
```diff
@@ -0,0 +1,351 @@
...
+            CKM_SHA512_HMAC => Self::check_key_op(key, CKK_SHA512_HMAC),
+            CKM_SHA3_224_HMAC => Self::check_key_op(key, CKK_SHA3_224_HMAC),
+            CKM_SHA3_256_HMAC => Self::check_key_op(key, CKK_SHA3_256_HMAC),
+            CKM_SHA3_384_HMAC => Self::check_key_op(key, CKK_SHA3_384_HMAC),
+            CKM_SHA3_512_HMAC => Self::check_key_op(key, CKK_SHA3_512_HMAC),
+            CKM_AES_CMAC => Self::check_key_op(key, CKK_AES),
+            _ => return err_rv!(CKR_KEY_TYPE_INCONSISTENT),
+        }
+    }
+    fn counter_kdf_new(
```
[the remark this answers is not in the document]
simo5: Yeah, not sure why fix-format didn't adjust it and clang does not complain... looks like space between functions is not part of what it checks/requires.

src/object.rs
simo5: Ah yeah ... I'll think about this, maybe there is another more generic term that can subsume both behaviors somehow.
Naming ... the easy part of programming :)

src/ossl/common.rs
simo5: Ok I will merge your PR first and then rebase this code on top of it and fix the conflict.

src/sp800_108.rs
simo5: I have a patch to change these errors into just truncating the value, will submit soon, thanks for spotting this.

src/sp800_108.rs
simo5: I have a patch to change these errors into just truncating the value, will submit soon, thanks for spotting this.

src/sp800_108.rs
simo5: We can't overflow the counter per spec, the size of the counter is an explicit way to control how many iterations are allowed.

src/tests.rs
simo5: I have another patch coming that adds a whole bunch of things.
I am thinking that I want to add more spaghetti in this PR, and then work on a separate PR to completely reorganize the test code in separate files.
Do you mind bearing with me until then :)

src/kdf.rs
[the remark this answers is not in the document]
rjrelyea: Yes. The confusion is generated by PKCS #11 which sometimes has counters in iteration values and sometimes not, but then has external counters (when the iteration isn't a counter). From a certain taxanomy (counter mode has a counter for the iteration, while feedback has the last block), it make sense, but once you can add counters to feedback it all becomes confusing when you put it into code. In any case it would be for confusing to someone who has figured out the PKCS #11 spec to change the name here.

src/kdf.rs
simo5: Yeah OpenSSL simply does not have it, so I deferred adding it at all.

src/object.rs
simo5: Although not exactly part of this PR this comment made me look more carefully at the flags and I realized there was a lot of inconsistency and the code had changed at some point and stopped enforcing some stuff it was supposed to.

I attached an extra patch to deal with that.

src/kdf.rs
```diff
@@ -0,0 +1,343 @@
...
+
+use super::mechanism;
+use mechanism::*;
+
+use super::bytes_to_vec;
+
+use std::fmt::Debug;
+
+const SP800_MODE_COUNTER: &[u8; 8] = b"counter\0";
+const SP800_MODE_FEEDBACK: &[u8; 9] = b"feedback\0";
```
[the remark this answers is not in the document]
simo5: they are moved already in a later patch in this patchset

src/mechanism.rs
```diff
@@ -252,6 +256,23 @@ pub trait Verify: MechOperation {
...
+    fn derive(
+        &mut self,
+        _: &object::Object,
+        _: &[CK_ATTRIBUTE],
+        _: &Mechanisms,
+        _: &ObjectFactories,
+    ) -> KResult<(Object, usize)> {
+        err_rv!(CKR_GENERAL_ERROR)
+    }
+    fn derive_additional_key(
```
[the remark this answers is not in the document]
simo5: I couldn't see anything

src/object.rs
```diff
@@ -218,15 +218,11 @@ bitflags! {
...
-        const Unsettable           = 0b00000001000000;
-        const UnsettableOnCreate   = 0b00000011000000;
-        const UnsettableOnGenerate = 0b00000101000000;
-        const UnsettableOnUnwrap   = 0b00001001000000;
-        const NeverSettable        = 0b00001111000000;
+        const RequiredOnCreate     = 0b00000000001000;
+        const RequiredOnGenerate   = 0b00000000010000;
+        const AlwaysRequired       = 0b00000000100000;
+        const SettableOnlyOnCreate = 0b00000010000000;
+        const NeverSettable        = 0b00000100000000;
```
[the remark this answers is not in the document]
simo5: added a commit with comments

src/kdf.rs
```diff
@@ -0,0 +1,343 @@
...
+            Box::new(Sp800KDFMechanism {
+                info: CK_MECHANISM_INFO {
+                    ulMinKeySize: 0,
+                    ulMaxKeySize: std::u32::MAX as CK_ULONG,
+                    flags: CKF_DERIVE,
+                },
+                mech_type: CKM_SP800_108_COUNTER_KDF,
+            }),
+        );
+    }
```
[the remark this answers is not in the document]
simo5: moved in the right commit

src/mechanism.rs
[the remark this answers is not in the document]
simo5: in traits though I intentionally left no spaces as there is nothing really to gain from them ...

src/kdf.rs
```diff
@@ -0,0 +1,343 @@
...
+                } else {
+                    return err_rv!(CKR_MECHANISM_PARAM_INVALID);
+                }
+            } else {
+                if params.ulIVLen == 0 {
+                    Vec::new()
+                } else {
+                    return err_rv!(CKR_MECHANISM_PARAM_INVALID);
+                }
+            },
```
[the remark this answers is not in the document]
simo5: eh I kinda like it in its horribleness, but will see what I can do.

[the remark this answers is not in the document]
simo5: I moved iv computation out, hopefully good enough

src/sp800_108.rs
```diff
@@ -0,0 +1,386 @@
...
+         *
+         * This is an attempt at supporting insanity :-) */
+
+        let mechanism = CK_MECHANISM {
+            mechanism: self.prf,
+            pParameter: std::ptr::null_mut(),
+            ulParameterLen: 0,
+        };
+        let mech = mechanisms.get(self.prf)?;
+        let mut op = mech.sign_new(&mechanism, key)?;
```
Jakuje, line 268: Slightly related, but RSA/EC keys have `key.check_key_ops` to check the key can be used for a given operation in `mech.sign_new()`, but HMAC keys do not have this. Should the check be present verifying the HMAC key can be used for signature and have `CKA_SIGN` attribute? If so, would it mean that the key used for derive here would need to be usable also for signing or given that this is done internally, the check should be skipped?

src/tests.rs
```diff
@@ -4506,3 +4506,662 @@ fn test_key() {
...
+            };
+            data.push(section);
+            continue;
+        }
+        let section = match data.last_mut() {
+            Some(s) => s,
+            None => continue,
+        };
+        if section.prf == CK_UNAVAILABLE_INFORMATION {
+            continue;
```
Jakuje, line 4610: just an idea, likely bad. Would it make sense to parse also these, pass them to the pkcs11 module and just expect graceful error as unsupported rather than skipping altogether?

simo5: Not sure why that would help, these are tests that us a prf we do not parse/support.
I could have simply removed the tests from the files (like they did in NSS), but I preferred to just ignore unsupportable algorithms (these are Three DES stuff). The supportable, but unimplemented algorithm (AES_CMAC) you can see is skipped explitily later and is instead parsed.

simo5:
I changed a bit the code to remove Mechanism trait functions that the 1.78 compiler warned as unused anywhere.

Reorganized the kdf code to avoid some duplication, and added a guard to set and check finalized so the various functions can only be called in the correct order and the correct number of times.

No functional changes.

src/ossl/common.rs
```diff
@@ -57,6 +58,14 @@ ptr_wrapper!(EvpMdCtx; EVP_MD_CTX; EVP_MD_CTX_free);
...
 ptr_wrapper!(EvpCipherCtx; EVP_CIPHER_CTX; EVP_CIPHER_CTX_free);
 ptr_wrapper!(EvpCipher; EVP_CIPHER; EVP_CIPHER_free);
+ptr_wrapper!(EvpKdfCtx; EVP_KDF_CTX; EVP_KDF_CTX_free);
+ptr_wrapper!(EvpKdf; EVP_KDF; EVP_KDF_free);
+
+pub const CIPHER_NAME_AES128: &[u8; 7] = b"AES128\0";
+pub const CIPHER_NAME_AES192: &[u8; 7] = b"AES192\0";
+pub const CIPHER_NAME_AES256: &[u8; 7] = b"AES256\0";
+pub const MAC_NAME_CMAC: &[u8; 5] = b"CMAC\0";
+pub const MAC_NAME_HMAC: &[u8; 5] = b"HMAC\0";
```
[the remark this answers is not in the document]
simo5: These are just generic names that can be used with OpenSSL by other code, so I intentionally placed them in ossl/common.rs

Jakuje (approved):

PR #39, opened 2024-05-06
Fixes various corner cases of AES and RSA operations

Jakuje:
* The OsslParams construction was ignoring some values, which caused all RSA keys generated were always default 2k
* The CCM wrapping output length was dependent on the DataLen parameter, which is not really available for the caller at the time of calculation so changing to the length of provided data allows the caller to set the parameter to reasonable value for final operation to actually work
* Fix the expected lenght calculation for `AES_CBC_PAD` mechanism.

With these changes, the current set of tests in OpenSSL `p11test` works.


simo5 (changes requested):
The last commit needs to changed to actually fail if data_len != params.datalen

simo5:
> The last commit needs to changed to actually fail if data_len != params.datalen

Also ok to drop it and we do it later

simo5 (approved):
LGTM

PR #41, opened 2024-05-08
More kdf

simo5:
This PR depends on PR #38 which introduces KDF infrastructure.

It will be rebased as needed (so all those commits will disappear once we merge the dependency).

Adds 2 new simple Derive mechanisms, mostly because it was easy to add them and they cause me to make minor changes that will come useful for other more complex derive functions to come.


simo5:
Converted to draft so we do not risk merging until we have the other PR merged.

Jakuje (approved):
did not find any obvious issues in the remaining changes on top of #38 .

PR #44, opened 2024-05-13
Refactor some more tests

simo5:
Use helpers to minimize the size of the tests, and mak them easier to read.

### DIFF
--- a/src/tests/util.rs
+++ b/src/tests/util.rs
@@ -26,6 +26,20 @@ macro_rules! ret_or_panic {
     };
 }
 
+macro_rules! err_or_panic {
+    ($ret:expr, $err:expr) => {
+        if !match $ret {
+            Ok(_) => false,
+            Err(e) => match e {
+                KError::RvError(r) => r.rv == $err,
+                _ => false,
+            },
+        } {
+            panic!("Should have returned error {}", $err);
+        }
+    };
+}
+
 pub fn get_test_data(
     session: CK_SESSION_HANDLE,
     name: &str,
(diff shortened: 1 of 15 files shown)


src/tests/util.rs
```diff
@@ -348,21 +348,17 @@ pub fn sig_gen_multipart(
...
-        &class as *const _,
-        CK_ULONG_SIZE
-    ));
+    capacity: usize,
+) -> Vec<CK_ATTRIBUTE> {
+    let mut template = Vec::<CK_ATTRIBUTE>::with_capacity(if capacity > 0 {
+        capacity
+    } else {
+        ulongs.len() + bytes.len() + bools.len()
+    });
```
Jakuje, line 361: This is just a though if it would not be better to change the `capacity` argument to something meaning "allocate N more" as in all the calls you use `ulongs.len() + bytes.len() + bools.len() + 1` or `0`. If the meaning would be "the given attributes + N", it would simplify both the calculation here to `ulongs.len() + bytes.len() + bools.len() + N` as well as for the callers. They would just provide 0 or 1.

Jakuje (approved):
just minor suggestion for the API, but otherwise looks good.

src/tests/util.rs
simo5: I was wondering if I should do that at all TBH, it is just an optimization to avoid resizing the vector in those cases where we push additional elements outside of the basic use.

But except for the import objects operations it turns out we never do.
I will just drop the whole thing.
These are just tests and optimizing a vector allocation here is kind of silly :)

simo5: Remove and actually had to refactor how the template is built as I was taking references to objects on the stack ... which are gone after the helper is called ... oops.

Jakuje (approved):

PR #47, opened 2024-05-17
Add AES [C]MAC support

simo5:
This adds CKM_AES_MAC and CKM_AES_CMAC support.

AES_MAC is an old MAC mechanism described in FIPS 113 and discontinued in 2008 but still available in the PKCS#11 spec. It has no real test vectors and is disabled in FIPS builds.

AES_CMAC is the standard CMAC for AES and is tested also via the SP800 108 test vectors.


src/ossl/aes.rs
```diff
@@ -1660,3 +1658,232 @@ impl Decryption for AesOperation {
...
+            _key: mackey,
+            _mac: mac,
+            ctx: ctx,
+            maclen: maclen,
+        })
+    }
+
+    fn begin(&mut self) -> KResult<()> {
+        if self.in_use {
+            return err_rv!(CKR_OPERATION_NOT_INITIALIZED);
```
Jakuje, line 1759: Is this appropriate error? Should it by `CKR_OPERATION_ACTIVE` or something similar?

src/ossl/aes.rs
```diff
@@ -1660,3 +1658,232 @@ impl Decryption for AesOperation {
...
+
+        Ok(())
+    }
+
+    fn finalize(&mut self, output: &mut [u8]) -> KResult<()> {
+        if self.finalized {
+            return err_rv!(CKR_OPERATION_NOT_INITIALIZED);
+        }
+        /* It is valid to finalize without any update */
+        self.in_use = true;
```
Jakuje, line 1785: should this go to `in_use = false` to allow new `begin()`?

Jakuje: (same in the aes_mac.rs)

src/ossl/aes.rs
simo5: I've used CKR_OPERATION_NOT_INITIALIZED everywhere, not just in this code.
I guess we could use CKR_OPERATION_ACTIVE, but need to check the spec for the exact semantics of when it is ok to return it.
Keep in mind this is incorrect use, so 99.9% of applications should enver see this error, I would not hand this PR on this, but perhaps file an issue to review all use of CKR_OPERATION_* errors?

src/ossl/aes.rs
simo5: No, a new begin MUST NOT be allowed, begin() is used to "protect" one shot operations from being initiated after update() operations have already been started.

A new one shot operation require cancelling of finalizing the current operation and then initiating a new one via C_SignInit

Jakuje:
With the code from #48:
```
Running codespell on commit message of 2a928558a7e8898ef4959b111bc76839c590e2ab...
1: Move some utilitiy function to a new misc.rs file
	utilitiy ==> utility
```

simo5:
> With the code from #48:
> 
> ```
> Running codespell on commit message of 2a928558a7e8898ef4959b111bc76839c590e2ab...
> 1: Move some utilitiy function to a new misc.rs file
> 	utilitiy ==> utility
> ```

Fixed

Jakuje (approved):

src/aes.rs
```diff
@@ -315,6 +315,74 @@ impl Mechanism for AesMechanism {
...
         Ok(Operation::Derive(Box::new(kdf)))
     }
+
+    fn mac_new(
+        &self,
+        mech: &CK_MECHANISM,
+        key: &Object,
+        op_type: CK_FLAGS,
+    ) -> KResult<Box<dyn Mac>> {
+        /* the mechanism adveritzes only SIGN/VERIFY to the callers
```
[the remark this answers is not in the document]
simo5: UK vs US spelling, either is fine

PR #48, opened 2024-05-20
Codespell commit messages

Jakuje:


simo5:
Shouldn't we also add running codespell on CI ?

Jakuje:
I thought ist not there but it is part of the formatter.

simo5 (approved):
LGTM

PR #49, opened 2024-05-20
Add HMAC backend for OpenSSL HMAC for the FIPS build

simo5:
Convert and unify test Vector code for AES CMAC to cover also HMAC.


Jakuje (approved):
lgtm

PR #50, opened 2024-05-20
Add AES Key Wrap support

### LINKED ISSUE #13: Implement CKM_RSA_AES_KEY_WRAP
Requires some thinking about using one operation from another operation (nesting).
Also requires implementing CKM_AES_KEY_WRAP_KWP

simo5:
Includes applicable test vectors from CMVP

Resolves #13


Jakuje:
Resolves #13?

Jakuje (approved):
code wise looks good, but it looks like the commit codespell does not like the `main..` revision even though I believe it should be correctly pulled at that time. I tried to fix it, but without success. Any idea what could be wrong?

simo5:
> code wise looks good, but it looks like the commit codespell does not like the `main..` revision even though I believe it should be correctly pulled at that time. I tried to fix it, but without success. Any idea what could be wrong?

In PRs there is no checked out main branch I think, so you would have to reference it as origin/main ?

simo5:
Thanks for the review, merging.

Jakuje:
> > code wise looks good, but it looks like the commit codespell does not like the `main..` revision even though I believe it should be correctly pulled at that time. I tried to fix it, but without success. Any idea what could be wrong?
> 
> In PRs there is no checked out main branch I think, so you would have to reference it as origin/main ?

It should be there:

https://github.com/latchset/kryoptic/blob/main/.github/workflows/check-style.yml#L36

PR #53, opened 2024-05-22
Make Token Label settable

simo5:
Refactor a bit how Token Info is managed and store some of the token data in a special object in the database.

Adjust test IDs to not stomp over the lower 10 IDs, consider them reserved for internal token objects.


src/pkcs11/interface.rs
```diff
@@ -14,12 +14,16 @@ pub struct CK_INTERFACE {
...
-    CKR_VENDOR_DEFINED + KRYERR_OFFSET + 1;
+/* Object types */
+pub const KRY_TOKEN_DATA: CK_OBJECT_CLASS = KRY_VENDOR_OFFSET + 1;
+
+/* Attributes */
+pub const KRA_MAX_LOGIN_ATTEMPTS: CK_ATTRIBUTE_TYPE = KRY_VENDOR_OFFSET + 1;
+pub const KRA_FLAGS: CK_ATTRIBUTE_TYPE = KRY_VENDOR_OFFSET + 2;
+
+/* Errors */
+pub const KRYERR_TOKEN_NOT_INITIALIZED: CK_ULONG = KRY_VENDOR_OFFSET + 1;
```
Jakuje, line 27: should we stick to the three-letters prefix and do `KRR_` for return codes?

src/pkcs11/interface.rs
```diff
@@ -14,12 +14,16 @@ pub struct CK_INTERFACE {
...
-pub const KRYATTR_OFFSET: CK_ULONG = 485259;
-pub const KRYATTR_MAX_LOGIN_ATTEMPTS: CK_ULONG =
-    CKA_VENDOR_DEFINED + KRYATTR_OFFSET + 1;
+pub const KRY_VENDOR_OFFSET: CK_ULONG = CKA_VENDOR_DEFINED + 485259;
 
-pub const KRYERR_OFFSET: CK_ULONG = 485259;
-pub const KRYERR_TOKEN_NOT_INITIALIZED: CK_ULONG =
-    CKR_VENDOR_DEFINED + KRYERR_OFFSET + 1;
+/* Object types */
+pub const KRY_TOKEN_DATA: CK_OBJECT_CLASS = KRY_VENDOR_OFFSET + 1;
```
Jakuje, line 20: Should we do the `KRO_` for object class?

The current way is slightly non-consistent.

Jakuje (approved):

src/pkcs11/interface.rs
simo5: Had not thought about those, but yeah makes sense!

src/pkcs11/interface.rs
simo5: Will change

PR #57, opened 2024-06-15
Implement common function to set default key object values

### LINKED ISSUE #56: Add CKA_KEY_GEN_ MECHANISM to all keys
The attribute identifies the generating mechanism or defaults to CK_UNAVAILABLE_INFORMATION
Can never be set only internal mechanisms can set it

simo5:
Adds a helper to set some common default values.

Add test infrastructure to test arbitrary attributes in objects.

Fixes #56

### DIFF
--- a/src/tests/util.rs
+++ b/src/tests/util.rs
@@ -511,3 +511,106 @@ pub fn generate_key_pair(
     }
     Ok((pub_key, pri_key))
 }
+
+struct CheckTemplate<'a> {
+    ulongs: Vec<(CK_ATTRIBUTE_TYPE, CK_ULONG, CK_ULONG)>,
+    bytes: Vec<(CK_ATTRIBUTE_TYPE, Vec<u8>, &'a [u8])>,
+    bools: Vec<(CK_ATTRIBUTE_TYPE, CK_BBOOL, CK_BBOOL)>,
+    tmpl: Vec<CK_ATTRIBUTE>,
+}
+
+pub fn check_attributes(
+    session: CK_SESSION_HANDLE,
+    handle: CK_OBJECT_HANDLE,
+    ulongs: &[(CK_ATTRIBUTE_TYPE, CK_ULONG)],
+    bytes: &[(CK_ATTRIBUTE_TYPE, &[u8])],
+    bools: &[(CK_ATTRIBUTE_TYPE, bool)],
+) -> Option<String> {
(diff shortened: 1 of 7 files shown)


src/tests/util.rs
```diff
@@ -511,3 +511,106 @@ pub fn generate_key_pair(
...
+}
+
+pub fn check_attributes(
+    session: CK_SESSION_HANDLE,
+    handle: CK_OBJECT_HANDLE,
+    ulongs: &[(CK_ATTRIBUTE_TYPE, CK_ULONG)],
+    bools: &[(CK_ATTRIBUTE_TYPE, bool)],
+    bytes: &[(CK_ATTRIBUTE_TYPE, &[u8])],
+) -> Option<String> {
+    let capacity = ulongs.len() + bytes.len() + bools.len();
```
Jakuje, line 529: nit: would be nicer to read if the parts would be sorted the same way as the arguments.
```suggestion
    let capacity = ulongs.len() + bools.len() + bytes.len();
```

src/tests/util.rs
```diff
@@ -511,3 +511,106 @@ pub fn generate_key_pair(
...
+    let mut template = CheckTemplate {
+        bytes: Vec::with_capacity(bytes.len()),
+        longs: Vec::with_capacity(ulongs.len()),
+        bools: Vec::with_capacity(bools.len()),
```
Jakuje, line 533: The `CheckTemplate` arguments sorts them yet in another order so I would find any usage of this API (even though its in tests and internal) almost impossible from memory without checking the actual order over and over again.

src/tests/util.rs
```diff
@@ -511,3 +511,106 @@ pub fn generate_key_pair(
...
+    }
+    for b in bools {
+        template
+            .bools
+            .push((b.0, 0, if b.1 { CK_TRUE } else { CK_FALSE }));
+        if let Some(t) = template.bools.last() {
+            template.tmpl.push(make_attribute!(
+                t.0,
+                std::ptr::addr_of!(t.1),
+                std::mem::size_of::<CK_BBOOL>()
```
Jakuje, line 564: we have this size shortcut defined already so lets use it.
```suggestion
                CK_BBOOL_SIZE
```

simo5: ok

Jakuje (approved):
lgtm

PR #59, opened 2024-06-17
RSA: Accept empty OAEP parameter label

Jakuje:
My reading of the PKCS#11 specs is that the RSA-OAEP parameters source needs to be `CKZ_DATA_SPECIFIED` regardless if the lablel is specified or not. Currently, the combination of `source=CKZ_DATA_SPECIFIED` and NULL source data and `ulSourceDataLen=0` is crashing now.

### DIFF
--- a/src/ossl/rsa.rs
+++ b/src/ossl/rsa.rs
@@ -281,10 +281,13 @@ fn parse_oaep_params(mech: &CK_MECHANISM) -> KResult<RsaOaepParams> {
             }
             None
         }
-        CKZ_DATA_SPECIFIED => Some(bytes_to_vec!(
-            unsafe { (*oaep_params).pSourceData },
-            unsafe { (*oaep_params).ulSourceDataLen }
-        )),
+        CKZ_DATA_SPECIFIED => match unsafe { (*oaep_params).ulSourceDataLen } {
+            0 => None,
+            _ => Some(bytes_to_vec!(
+                unsafe { (*oaep_params).pSourceData },
+                unsafe { (*oaep_params).ulSourceDataLen }
+            )),
+        },
         _ => return err_rv!(CKR_MECHANISM_PARAM_INVALID),
     };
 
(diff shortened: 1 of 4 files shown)


simo5:
Uhm the spec only sasy this:
> Array of CK_BYTE containing the value of the encoding parameter. If the parameter is empty, pSourceData must be NULL and ulSourceDataLen must be zero.

Which is awfully underspecified, but clearly we should never crash.

src/ossl/rsa.rs
```diff
@@ -281,10 +281,13 @@ fn parse_oaep_params(mech: &CK_MECHANISM) -> KResult<RsaOaepParams> {
...
-            unsafe { (*oaep_params).pSourceData },
-            unsafe { (*oaep_params).ulSourceDataLen }
-        )),
+        CKZ_DATA_SPECIFIED => match unsafe { (*oaep_params).ulSourceDataLen } {
+            0 => None,
+            _ => Some(bytes_to_vec!(
+                unsafe { (*oaep_params).pSourceData },
+                unsafe { (*oaep_params).ulSourceDataLen }
+            )),
+        },
```
simo5, line 290: What about introducing an option_bytes_to_vec() macro ?

simo5 (approved):
Just a nit, if you feel like it, otherwise ack.

src/ossl/rsa.rs
Jakuje: I did not want to introduce more macros just for wrapping the output into option. Most of the cases already check the length/pointer for NULL and I do not think we have this use of option anywhere else now.

Looking a bit at the code, it looks like the similar issue will likely be in AES code when `nonce/iv` will be NULL, so I would like to fix that too.

Not sure if it would be better to have "safe" macro doing also the Option step and instead of crashing returning None, which could be handled in the caller, but some uses of this already require specific length so it would be needles double-checking.

simo5 (approved):
LGTM

PR #61, opened 2024-06-26
Add support for ECDH

Jakuje:
This adds a support for the simple ECDH with and without cofactor and with some possible shared data and hashing.

Test vectors I found now cover just the raw ECDH without shared data as I did not find usable vectors for the other variants. The vectors inform about failure reason, which can be in the following steps, which we do not execute (they are also in the pyca test vectors but I was not able to figure out how they are filtered out).

I am also not sure if it makes sense to run both init and resp test vectors (I saw both of them included in the pyca, but only one executed?).

I would also like to do some tests shared data to at least execute the code paths to OpenSSL.

Possible extension might be the MQV (not sure if implemented in OpenSSL though).

Opening now as a draft as it will most likely need some final touches.


src/tests/ecdh_vectors.rs
```diff
@@ -0,0 +1,363 @@
...
+            }
+        }
+    }
+}
+
+#[test]
+#[parallel]
+fn test_ecdh_vector_init() {
+    let test_data = parse_ecdh_vector(
+        "testdata/KASValidityTest_ECCStaticUnified_KDFConcat_NOKC_init.fax",
```
simo5, line 331: I don't think we need 2 sets of vetors, chose the one that is most representative (IIRC they just contain the same number and type of tests, the difference is one is the vector that was handed out and the other is an example of responses sent back for validation.

Usually the resp vector is the more useful, but often they have just identical type of data.

src/ossl/ecc.rs
```diff
@@ -709,3 +714,162 @@ impl Verify for EccOperation {
...
+}
+
+pub const X962KDF_NAME: &[u8; 8] = b"X963KDF\0";
+pub const NONE_NAME: &[u8; 1] = b"\0";
+fn ckd_to_type_slice(mech: CK_EC_KDF_TYPE) -> &'static [u8] {
+    match mech {
+        CKD_SHA1_KDF | CKD_SHA224_KDF | CKD_SHA256_KDF | CKD_SHA384_KDF
+        | CKD_SHA512_KDF | CKD_SHA3_224_KDF | CKD_SHA3_256_KDF
+        | CKD_SHA3_384_KDF | CKD_SHA3_512_KDF => X962KDF_NAME,
+        // TODO BLAKE2B digests ?
```
simo5, line 748: we don't do blake at this time, no need for TODO

src/ossl/ecc.rs
```diff
@@ -709,3 +714,162 @@ impl Verify for EccOperation {
...
+}
+
+fn ckd_to_digest_name(mech: CK_EC_KDF_TYPE) -> KResult<Vec<u8>> {
+    let digest = ckd_to_digest_name_slice(mech);
+    let mut v = Vec::with_capacity(digest.len());
+    v.extend_from_slice(digest);
+    Ok(v)
+}
+
+pub const X962KDF_NAME: &[u8; 8] = b"X963KDF\0";
```
simo5, line 741: Use OSSL_KDF_NAME_X963KDF, see comment on the code below for more

src/ossl/ecc.rs
```diff
@@ -709,3 +714,162 @@ impl Verify for EccOperation {
...
+        )?;
+        if self.kdf != CKD_NULL {
+            params = params
+                .add_utf8_string(
+                    name_as_char(OSSL_EXCHANGE_PARAM_KDF_TYPE),
+                    &ckd_to_type(self.kdf)?,
+                )?
+                .add_utf8_string(
+                    name_as_char(OSSL_EXCHANGE_PARAM_KDF_DIGEST),
+                    &ckd_to_digest_name(self.kdf)?,
```
simo5, line 795: As above just use:
```
.add_const_c_string(
    name_as_char(OSSL_EXCHANGE_PARAM_KDF_TYPE),
    name_as_char(ckd_to_digest_name(self.kdf)),
)?
```

where ckd_to_digest_name just returns the const buffer name from ossl/bindings like `mech_type_to_digest_name()` (in ossl/common.c) does.

simo5 (changes requested):
Some required simplifications noted, mostly looks good.

src/kdf.rs
```diff
@@ -54,6 +54,7 @@ impl Mechanism for Sp800KDFMechanism {
     }
 
     fn derive_operation(&self, mech: &CK_MECHANISM) -> KResult<Operation> {
+        // FIXME is this check needed when the same one is in fn_derive_key ?
```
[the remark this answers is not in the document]
Jakuje: Removed the comments and added them also to the ecc derive to keep consistent.

src/ossl/ecc.rs
```diff
@@ -709,3 +714,162 @@ impl Verify for EccOperation {
...
+
+        let mut params = OsslParam::with_capacity(6).set_zeroize().add_int(
+            name_as_char(OSSL_EXCHANGE_PARAM_EC_ECDH_COFACTOR_MODE),
+            if self.cofactor { 1 } else { -1 } as i32,
+        )?;
+        if self.kdf != CKD_NULL {
+            params = params
+                .add_utf8_string(
+                    name_as_char(OSSL_EXCHANGE_PARAM_KDF_TYPE),
+                    &ckd_to_type(self.kdf)?,
```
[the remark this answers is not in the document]
Jakuje: Rewrote. Thanks. I took the inspiration exactly there, but somehow messed the conversions ...

src/ossl/ecc.rs
```diff
@@ -709,3 +714,162 @@ impl Verify for EccOperation {
...
+                ctx.as_mut_ptr(),
+                secret.as_mut_ptr(),
+                &mut secret_len,
+            )
+        };
+        if res != 1 {
+            return err_rv!(CKR_DEVICE_ERROR);
+        }
+
+        obj.set_attr(from_bytes(CKA_VALUE, secret.to_vec()))?;
```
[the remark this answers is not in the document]
Jakuje: dunno. Probably some copy&pasta. Fixed.

src/tests/ecdh.rs
```diff
@@ -0,0 +1,114 @@
...
+    );
+    let session = testtokn.get_session(true);
+
+    /* login */
+    testtokn.login();
+
+    /* private key */
+    let mut handle: CK_ULONG = CK_INVALID_HANDLE;
+    let template =
+        make_attr_template(&[], &[(CKA_UNIQUE_ID, "11".as_bytes())], &[]);
```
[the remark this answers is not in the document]
Jakuje: This test actually reuses the ecc test data which I wanted to avoid introducing new ones. I will try to handle that later with adding some more tests.

src/ecc.rs
```diff
@@ -437,8 +483,109 @@ impl Mechanism for EccMechanism {
     }
 }
 
+#[derive(Debug)]
+struct ECDHOperation<'a> {
+    kdf: CK_EC_KDF_TYPE,
+    cofactor: bool,
```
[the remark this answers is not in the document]
Jakuje: I was not sure when I started writing this how much will be different in the openssl layer based on these two algorithms, but it is just one byte in params.

 I thought making it more explicit than implicit based on the pkcs11 mechanism type in the openssl layer (and mixing the layers).

src/ossl/ecc.rs
```diff
@@ -709,3 +714,162 @@ impl Verify for EccOperation {
...
+            }
+            Err(_) => (),
+        };
+        params = params.finalize();
+
+        let mut pkey = object_to_ecc_private_key(key)?;
+        let mut ctx = new_pkey_ctx_from_obj(&mut pkey)?;
+        // First we need to initialize the derive operation -- otherwise OpenSSL will fail to set
+        // params specific for the derive operation.
+        let res = unsafe { EVP_PKEY_derive_init(ctx.as_mut_ptr()) };
```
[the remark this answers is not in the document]
Jakuje: Good point. Fixed.

Jakuje:
Addressed the comments. I would like to add some more tests for different configuration (shared data, cofactor, digests before merging (and fix the UNIQUE_ID as mentioned in your review).

src/ecc.rs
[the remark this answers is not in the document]
simo5: in general I found that having the mechanism is more valuable, but it is just a judgment call, mixing layers is not a problem, the ossl/* code is just part of the mechanism implementation as anything else.

src/ossl/ecc.rs
```diff
@@ -709,3 +714,153 @@ impl Verify for EccOperation {
         Ok(self.output_len)
     }
 }
+
+fn ckd_to_digest_type_name(mech: CK_EC_KDF_TYPE) -> *const c_char {
```
simo5, line 718: nit: this should probably be ckd_to_digest_name, without "type"

Jakuje:
The X963KDF variant is obviously broken: openssl/openssl/issues/24750 -- not sure if it is important though. I can workaround the tests with setting the expected key length ...

src/ossl/ecc.rs
Jakuje: Thanks. Fixed.

simo5:
> The X963KDF variant is obviously broken: [openssl/openssl/issues/24750](https://github.com/openssl/openssl/issues/24750) -- not sure if it is important though. I can workaround the tests with setting the expected key length ...

I buy the upstream argument that X963KDF requires to explicitly request for a length, I do not like that it has to be passed in as a param ... but well ...what can you do, that's their API.

src/tests/ecdh.rs
```diff
@@ -0,0 +1,154 @@
...
+            ulParameterLen: iv.len() as CK_ULONG,
+        }
+    ));
+    assert_eq!(dec.len(), data.len());
+    assert_eq!(data.as_bytes(), dec.as_slice());
+
+    /* derive with shared data (and cofactor) */
+    let shared = "shared data";
+    let mut params = CK_ECDH1_DERIVE_PARAMS {
+        kdf: CKD_SHA512_KDF,
```
simo5, line 116: Can we add an explicit test with the SP800 variants?
You may look at how I did testing in tests/kdfs.rs where I have an array of values and run through the testing multiple times so many combinations are tested.

simo5 (approved):
Only NITS, code-wise looks good to me.

src/ossl/ecc.rs
```diff
@@ -709,3 +714,153 @@ impl Verify for EccOperation {
...
+        }
+        if self.shared.len() > 0 {
+            params = params.add_octet_string(
+                name_as_char(OSSL_EXCHANGE_PARAM_KDF_UKM),
+                &self.shared.to_vec(),
+            )?;
+        }
+
+        let mut obj = objfactories.derive_key_from_template(key, template)?;
+        match obj.get_attr_as_ulong(CKA_VALUE_LEN) {
```
simo5, line 805: After some discussion internally sounds like this check is not up to par.
Probably something like  https://github.com/latchset/kryoptic/blob/eee73043e7232714f17c857e034e1bb65de8acaa/src/kdf.rs#L778 should be done here.

Some key type (like AES) already check that CKA_VALUE_LEN is acceptable but only if non 0: https://github.com/latchset/kryoptic/blob/eee73043e7232714f17c857e034e1bb65de8acaa/src/aes.rs#L128

simo5 (changes requested):

src/ossl/ecc.rs
Jakuje: Agreed. Will add the check.

Jakuje: This should be resolved and similar check was implemented also in the ECDH case.

src/tests/ecdh.rs
Jakuje: Added test for SP800 variant. I thought I could use the test vectors for that, but from their description I was not able to figure out how to plug them together to get the right answer. The parsing and calling code is there, but the resulting key is not checked.

src/ossl/ecc.rs
```diff
@@ -709,3 +714,153 @@ impl Verify for EccOperation {
...
+                    m as u32,
+                )?;
+            }
+            Err(_) => (),
+        };
+        params = params.finalize();
+
+        let mut pkey = object_to_ecc_private_key(key)?;
+        let mut ctx = new_pkey_ctx_from_obj(&mut pkey)?;
+        // First we need to initialize the derive operation -- otherwise OpenSSL will fail to set
```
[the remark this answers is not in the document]
Jakuje: Removed.

src/ossl/ecc.rs
```diff
@@ -709,3 +716,149 @@ impl Verify for EccOperation {
...
+            } as i32,
+        )?;
+        let factory =
+            objfactories.get_obj_factory_from_key_template(template)?;
+
+        let keylen = match template.iter().find(|x| x.type_ == CKA_VALUE_LEN) {
+            Some(a) => a.to_ulong()? as usize,
+            None => {
+                let max = hash::hash_size(
+                    self.hash.ok_or(to_rv!(CKR_MECHANISM_INVALID))?,
```
simo5, line 768: How does this work given self.hash is an Option ?
What about *not* making it an Option and instead return KResult from it, and call the function here instead of saving the result on self.hash ?

src/ossl/ecc.rs
```diff
@@ -709,3 +716,149 @@ impl Verify for EccOperation {
...
+            EVP_PKEY_derive(
+                ctx.as_mut_ptr(),
+                std::ptr::null_mut(),
+                &mut secret_len,
+            )
+        };
+        if res != 1 {
+            return err_rv!(CKR_DEVICE_ERROR);
+        }
+        let mut secret = vec![0u8; secret_len];
```
simo5, line 838: change order between this one and the next?

src/ossl/ecc.rs
```diff
@@ -709,3 +716,149 @@ impl Verify for EccOperation {
...
+        };
+        if res != 1 {
+            return err_rv!(CKR_DEVICE_ERROR);
+        }
+
+        let mut tmpl = template.to_vec();
+
+        tmpl.push(CK_ATTRIBUTE::from_slice(
+            CKA_VALUE,
+            &secret[(secret_len - keylen)..],
```
simo5, line 857: It is surprising that you have to pick the tail of the buffer ...

src/ossl/ecc.rs
```diff
@@ -709,3 +716,149 @@ impl Verify for EccOperation {
...
+        ));
```
simo5, line 858: Don't you have to also add CKA_VALUE_LEN if it was not provided in the original template?

src/ossl/ecc.rs
```diff
@@ -709,3 +716,149 @@ impl Verify for EccOperation {
...
+            } else {
+                -1
+            } as i32,
+        )?;
+        let factory =
+            objfactories.get_obj_factory_from_key_template(template)?;
+
+        let keylen = match template.iter().find(|x| x.type_ == CKA_VALUE_LEN) {
+            Some(a) => a.to_ulong()? as usize,
+            None => {
```
simo5, line 766: I think you may want to add a test where you do not specify CKA_VALUE_LEN so we ensure this code path works as intended.

src/ecc.rs
```diff
@@ -437,8 +457,99 @@ impl Mechanism for EccMechanism {
...
+        CKD_SHA1_KDF | CKD_SHA1_KDF_SP800 => Some(CKM_SHA_1),
+        CKD_SHA224_KDF | CKD_SHA224_KDF_SP800 => Some(CKM_SHA224),
+        CKD_SHA256_KDF | CKD_SHA256_KDF_SP800 => Some(CKM_SHA256),
+        CKD_SHA384_KDF | CKD_SHA384_KDF_SP800 => Some(CKM_SHA384),
+        CKD_SHA512_KDF | CKD_SHA512_KDF_SP800 => Some(CKM_SHA512),
+        CKD_SHA3_224_KDF | CKD_SHA3_224_KDF_SP800 => Some(CKM_SHA3_224),
+        CKD_SHA3_256_KDF | CKD_SHA3_256_KDF_SP800 => Some(CKM_SHA3_256),
+        CKD_SHA3_384_KDF | CKD_SHA3_384_KDF_SP800 => Some(CKM_SHA3_384),
+        CKD_SHA3_512_KDF | CKD_SHA3_512_KDF_SP800 => Some(CKM_SHA3_512),
+        _ => None,
```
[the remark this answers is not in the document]
simo5: probably better to return KResult<CK_MECHANISM_TYPE> here considering the '_' should not happen.

simo5 (changes requested):

src/ossl/ecc.rs
```diff
@@ -709,3 +716,149 @@ impl Verify for EccOperation {
         Ok(self.output_len)
     }
 }
+
+pub const NONE_NAME: &[u8; 1] = b"\0";
```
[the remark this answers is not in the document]
Jakuje: This is the empty c string. Maybe we can find one in openssl headers though.

I initially though this is for both the NULL (plain)  and NIST, but looks like the NIST KDF does no go there and is not implemented in openssl at all so I will drop that.

https://github.com/openssl/openssl/blob/7afa7731e924d5ac10fc992d8cd777f407d33af9/providers/implementations/exchange/ecdh_exch.c#L259

If we would like to support this, I think we would have to implement it with the separate KDF calls then.

src/ossl/ecc.rs
Jakuje: Will do.

src/ossl/ecc.rs
Jakuje: The pkcs11 way if I understand it right: 

> This mechanism derives a secret value, and truncates the result according to the CKA_KEY_TYPE attribute of the template and, if it has one and the key type supports it, the CKA_VALUE_LEN attribute of the template.  (**The truncation removes bytes from the leading end of the secret value.**)  The mechanism contributes the result as the CKA_VALUE attribute of the new key; other attributes required by the key type must be specified in the template.

https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061194

src/ossl/ecc.rs
Jakuje: The `factory.create(tmpl.as_slice())?` already does it with `obj.check_or_set_attr()`.

src/ossl/ecc.rs
Jakuje: Will do.

src/ossl/ecc.rs
simo5: is this true for all cases or only for the the case where no hashing is used?

src/ossl/ecc.rs
simo5: cool, thanks for reminding me that!

src/ossl/ecc.rs
Jakuje: The intention was to fail early as part of validating the parameters in src/ecc.rs instead of this deep in the openssl layer, but it introduces this weird unwrapping here .. not sure if it is better.. Certainly this error should be `CKR_MECHANISM_PARAM_INVALID` though).

src/ossl/ecc.rs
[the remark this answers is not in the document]
Jakuje: This might even work without explicit setting of this to empty string if we want the plain ... might be nicer to exclude that, but initially I wanted to be as specific as possible in regards to the params.

src/ossl/ecc.rs
simo5: you can call the function twice, it is cheap.
Just return CKR_GENERAL_ERROR from it and convert that to return CKR_MECHANISM_PARAM_INVALID when you use it in the first invocation to validate parameters.
like:
```
if kdf_type_to_hash_mech(p.kdf).is_err() { 
    return CKR_MECHANISM_PARAM_INVALID };
}
```

src/ossl/ecc.rs
Jakuje: I read it as in all cases.

src/ossl/ecc.rs
Jakuje: This looks like this part was wrong. We should not look at hash here, but the plain ECDH max is the curve bits and X9.63 requires the length explicitly so there is no default max length ...

src/ossl/ecc.rs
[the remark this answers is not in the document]
Jakuje: Removed the setting of empty string. I set only the x963 when needed now.

src/ecc.rs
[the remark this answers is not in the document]
Jakuje: changed to KResult.

src/ossl/ecc.rs
Jakuje: This applies to the raw ECDH only, which has fixed-length output based on the field length. The X9.63 (with hashing kdf) requires the length as an input already so no truncation is happening there (as `secret_len = keylen` here). I might make it more explicit if you want.

src/ossl/ecc.rs
Jakuje: I rewrote this as this was plainly wrong.

src/ossl/ecc.rs
```diff
@@ -709,3 +715,156 @@ impl Verify for EccOperation {
         Ok(self.output_len)
     }
 }
+
+fn new_pkey_ctx_from_obj(pkey: &mut EvpPkey) -> KResult<EvpPkeyCtx> {
```
simo5, line 719: I think you need to use lifetimes here, otherwise the pkey may be deallocated while you are still holding a reference to it.

Rust can't see that you are because this is done through unsafe constructs in the C function being called.

Arguably lifetimes should be also used in EvpPkeyCtx::from_ptr(), I will look into that later (opening an issue).

src/ecc.rs
```diff
@@ -437,8 +457,104 @@ impl Mechanism for EccMechanism {
...
+            finalized: false,
+            mech: mechanism,
+            kdf: p.kdf,
+            hash: if p.kdf == CKD_NULL {
+                None
+            } else {
+                Some(kdf_type_to_hash_mech(p.kdf)?)
+            },
+            shared: shared,
+            public: unsafe {
```
simo5, line 557: given you are using to_vec() later, why not immediately use bytes_to_vec! here?
It is going to be safer, as it immediately alloacetes and copies the buffer, no risk the buffer disappears before it is used.

simo5 (changes requested):
Mostly minor nits.

However the lifetimes one is something I need to check carefully elsewhere as well. I think in general I made sure to avoid issues by storing the actual pointers in the operation context in other places, so Rust would not drop() them until the operation was concluded, but using lifetimes is the correct way to go about them.
If you store the pkey on the operation context you can avoid dealing with lifetimes in the function for now, and I will deal with it comprehensively later.

src/ecc.rs
```diff
@@ -437,8 +457,104 @@ impl Mechanism for EccMechanism {
...
+            }
+        }
+        if p.pPublicData == std::ptr::null_mut() || p.ulPublicDataLen == 0 {
+            return err_rv!(CKR_MECHANISM_PARAM_INVALID);
+        }
+
+        let shared = if p.pSharedData != std::ptr::null_mut()
+            && p.ulSharedDataLen != 0
+        {
+            unsafe {
```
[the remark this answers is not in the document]
Jakuje: I think I was not satisfied with the `bytes_to_slice` not checking for the NULL pointer (only the len). This case is indeed invalid, but I thought it would be better to report error than crash.

I think the `bytes_to_vec()` can not handle this case either so I went to implement it from scratch. But I agree that it might be better to teach the `bytes_to_vec()` to handle empty pairs and safely return empty vec instead of checking the length like in the `aes.c` before every call that is non-mandatory value:

https://github.com/latchset/kryoptic/blob/eee73043e7232714f17c857e034e1bb65de8acaa/src/ossl/aes.rs#L246-L249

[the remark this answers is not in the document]
simo5: yes, or we can have a mcro that takes an other parameter, to which we can key off different behaviors:
for example if you pass in an error that is the error to return if either of the parameters is null/0 instead of returning an empty vec.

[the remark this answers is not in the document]
simo5: by which I mean you would add another implementation of the same macro when there is an additional parameter, I am ok for the default to return an empty vec I think.

src/ossl/ecc.rs
```diff
@@ -149,6 +150,13 @@ fn object_to_ecc_public_key(key: &Object) -> KResult<EvpPkey> {
     EvpPkey::from_ptr(pkey)
 }
 
+/// Convert the PKCS #11 public key object to OpenSSL EVP_PKEY
+fn object_to_ecc_public_key(key: &Object) -> KResult<EvpPkey> {
+    let curve_name = get_curve_name_from_obj(key)?;
+    let ec_point = get_ec_point_from_obj(key)?;
+    make_ecc_public_key(curve_name, ec_point)
```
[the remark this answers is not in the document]
Jakuje: I wanted to keep the function similar I have for the private key for readability. I can certainly avoid the intermediate variables.

src/ossl/ecc.rs
```diff
@@ -709,3 +715,156 @@ impl Verify for EccOperation {
...
+            | CKD_SHA3_384_KDF | CKD_SHA3_512_KDF => {
+                params = params
+                    .add_const_c_string(
+                        name_as_char(OSSL_EXCHANGE_PARAM_KDF_TYPE),
+                        OSSL_KDF_NAME_X963KDF.as_ptr() as *const c_char,
+                    )?
+                    .add_const_c_string(
+                        name_as_char(OSSL_EXCHANGE_PARAM_KDF_DIGEST),
+                        mech_type_to_digest_name(
+                            self.hash
```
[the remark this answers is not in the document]
Jakuje: ok. Will change

src/ossl/ecc.rs
Jakuje: Good point. I did not work with the lifetimes much yet so I am not sure how to easily approach it now, given that the contexts are all defined in

src/ecc.rs
Jakuje: Done.

src/ecc.rs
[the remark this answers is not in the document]
Jakuje: Changed to vec, fixed the `bytes_to_vec!()` macro and the obvious usages in aes. Skipped the more complicated/mandatory in kdf.

simo5 (approved):

PR #62, opened 2024-06-27
This is the next step in changing how storage works

simo5:
As a shortcut the initial implementation on PIN handling was done by storing plain text PINs in objects.

This PR changes completely how PINs are handled and how information about their status is exposed in the token.

The SO PIN is only used for initialization operations, and is stored as the result of a PBKDF2 operation with 10000 iterations.

The USER PIN uses the same derivation, but produces an AES key that is used to wrap a randomly generated other AES key (KEK) using AES GCM (so that the wrapped key is authenticated upon decryption).

~The KEK is currently not used, but this is the ground work to change the storage layer to encrypt all sensitive values before storing them.~
The KEK is used to encrypt all the attributes marked as Sensitive for all object types.

If the SO Pin is used to change the User PIN the KEK will be lost and the all the secrets in the Database will become unusable. In the future we may want to consider whether the KEK should be also encrypted in the SO PIN as a recovery option, but currently this is not done.


src/kdf.rs
```diff
@@ -747,12 +747,13 @@ impl Mechanism for PBKDF2Mechanism {
...
                     || params.ulPasswordLen == 0
                 {
-                    return err_rv!(CKR_MECHANISM_PARAM_INVALID);
+                    self.mock_password_object(vec![0u8; 0])?
+                } else {
+                    self.mock_password_object(bytes_to_vec!(
+                        params.pPassword,
+                        params.ulPasswordLen
+                    ))?
                 }
```
Jakuje, line 756: With change from https://github.com/latchset/kryoptic/pull/61/commits/0f6ec1842727585ef2248b24f973522da7e40ba1 you could squash this back to
```
                self.mock_password_object(bytes_to_vec!(
                    params.pPassword,
                    params.ulPasswordLen
                ))?
```

src/token.rs
```diff
@@ -39,6 +41,23 @@ const SO_PIN_UID: &str = "0";
 const USER_PIN_UID: &str = "1";
 const TOKEN_INFO_UID: &str = "2";
 
+const MAX_LOGIN_ATTEMPTS: CK_ULONG = 10;
+
+const USER_PIN_IV: &str = "USRPIN IV UNWRAP";
+const USER_PIN_AAD: &str = "USRPIN AUTH_DATA";
+const DEFPIN_SALT: &str = "DEFAULT SALT DATA"; /* at least 16 long for fIPS */
```
Jakuje, line 48:
```suggestion
const DEFPIN_SALT: &str = "DEFAULT SALT DATA"; /* at least 16B long for FIPS */
```

src/token.rs
```diff
@@ -441,215 +306,574 @@ impl Token {
...
+        label: String,
+        wrapped: Vec<u8>,
+    ) -> KResult<()> {
+        match self.storage.fetch_by_uid(&uid) {
+            Ok(o) => {
+                let mut obj = o.clone();
+                obj.set_attr(attribute::from_string(CKA_LABEL, label))?;
+                obj.set_attr(attribute::from_bytes(CKA_VALUE, wrapped))?;
+                obj.set_attr(attribute::from_ulong(KRA_LOGIN_ATTEMPTS, 0))?;
+                self.storage.store(&uid, obj)?;
```
Jakuje, line 341: So this will be completely implemented outside of the storage layer? This means the `enc` column can be removed from the database schema, as well as from todos in the sqlite backend?

https://github.com/latchset/kryoptic/blob/eee73043e7232714f17c857e034e1bb65de8acaa/src/storage/sqlite.rs#L43

src/token.rs
```diff
@@ -905,7 +1228,10 @@ impl Token {
                 if let Some(o) = self.session_objects.get_mut(&o_handle) {
                     o
                 } else {
-                    t = Some(self.storage.fetch_by_uid(s)?);
+                    /* we need to decrypt as the new object encrypts
+                     * to a different uid any wayi, and insert_object
```
Jakuje, line 1232:
```suggestion
                     * to a different uid any way, and insert_object
```

src/token.rs
simo5: That's a good q. On the one hand I think it would be nice to mark encrypted vs clear text values somehow. OTOH the solution i have now is cleaner than trying to cram it at the db layer.

Should I just drop that column for now?

Jakuje: If it is not used, I would drop it.
It might become needed when we will need some backward compatibility, we will change what is encrypted or how. But for now, it sits there unused and I would argue it might be confusing for readers.

simo5: dropping

simo5:
Fixed the broken commit and rebase on main

src/token.rs
```diff
@@ -662,80 +886,174 @@ impl Token {
...
+    fn encrypt_value(&self, uid: &String, val: &Vec<u8>) -> KResult<Vec<u8>> {
+        if let Some(ref kek) = self.kek {
+            let aad = uid.as_bytes();
+            let tag_len = 8;
+            let mut iv = [0u8; 12];
+            get_random_data(&mut iv)?;
+            let mut params = CK_GCM_PARAMS {
+                pIv: iv.as_ptr() as *mut CK_BYTE,
+                ulIvLen: 12,
+                ulIvBits: 96,
```
Jakuje, line 931: nit: would be good to keep the 12 in sync here by using either local `iv_len` variable or even constant to share it with the deciphering code.

The `ulIvBits` is `8*iv_len` so it could also be modified to avoid another magic number. Or it can be even skipped as this is not used by the current implementation.

src/token.rs
```diff
@@ -662,80 +886,174 @@ impl Token {
...
+                mechanism: CKM_AES_GCM,
+                pParameter: &mut params as *mut interface::CK_GCM_PARAMS
+                    as *mut _,
+                ulParameterLen: std::mem::size_of::<CK_GCM_PARAMS>()
+                    as CK_ULONG,
+            };
+            let aes = self.mechanisms.get(CKM_AES_GCM)?;
+            let mut op = aes.encryption_new(&mech, &kek)?;
+            let mut clen: CK_ULONG =
+                op.encryption_len(val.len() as CK_ULONG)? as CK_ULONG + 12;
```
Jakuje, line 946: another use of magic number 12 that should be replaced by `iv_len` or some shared value. There are two more on lines 953 and 957.

src/token.rs
```diff
@@ -662,80 +886,174 @@ impl Token {
...
+                    as *mut _,
+                ulParameterLen: std::mem::size_of::<CK_GCM_PARAMS>()
+                    as CK_ULONG,
+            };
+            let aes = self.mechanisms.get(CKM_AES_GCM)?;
+            let mut op = aes.decryption_new(&mech, &kek)?;
+            let mut plen: CK_ULONG =
+                op.decryption_len((val.len() - 12) as CK_ULONG)? as CK_ULONG;
+            let mut plain = Vec::<u8>::with_capacity(plen as usize);
+            op.decrypt(&val.as_slice()[12..], plain.as_mut_ptr(), &mut plen)?;
```
Jakuje, line 1007: same as the encryption -- the magic number 12 should be at least variable, better some constant.

Jakuje (commented):
With the encryption, the testsuite runs waaay slower.

simo5:
> With the encryption, the testsuite runs waaay slower.

I did not notice (to my surprise) a significant slow down, can you quantify what you see and in what conditions?

src/token.rs
```diff
@@ -442,215 +307,574 @@ impl Token {
...
 
-        /* remove all private session objects */
-        for handle in priv_handles {
-            let _ = self.session_objects.remove(&handle);
-            self.handles.remove(handle);
-        }
+    fn wrap_kek(
+        &mut self,
+        wrapper: &Object,
+        mut wrapped: Object,
```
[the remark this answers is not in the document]
simo5: no a bad idea, will change

src/token.rs
```diff
@@ -442,215 +307,574 @@ impl Token {
...
+    fn unwrap_kek(&self, wrapper: &Object, wrapped: &[u8]) -> KResult<Object> {
+        let tag_len = 8;
+        let params = CK_GCM_PARAMS {
+            pIv: USER_PIN_IV.as_ptr() as *mut CK_BYTE,
+            ulIvLen: USER_PIN_IV.len() as CK_ULONG,
+            ulIvBits: (USER_PIN_IV.len() * 8) as CK_ULONG,
+            pAAD: USER_PIN_AAD.as_ptr() as *mut CK_BYTE,
+            ulAADLen: USER_PIN_AAD.len() as CK_ULONG,
+            ulTagBits: (tag_len * 8) as CK_ULONG,
+        };
```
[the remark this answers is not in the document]
simo5: ok

Jakuje:
Just rough numbers on my machine show its 10x slower:
```
[jjelen@t14s kryoptic (storage3)]$ make check
test result: ok. 190 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 25.06s
[jjelen@t14s kryoptic (main)]$ make check
test result: ok. 184 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.60s
```

src/token.rs
simo5: going to create a helper function for CK_GCM_PARAMS for enc/dec as well, and will handle it there

src/token.rs
simo5: I will add a const variable with a name that makes it clearer

simo5:
> Just rough numbers on my machine show its 10x slower:
> 
> ```
> [jjelen@t14s kryoptic (storage3)]$ make check
> test result: ok. 190 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 25.06s
> [jjelen@t14s kryoptic (main)]$ make check
> test result: ok. 184 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.60s
> ```

Yeah, I just realized my machine takes 22s now ...
I will see if I can add a no-encryption mode, so we can use it for most tests

simo5:
Turned out 90% of the slow down was not due to encryption but due to the pbkdf to unlock the pin.

With encryption turned off I reduced local testing time from ~28s to ~23s, with pbkdf2 rounds brought down from 10000 to 1000 I got from ~23s to ~2.3s ...

simo5:
I addressed the requested changes and added commits to bring down the time in testing to:
test result: ok. 190 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.38s

src/token.rs
```diff
@@ -662,80 +897,181 @@ impl Token {
...
+        for h in handles {
+            let _ = self.session_objects.remove(&h);
+            self.handles.remove(h);
+        }
+    }
+
+    /* We should probably have lifetimes to ensure iv and aad are around for
+     * the lifetime of the returned structure, but this will require substantial
+     * reworking of the bindings, so for now we just get this comment.
+     * ENSURE the arguments stay in scope until CK_GCM_PARAMS is needed
```
Jakuje, line 936:
```suggestion
     * ENSURE the arguments stay in scope as long as CK_GCM_PARAMS is needed
```

src/token.rs
```diff
@@ -662,80 +897,181 @@ impl Token {
...
+            pAAD: aad.as_ptr() as *mut CK_BYTE,
+            ulAADLen: aad.len() as CK_ULONG,
+            ulTagBits: 64 as CK_ULONG,
+        }
+    }
 
-        self.update_token_pin_flags();
+    fn encrypt_value(&self, uid: &String, val: &Vec<u8>) -> KResult<Vec<u8>> {
+        if let Some(ref kek) = self.kek {
+            let mut iv = [0u8; DEFAULT_IV_SIZE];
```
Jakuje, line 951: I would use just `IV_SIZE` or `CRYPT_IV_SIZE` as I do not expect this will be somehow configurable in the long term, but its just a nit ...

Jakuje (approved):
just minor nits.

simo5:
Thanks for the review, I am too lazy to go through another round just for some minor wording, so merging as is :-)

PR #64, opened 2024-07-01
Make OpenSSL context wrappers safer

### LINKED ISSUE #63: Review helper construct's lifetime promises
In some helper structures in ossl/common.c we play with pointers of contexts and other structures, we need to review if lifetimes need to be explicitly used, to avoid cases where Rust may think it is safe to drop some memory while openssl embedded a pointer to the same memory in some internal context, like with pkey structures.

simo5:
Instead of exposing raw pointers which is quite unsafe, internalize as many unsafe operations as possible on these raw pointers so that the actual interface of these wrapped type is mostly safe.

Declare interfaces that still deal with raw pointers as unsafe so that proper care is done when handling them.

Fixes #63

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -58,7 +58,6 @@ pub struct Object {
     handle: CK_OBJECT_HANDLE,
     session: CK_SESSION_HANDLE,
     attributes: Vec<Attribute>,
-    modified: bool,
     zeroize: bool,
 }
 
@@ -78,7 +77,6 @@ impl Object {
             handle: CK_INVALID_HANDLE,
             session: CK_INVALID_HANDLE,
             attributes: Vec::new(),
-            modified: false,
             zeroize: false,
         }
     }
@@ -96,7 +94,6 @@ impl Object {
             let uuid = Uuid::new_v4().to_string();
             self.attributes
                 .push(attribute::from_string(CKA_UNIQUE_ID, uuid));
-            self.modified = true;
         }
     }
 
@@ -109,7 +106,6 @@ impl Object {
             }
             obj.attributes.push(attr.clone());
         }
-        obj.modified = true;
         Ok(obj)
     }
 
@@ -119,14 +115,6 @@ impl Object {
 
     pub fn get_handle(&self) -> CK_OBJECT_HANDLE {
         self.handle
-    }
-
-    pub fn reset_modified(&mut self) {
-        self.modified = false;
-    }
-
-    pub fn is_modified(&self) -> bool {
-        self.modified
     }
 
     pub fn set_session(&mut self, s: CK_SESSION_HANDLE) {
@@ -155,7 +143,6 @@ impl Object {
             Some(idx) => self.attributes[idx] = a,
             None => self.attributes.push(a),
         }
-        self.modified = true;
         Ok(())
     }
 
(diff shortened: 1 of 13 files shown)


src/object.rs
```diff
@@ -58,7 +58,6 @@ pub struct Object {
     handle: CK_OBJECT_HANDLE,
     session: CK_SESSION_HANDLE,
     attributes: Vec<Attribute>,
-    modified: bool,
```
Jakuje, line 61: Does rust have something like `gcc`s `-Werror` to treat warnings as an error. I think this could be handy for the CI to catch these issues in the future.

src/token.rs
```diff
@@ -160,6 +160,38 @@ impl Handles {
     }
 }
 
+/* Sow = Silly Object Wrapper */
+struct Sow<'a, T> {
+    ro: Option<&'a T>,
+    vo: Option<T>,
```
Jakuje, line 166: I think these would be better to have either comment or expanded names so the code would be readable. I understand that they are likely abbreviation for `ref object` and `value object`, but it might not be obvious on first sight. And given that they are internal, shortening them like this wont save users any time writing the names.

src/token.rs
Jakuje, line 1: Typos in the commit message: `erference -> reference`, `wranings -> warnings`.

Jakuje (commented):
just nits, looks good and saves a lot of repeated code!

src/object.rs
simo5: I am not sure, but we could add a grep in CI and fail the test if we really want to, but there is still a couple of warning that *will* be fixed soonish and I do not want to force ourselves to remove and readd code just because of them.
Also in FIPS mode some things end up being unused but I do not want to have to add [cfg(not(feature = "fips"))] around things like interfaces in mechanism.rs

src/token.rs
simo5: Chuecked the whole thing

src/ossl/common.rs
```diff
@@ -8,60 +8,270 @@ use interface::*;
...
+        Ok(EvpPkeyCtx { ptr: ptr })
+    }
+
+    /*
+    pub fn empty() -> EvpPkeyCtx {
+        EvpPkeyCtx {
+            ptr: std::ptr::null_mut(),
+        }
+    }
+    */
```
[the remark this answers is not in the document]
simo5: This was just stop gap, I asked on our rust channel and they suggested to use Cow, which does exactly what I need. I should have known, I used Cow previously, but somehow did not realize it could be easily used this way.
Replaced with simpler code

src/token.rs
simo5: This was just stop gap, I asked on our rust channel and they suggested to use Cow, which does exactly what I need. I should have known, I used Cow previously, but somehow did not realize it could be easily used this way.
Replaced with simpler code

src/ossl/common.rs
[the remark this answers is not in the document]
simo5: Yeah empty() is quite unsafe, there were uses before I did various changes to limit the use, and the forgot to remove the comment, will drop all uses of empty in a fixup.

Jakuje (approved):
lgtm

Jakuje (approved):

PR #65, opened 2024-07-02
Add support for default token

### LINKED ISSUE #37: Support default storage location
Should we add a default place to look at when no env var is set ?
(perhaps in a separate PR)

_Originally posted by @simo5 in https://github.com/latchset/kryoptic/issues/35#issuecomment-2066696699_

This was discussed previously and we considered some defaults like:
 * if `root`: some system wide location
 * else: try `XDG_CONFIG_HOME` [1] or something like that

The environment variable and pReserved argument to Initialize should override the defaults.

[1] https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

simo5:
This depends on (and includes all of the commits of) #62

Adds ability to search for a default token in a specific set of locations.

For example a user can create a personal token by just `touch ~/.local/share/kryoptic/token.sql` and then use a tool like pkcs11-tool to actually initialize the token.

Fixes #37


simo5:
apparently since the switch to serial v parallel test success is not deterministic anymore, it sometimes depends on the order in which tests are run.
@Jakuje is there even more control over serial tests? for example ability to set dependencies between tests, or at least an order of execution ?

simo5:
I do not get what is causing the test failure.
I have fixed an issue that could be caused by things running in different order, and the test now reliably passes for me locally ...

simo5:
Somehow one of the other test tokens gets slot number 0 even though our test infrastructure always starts from 1 when setting test file names ...

src/lib.rs
```diff
@@ -326,33 +327,70 @@ macro_rules! global_wlock {
...
+    /* Freedesktop specification for data dirs first
+     * then fallback to use $HOME/.local/share, if that is also not
+     * available see if we have access to a system store */
+    let datafile = match env::var("XDG_DATA_HOME") {
+        Ok(xdg) => format!("{}/kryoptic/{}", xdg, DEFAULT_CONF_NAME),
+        Err(_) => match env::var("HOME") {
+            Ok(home) => {
+                format!("{}/.local/share/kryoptic/{}", home, DEFAULT_CONF_NAME)
+            }
+            Err(_) => format!("/var/kryoptic/public/{}", DEFAULT_CONF_NAME),
```
Jakuje, line 366: should we fall back only on the presence of the HOME env variable or should we actually check the path exists before falling back?
I think that every well-behaved shell will have HOME env variable so on the first sight, I do not see the obvious case where the last case would kick in. Not sure if it is good or bad though.

src/lib.rs
```diff
@@ -370,6 +408,28 @@ extern "C" fn fn_initialize(_init_args: CK_VOID_PTR) -> CK_RV {
...
+            match wstate.get_token_from_slot(*s) {
+                Ok(t) => {
+                    if filename.eq(t.get_filename()) {
+                        return CKR_CRYPTOKI_ALREADY_INITIALIZED;
+                    }
+                }
+                Err(_) => return CKR_GENERAL_ERROR,
+            }
+        }
+        slotnum = (slot_high + 1) as u64;
```
Jakuje, line 430: what happens here if somebody creates a slot with ID UIN64_MAX value and then tries to initialize one more? Do we overflow here? Should we try to find first unused slot number? Could this have some side effects such as potential issues of slot ID reuse that could make login fails and token locked?

Jakuje (approved):
ack then

src/lib.rs
simo5: NOTE: I was wrong on the above assertion, "as" does not crash at all, it instead does a binary cast and potentially a truncation if the receiving type is smaller.

But in this case it does not matter because if "slot_high +1" were to overflow we'd get a panic with error: "attempt to add with overflow" (just tested).

PR #69, opened 2024-07-10
Fix creation of session objects

### LINKED ISSUE #68: Creating a session object does not require login
This is what is causing pkcs11-provider HKDF test to fail.
But creating a session object should succeeded even when not logged in, as long as we have a writable session.

simo5:
Creating a session object only requires a writable session, not full login.

Fixes #68


simo5:
Fixes the pkcs11-provide HKDF CI test, which can be enabled once this is merged

Jakuje (approved):

PR #72, opened 2024-07-23
Add support for CKA_ALWAYS_AUTHENTICATE behavior

### LINKED ISSUE #70: Support for CKA_ALWAYS_AUTHENTICATE
The `CKA_ALWAYS_AUTHENTICATE` should enforce the private key is used only immediately after providing the user consent (PIN).

The attribute can be provided as any other object during writing or generating private key and should be visible for the user.

This login state should be tracked as part of the session, which carries already information if the session is in logged in state. This needs to set specific flag the Login was called and it needs to be reset when different function will be called.

simo5:
When a private key has this attribute set to true it commands that any operation (decryption/signature/etc..) done with this key requires an extra authentication to authorize the individual operation.

This is done by setting flags on the session that will control proper authentication is checked if necessary, and support for the special CKU_CONTEXT_SPECIFIC user type into the fn_login() function.

Fixes #70

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -651,16 +651,44 @@ extern "C" fn fn_login(
     }
     let vpin: Vec<u8> = bytes_to_vec!(pin, pin_len);
     let mut token = res_or_ret!(rstate.get_token_from_slot_mut(slot_id));
-    match token.login(user_type, &vpin) {
-        CKR_OK => match rstate.change_session_states(slot_id, user_type) {
-            Ok(()) => CKR_OK,
-            Err(e) => {
+    if user_type == CKU_CONTEXT_SPECIFIC {
+        let session = res_or_ret!(rstate.get_session_mut(s_handle));
+        match session.get_operation() {
+            Err(e) => match err_to_rv!(e) {
+                CKR_USER_NOT_LOGGED_IN => (),
+                _ => return CKR_OPERATION_NOT_INITIALIZED,
+            },
+            Ok(_) => return CKR_OPERATION_NOT_INITIALIZED,
+        }
+    }
+
+    let result = token.login(user_type, &vpin);
+
+    if user_type == CKU_CONTEXT_SPECIFIC {
+        match result {
+            CKR_OK => {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -898,7 +926,8 @@ extern "C" fn fn_encrypt_init(
     let mech = res_or_ret!(token.get_mechanisms().get(data.mechanism));
     if mech.info().flags & CKF_ENCRYPT == CKF_ENCRYPT {
         let operation = res_or_ret!(mech.encryption_new(data, &obj));
-        session.set_operation(Operation::Encryption(operation));
+        session
+            .set_operation(Operation::Encryption(operation), obj.always_auth());
```
Jakuje, line 930: Does it make sense to check for this on public key operation which generally does not need any authentication?

src/lib.rs
```diff
@@ -1413,7 +1443,7 @@ extern "C" fn fn_verify_init(
     let mech = res_or_ret!(token.get_mechanisms().get(data.mechanism));
     if mech.info().flags & CKF_VERIFY == CKF_VERIFY {
         let operation = res_or_ret!(mech.verify_new(data, &obj));
-        session.set_operation(Operation::Verify(operation));
+        session.set_operation(Operation::Verify(operation), obj.always_auth());
```
Jakuje, line 1446: Same here. The `CKA_ALWAYS_AUTHENTICATE` is defined only for private key objects:

https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976581

And this operation is done on the public key object so this should be
```suggestion
        session.set_operation(Operation::Verify(operation), false);
```

Jakuje (changes requested):

src/lib.rs
simo5: The attribute is available only on private key objects

simo5: Oh, I get it, I just applied these to all operations and didn't think encryption and verification does not need it indeed ... yeah I will change those two places, good catch.

Jakuje (approved):

PR #79, opened 2024-08-07
Add vendor implementation of SSHKDF

simo5:
SSH's KDF (RFC 4253 Section 7.2) is different enough from other KDF's (notably is uses a Hash instead of a proper HMAC) that none of the existing KDFs can be used, so we need a new mechanism for it.

This is a vendor implementation so we can test that the interface is good enough and make it available for SSH implementations to use if they so desire.


src/pkcs11/extensions.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct KR_SSHKDF_PARAMS {
+    pub prfHashMechanism: CK_MECHANISM_TYPE,
+    pub derivedKeyType: CK_BYTE,
+    pub pExchangeHash: *mut CK_BYTE,
+    pub ulExchangeHash: CK_ULONG,
+    pub pSessionId: *mut CK_BYTE,
+    pub ulSessionId: CK_ULONG,
```
Jakuje, line 17: the pkcs11 naming usually goes like this to easily distinguish the length and data:
```suggestion
    pub ulExchangeHashLen: CK_ULONG,
    pub pSessionId: *mut CK_BYTE,
    pub ulSessionIdLen: CK_ULONG,
```

src/sshkdf.rs
```diff
@@ -0,0 +1,184 @@
...
+                    flags: CKF_DERIVE,
+                },
+            }),
+        );
+    }
+
+    fn new(mech: &CK_MECHANISM) -> KResult<SSHKDFOperation> {
+        let params = cast_params!(mech, KR_SSHKDF_PARAMS);
+
+        match hmac_size(params.prfHashMechanism) {
```
Jakuje, line 84: should this go as a `hash_size` given that it should be hash and not hmac?

src/ossl/sshkdf.rs
```diff
@@ -0,0 +1,72 @@
...
+        let mut dkey = objfactories.derive_key_from_template(key, template)?;
+        let keysize = match dkey.get_attr_as_ulong(CKA_VALUE_LEN) {
+            Ok(n) => n as usize,
+            Err(_) => return err_rv!(CKR_TEMPLATE_INCOMPLETE),
+        };
+        if keysize == 0 || keysize > (u32::MAX as usize) {
+            return err_rv!(CKR_KEY_SIZE_RANGE);
+        }
+
+        let params = OsslParam::with_capacity(5)
```
Jakuje, line 28:
```suggestion
        let params = OsslParam::with_capacity(6)
```
to include the terminating NULL record?

src/pkcs11/extensions.rs
simo5: oh right, I will change that

src/sshkdf.rs
simo5: uhmmm I am going to add another function that just tells me if the mechanism is a valid hash, meant to do that when I copied the skeleton of this function and then forgot

src/ossl/sshkdf.rs
simo5: ah, good point

simo5:
@Jakuje discussed with Bob, and I think I will make change to this code also to always return IVs (type A and B) as CKO_DATA object instead of CKO_SECRET_KEY, as IVs do not make sense as keys

src/pkcs11/extensions.rs
simo5: Fixed

src/sshkdf.rs
simo5: handled

Jakuje (approved):

PR #80, opened 2024-08-09
Add support for EdDSA

Jakuje:


src/ecc.rs
```diff
@@ -21,6 +21,8 @@ use object::{
 use once_cell::sync::Lazy;
 use std::fmt::Debug;
 
+include! {"ecc-misc.rs"}
```
simo5, line 24: why include this instead of simply importing via "use" ?

simo5 (approved):
One question, but only a curiosity, it can be merged.

src/ecc.rs
Jakuje: I think initially I had there some macros or some other stuff that did not like the `use` but in the end it ended up quite slim which might work with normal use. Let me try to change that

Jakuje: Updated to avoid the `include`.

PR #81, opened 2024-08-21
Improve use of CK_ATTRIBUTE arrays

simo5:
This code improves usage of CK_ATTRIBUTE arrays by creating a structure partially inspired by OsslParam, but using better methods to deal with lifetimes and avoiding copies where not needed.

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -8,6 +8,7 @@ use super::{bytes_to_vec, err_not_found, err_rv, sizeof};
 use error::{KError, KResult};
 use interface::*;
 
+use std::borrow::Cow;
 use zeroize::Zeroize;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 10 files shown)


src/attribute.rs
```diff
@@ -520,37 +521,191 @@ impl CK_ATTRIBUTE {
...
-            pValue: value as *const CK_ULONG as *mut std::ffi::c_void,
-            ulValueLen: sizeof!(CK_ULONG),
+impl<'a> CkAttrs<'a> {
+    pub fn new() -> CkAttrs<'static> {
+        Self::with_capacity(0)
+    }
+
+    pub fn with_capacity(capacity: usize) -> CkAttrs<'static> {
+        CkAttrs {
+            v: Vec::new(),
```
Jakuje, line 550: should this vector have the default capacity too?

Jakuje (approved):
just one nit. Otherwise lgtm

src/attribute.rs
simo5: Not really, in most cases you should not really add anything in it and instead just depend on lifetimes passing in references that do not need to be copied at all.

Only the add_owned_something() function need that vector, and I tried hard to avoid them most of the time as it is inefficient to copy stuff all the time when a simple pointer is enough.

The problem was making sure Rust catches when we are referencing a pointer so it won't free it before CkAttrs is, which the whole lifetime on the struct and the impl and the functions takes care of now that the raw CK_ATTRIBUTES slice could not.

PR #83, opened 2024-08-23
More safety for FFI array structures

simo5:
Some more fixes/optimizations for the CkAttrs struct
Change OsslParam to be more efficient like CkAttrs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,4 @@ zeroize = "1.6.0"
 
 [features]
 fips = ["rusqlite/bundled"]
+slow = [] # Enables slow tests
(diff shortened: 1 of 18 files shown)


Cargo.toml
```diff
@@ -40,3 +40,4 @@ zeroize = "1.6.0"
 
 [features]
 fips = ["rusqlite/bundled"]
+slow = [] # Enables slow tests
```
Jakuje, line 43: should we run the slow tests in CI? Are they enabled by default?

Jakuje (approved):

Cargo.toml
simo5: They are not enabled by default, and I am not sure we should enable them in CI.
Perhaps we should have a build after a PR merge that makes a pass with slow turned on ...

PR #84, opened 2024-08-26
Do not embed nulls in token info

simo5:
The copy function is filling token info fixed size strings with spaces However we failed to catch the trailing null that apparently comes in a &str type, ending up with embedding an illegal null in these token info strings.


Jakuje (approved):
good catch!

PR #86, opened 2024-08-28
Refactor Error handling

simo5:
Change codebase to use the customery Error and Result names instead of KError and KResult which are not familiar names in the ecosystem.

Simplify the actual code and provide more syntactic sugar features like automatic conversion of error types in some cases.

Redesign inspired by the recommendations at:
https://nrc.github.io/error-docs/error-design/error-type-design.html


Jakuje (approved):
lgtm

PR #87, opened 2024-08-29
Ensure default values are always stored

simo5:
Some of the common key attributes were not stored unless explicitly specified in the template. We should always store the default values because applications are allowed to query those attributes at any time and we need to be able to return them when they do.

Found by using pkcs11-tool --list-objects which would spit a lot of errors about unavailable attributes.


Jakuje (approved):
lgtm

PR #88, opened 2024-09-03
Improve safety of interfaces

### LINKED ISSUE #77: Consider reviewing all uses of "as" conversions
The "as" conversion truncates values when the receiving type is smaller leading to difficult to reason edge cases. We should review all uses of "as" to asses where try_from() is more appropriate.

simo5:
Now that the code is more mature we can address two long standing safety issues.

1) The pervasive use of blind casting using the 'as' construct (#77), which is unwise in this code base that crosses the FFI boundary in so many places.
In fact a bug was found thanks to bound checks using TryFrom (See the relevant commit)

2) The use of raw pointers in the inner "operation" interfaces, which forced the use of even more unsafe code in the internal functions than was really needed. Again various potential issues were addressed implicitly here (no explicit test failures).

Fixes #77

### DIFF
--- a/src/sp800_108.rs
+++ b/src/sp800_108.rs
@@ -6,7 +6,7 @@ use super::error;
 use super::interface;
 use super::mechanism;
 use super::object;
-use super::{err_rv, sizeof};
+use super::{err_rv, map_err, sizeof};
 
 use attribute::from_bytes;
 use error::Result;
@@ -26,7 +26,7 @@ macro_rules! maxsize {
     ($size: expr) => {
         match $size {
             8 | 16 | 24 | 32 | 40 | 48 | 56 => (1 << $size) - 1,
-            64 => u64::MAX as usize,
+            64 => usize::try_from(u64::MAX)?,
             _ => panic!("Invalid size"),
         }
     };
@@ -110,7 +110,7 @@ impl Sp800Operation {
             Box::new(Sp800KDFMechanism {
                 info: CK_MECHANISM_INFO {
                     ulMinKeySize: 0,
-                    ulMaxKeySize: std::u32::MAX as CK_ULONG,
+                    ulMaxKeySize: CK_ULONG::try_from(u32::MAX).unwrap(),
                     flags: CKF_DERIVE,
                 },
             }),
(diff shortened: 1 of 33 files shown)


src/sp800_108.rs
```diff
@@ -367,6 +373,8 @@ impl Sp800Operation {
         }
     }
 
+    /* NOTE: In this function the len is intnetionally truncated by
```
Jakuje, line 376:
```suggestion
    /* NOTE: In this function the len is intentionally truncated by
```
The same in the previous block.

src/storage/sqlite.rs
```diff
@@ -249,6 +252,35 @@ impl SqliteStorage {
         }
         Ok(objid)
     }
+
+    fn num_to_val(ulong: CK_ULONG) -> error::Result<Value> {
+        /* CK_UNAVAILABLE_INFORMATION need to be special cased */
+        /* for storage compatibility CK_ULONGs can only be stored as u32
```
Jakuje, line 258:
```suggestion
        /* CK_UNAVAILABLE_INFORMATION need to be special case */
        /* For storage compatibility CK_ULONGs can only be stored as u32
```

src/storage/sqlite.rs
```diff
@@ -249,6 +252,35 @@ impl SqliteStorage {
...
+         * values and PKCS#11 spec pay attentions to never allocate numbers
+         * bigger than what can be stored in a u32. However the value of
+         * CK_UNAVAILABLE_INFORMATION is defined as CK_ULONG::MAX which is
+         * a larger number than what we can store in a u32.
+         * Sqlite however can store bugger than u32 numbers, so we store
```
Jakuje, line 263:
```suggestion
         * Sqlite however can store bigger than u32 numbers, so we store
```

Jakuje (commented):
code wise looks good. Just few typos noticed

src/storage/sqlite.rs
simo5: Ah nice you pointed this out, as this explanation is actually outdated.
This was the explanation for the first attempt, I since changed the code to store -1 instead (still with the same exception approach for otherwise handling the case), which is more fitting.

Jakuje (approved):

PR #90, opened 2024-09-05
Add support for linking openssl dynamically

simo5:
For non-FIPS builds we can link to the system libcrypto without issues, so add that as an option.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [ossl3, fips, release]
+        name: [ossl3, fips, release, dynamic]
     container: fedora:latest
     steps:
       - name: Get Date for DNF cache entry
@@ -82,6 +82,9 @@ jobs:
           if [ "${{ matrix.name }}" = "release" ]; then
             cargo build -vv --release
           fi
+          if [ "${{ matrix.name }}" = "dynamic" ]; then
+            cargo build -vv --features dynamic
+          fi
 
       - name: Test
         run: |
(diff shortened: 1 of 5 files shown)


build.rs
```diff
@@ -136,7 +137,34 @@ fn build_ossl() {
         .expect("Couldn't write bindings!");
 }
 
+#[cfg(feature = "dynamic")]
```
Jakuje, line 140: should we do the CI target with dynamic linking to make sure this keeps working?

build.rs
```diff
@@ -136,7 +137,34 @@ fn build_ossl() {
...
+        .allowlist_item("ossl_.*")
+        .allowlist_item("OSSL_.*")
+        .allowlist_item("openssl_.*")
+        .allowlist_item("OPENSSL_.*")
+        .allowlist_item("CRYPTO_.*")
+        .allowlist_item("c_.*")
+        .allowlist_item("EVP_.*")
+        .allowlist_item("evp_.*")
+        .allowlist_item("BN_.*")
+        .allowlist_item("LN_aes.*")
```
Jakuje, line 157: Can this list be somehow reused from the non-dynamic linking as it is exactly same and keeping them in sync is quite error-prone?

build.rs
simo5: good idea!

build.rs
simo5: Yeah, let me add an auxiliary function for that

.github/workflows/build.yml
```diff
@@ -94,6 +97,9 @@ jobs:
           if [ "${{ matrix.name }}" = "release" ]; then
             cargo test --release
           fi
+          if [ "${{ matrix.name }}" = "dynamic" ]; then
+            cargo test --features dynamic
```
Jakuje, line 101: maybe one more idea that can be put in place later in follow-up PR or in long term is to run this on different OS/versions, not just Fedora as this is no longer staticaly compiled from "known-to-be-good" openssl version.

Jakuje (approved):

PR #95, opened 2024-09-16
Bump bindgen to 0.69

Jakuje:
The 0.68 is no longer available in Fedora packages and I do not think there is a reason to stay behind (if it will generally work with newer version).

I am not sure if this is the best approach or we should make it dependent on wider range of versions (or if we should set up dependabot or something to keep the dependencies updated in the long term).


simo5 (approved):
LGTM

simo5:
Dependabot tends to be very noisy, I'd avoid.

PR #96, opened 2024-09-16
Packaging template + architecture specific fixes

Jakuje:
Add packaging template + some more metadata to the cargo.

I tried to build this in copr, but it looks like Fedora repositories exploded so not all worked:

https://copr.fedorainfracloud.org/coprs/jjelen/kryoptic/

And where the repositories worked, we are failing on some architecture/size specific issues that are also fixed with this change.

Overall, having packing to build RPM package in copr on different architectures might be a good CI.


Jakuje:
Ugh ... it wont work this way. I guess I will have to cast to `::c_char` to handle arch specific issues ...

Jakuje:
oh .. there is more ...

packaging/kryoptic.spec
```diff
@@ -0,0 +1,69 @@
...
+%cargo_generate_buildrequires -n -f dynamic
+
+%build
+%cargo_build -n -f dynamic
+%{cargo_license_summary -n -f dynamic}
+%{cargo_license -n -f dynamic} > LICENSE.dependencies
+
+%install
+%cargo_install -n -f dynamic
+# FIXME: install shared library
```
simo5, line 54: Is this a FIXME that should be kept ?

packaging/kryoptic.spec
```diff
@@ -0,0 +1,69 @@
...
+%check
+%cargo_test -n -f dynamic
+%endif
+
+%files
+%license LICENSE.txt
+%license LICENSE.dependencies
+%doc README.md
+%doc dummy.txt
+%{_bindir}/conformance
```
simo5, line 66: We should delete this binary for now, it does nothing.

Jakuje:
Do we care about i686 builds? We assume that CK_ULONG is u64 in many places in the code:
```
error[E0308]: mismatched types
   --> src/storage/json.rs:96:62
    |
96  |                         Some(n) => attribute::from_ulong(id, n),
    |                                    ---------------------     ^ expected `u32`, found `u64`
    |                                    |
    |                                    arguments to this function are incorrect
    |
note: function defined here
   --> src/attribute.rs:331:29
    |
287 |         pub fn $fn1(t: CK_ULONG, val: $rtype) -> Attribute {
    |                                  -----------
...
331 | conversion_from_type! {make from_ulong; from_type_ulong; from_string_ulong; from CK_ULONG; as NumType; via ulong_to_vec}
    |                             ^^^^^^^^^^
help: you can convert a `u64` to a `u32` and panic if the converted value doesn't fit
    |
96  |                         Some(n) => attribute::from_ulong(id, n.try_into().unwrap()),
    |                                                               ++++++++++++++++++++
```

simo5 (changes requested):
One question and one change request

packaging/kryoptic.spec
Jakuje: Good catch! I had it in one of previous iterations, but it got lost. Will fix.

packaging/kryoptic.spec
Jakuje: Agreed! Will remove.

Jakuje:
The last build works everywhere but on i386:
https://copr.fedorainfracloud.org/coprs/jjelen/kryoptic/build/8027167/

There is something that looks like an issue of the bindgen (but only warning though):
```
warning: literal out of range for `u32`
 --> src/conformance/../pkcs11/bindings.rs:7:50
  |
7 | pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = 18446744073709551615;
  |                                                  ^^^^^^^^^^^^^^^^^^^^
  |
  = note: the literal `18446744073709551615` does not fit into the type `u32` whose range is `0..=4294967295`
  = note: `#[warn(overflowing_literals)]` on by default
```
The following
```
#define CK_UNAVAILABLE_INFORMATION ~0UL
```
is converted to
```
pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = 18446744073709551615;
```
which does not sound right. I will probably add it as a separate issue later.

Jakuje:
The code now compiles, but some tests fail on 32b arch so it still needs some love:

https://copr.fedorainfracloud.org/coprs/jjelen/kryoptic/build/8027499/

simo5:
I force updated your branch as I have merged another PR that caused conflicts

simo5 (approved):
LGTM, and I agree we can fix i386 later

PR #98, opened 2024-09-18
Initial support for selecting individual mechs to be built in

simo5:
This change allows us to select which mechanisms to build in and allows for a relatively easy way to add or remove mechanisms in custom builds.


simo5:
Given this change does not involve any functionality change and it is just a reorganization of existing code I will merge without a second review to allow other work to progress.

Issue #100, opened 2024-10-02
The default token location is confusing


rcritten:
The default token file location is determined in this order:

1. If KRYOPTIC_CONF is defined
2. if XDG_DATA_HOME is defined then $XDG_DATA_HOME/kryoptic/token.sql
3. if HOME is defined then $HOME/.local/share/kryoptic/token.sql
4. /var/kryoptic/public/token.sql

I'd suggest moving 4 to before 2.

The problem with relying on the environment variable is it has to be set or inherited by every user that will interact with the token. I think this would be cumbersome to do for IPA development (we don't recommend users use software HSMs in production). It is very difficult to troubleshoot just exactly which file is being opened. I had to resort to strace and sift through the output.

This is likely to cause issues with any secondary tooling like certmonger which runs as a daemon. We'd have to, only for this one provider, set an environment variable in the systemd scripts.

Similarly we'd have to add it to .bashrc to ensure that  the token will be visible to tools like certutil.

An alternative, which might bring new issues, would be to define a configuration file that describes the location. I think the problem with my location suggestion and with a configuration file it could preclude users from having a nice default of their own if there is a system-level token because it is an existence check. If the file exists then it is used.

simo5:
I was wondering if ti made sense to create a default config file in /etc/kryoptic/kryoptic.conf, but allow a $XDG_CONFIG_DIR one to override it if present.

What do you think ?

simo5:
and change KRYOPTIC_CONF to point to a config file rather then directly to a database file ...

rcritten:
Yes that makes sense.

rcritten:
It would be nice if the configuration file and its directives was documented in README.md

simo5:
Yep, I'll open an issue for improving docs

PR #102, opened 2024-10-04
Add configuration file support

simo5:
To aid cases where an environment variable cannot be easily set nor init arguments can be easily passed we need a place where we can store the tokens configuration, including the ability to specify multiple slots.

The ability to use init args and ignore any configuration is retained but the init arguments format has been enhanced to be a set of comma separated key/value pairs so that additional info can be passed there in the future.

On build a CONFDIR prefix can be specified as:
CONFDIR=/etc cargo build

Resolves: #100


src/lib.rs
```diff
@@ -415,83 +386,174 @@ fn finalize_fips_approval(
...
+                            }
+                        };
+                    } else {
+                        match dbv[0] {
+                            "sqlite" => {
+                                slot.dbtype =
+                                    Some(storage::SQLITEDB.to_string());
+                                slot.dbpath = Some(dbv[1].to_string());
+                            }
+                            "jsondb" => {
```
Jakuje, line 469: should this be json or jsondb? As for now, all the other, sqlite and memory have the same name as their `storage::` counterparts, but json has different. Not sure if this is overlook, but having them in sync would make it easier to orient in that.
Some examples/documentation of the configuration file syntax, options and values would also be good to be able to understand how the conf files should look like and be constructed.

simo5: I should change this to json indeed. I am preparing another PR based on this one that also brought this up to my attention.

src/tests/token.rs
```diff
@@ -316,3 +326,66 @@ fn test_interface_invalid_version() {
...
+            format!("{}/{}", TESTDIR, "test_config_multiple.json"),
+            "TOKEN 2",
+        ),
+    ];
+    let mut config = String::new();
+    let mut tokens = Vec::<TestToken>::new();
+    for db in &dbs {
+        let mut token = TestToken::new(db.1.clone(), false);
+        token.setup_db(None);
+        /* here we hand code a config file.
```
Jakuje, line 351:
```suggestion
        /* here we hard code a config file.
```

src/lib.rs
```diff
@@ -415,83 +386,187 @@ fn finalize_fips_approval(
...
+        let conf = match Config::from_file(&filename) {
             Ok(n) => n,
-            Err(_) => return CKR_ARGUMENTS_BAD,
+            Err(_) => {
+                let mut gc = Self::empty_config();
+                /* backwards compatibility where we used to only specify
+                 * a file, this does not support all older options, just
+                 * the more common one of specifying a .sql file with no
+                 * slot specification. */
+                if filename.ends_with(".sql") {
```
Jakuje, line 416: At least for opensc, the KRYOPTIC_CONF had both db file and slot number (which matches the init args semantics) so this condition wont hit if I see right with the example from OpenSC:
https://github.com/OpenSC/OpenSC/blob/f7070f4c9f9c6e1cb8d2efe838f73d54c78553da/src/tests/p11test/runtest.sh#L128
But we do not have any automation there yet so I can change it in OpenSC tests.

I see the pkcs11-provider uses just the path to the db though:
https://github.com/latchset/pkcs11-provider/blob/main/tests/kryoptic-init.sh#L30

Jakuje (approved):

src/lib.rs
simo5: Passing in a slot number that way was always mant to be just for testing, I should have wrapped that syntax in #[cfg(test)] indeed.

You should definitely change opensc to use the db=sqlite:/path/to/file.sql,slot=1 or maybe use a config file.

Speaking of which, one question I had about this PR is whether we should always pass parameters this way or if we should also have a way to just pass in a config file name to read from like: config=/path/to/config/file.conf

Jakuje:
regarding the configuration file format, I have just few thoughts:
 * the type + path sounds to me like a duplicated information as the type could be derived from the filename (except for the memory, but that is special). It might also not work that well when we will support NSS DB, which will likely be just a path to a directory? So this is just a though rather than something to change.
 * In softhsm, I was adding a support to configure allowed mechanism through configuration file mostly for testing purposes to be able to emulate a pkcs11 module which does not support for example RSA-PSS, raw RSA or something else. But this can be added later if needed (extending the configuration file with more non-mandatory keys should not be an issue).

simo5:
> regarding the configuration file format, I have just few thoughts:
> 
>     * the type + path sounds to me like a duplicated information as the type could be derived from the filename (except for the memory, but that is special). It might also not work that well when we will support NSS DB, which will likely be just a path to a directory? So this is just a though rather than something to change.

I did not want to assume the URI could always tell, and wanted a preamble exactly due to nssdb. As you know I am working on it, and the config structure of NSS is completely different. I have an heuristic to check if the arguments passed in pReserved are NSSDB vs our format and then have completely separate parsing routines ...

Also while NSSDB has the complete set of options in their parser, the standard config for kryoptic cannot be fully represented on the arguments, beyond the basic db+slot argument you have to pass in a configuration file to define anything else.

So the questions is ... should we just support the fallback to just the uri for legacy/expedenciy purposes and really only have a kryoptic_conf=<filename> parameter support instead for anything more complicated including the need to specify a slot number ?

On one hand that would require writing out a configuration file even  for simple stuff. OTOH I wonder if most admins will rather have a config file written somewhere than a long weird configuration string that admits no spaces and stuff.

I am willing to change this code radically if needed.

>     * In softhsm, I was adding a support to configure allowed mechanism through configuration file mostly for testing purposes to be able to emulate a pkcs11 module which does not support for example RSA-PSS, raw RSA or something else. But this can be added later if needed (extending the configuration file with more non-mandatory keys should not be an issue).

Yes, I have no issues adding more config options in the file, I would not add them to the argument parsing though, hence see above question.

Jakuje:
> So the questions is ... should we just support the fallback to just the uri for legacy/expedenciy purposes and really only have a kryoptic_conf= parameter support instead for anything more complicated including the need to specify a slot number ?

I would say so. The configuration file is nice for anything more complicated than "I just want a pkcs11 token working" -- for this, having just env variable pointing to the sql file should be enough.

> Yes, I have no issues adding more config options in the file, I would not add them to the argument parsing though, hence see above question.

Agree. I think this is not urgent. I just wanted to point it out, that this was very helpful when we did introduce support for PSS and wanted to test this in different libraries against pksc11. But I do not think it is needed now.

simo5:
Ok, so based on recent comments and discussion I changed and simplified the code.
Now the only allowed arguments in the reserved C_initialized argument are:
1) kryoptic_conf=/path/to/config.file
2) [dbtype:]/path/to/database.file (ex: json:/file.json or /db.sql) 

2, is for quick adding a database with defaults otherwise, and support the simplest case available in the previous code (CI compatibility).

A slot is automatically assigned for legacy files

simo5:
I also had to squash together the test and code commits, because this changed forced me to change tests at the same time due to less backwards compatibility with the testing infrastructure

simo5:
One more push where I moved all config related stuff in src/config.rs where it makes more sense at this point

PR #103, opened 2024-10-09
Refactor ossl/fips code to use modules

simo5:
Using includes was a useful hack to get started, but now that the code is more stable move to use modules instead of includes whenever possible.


simo5:
This PR just shifts around code, but has not added or changed any functionality.
Given CI is happy I feel confident pushing it w/o second review so I can rebase on top the other PRs still open.

PR #104, opened 2024-10-14
Refactor attribute code to use methods

simo5:
Instead of using functions, implement methods on strcuts and enums. This looks more ergonomic and gives a better way to import the objects as well as a clear signal of what the output of the function actually is.

### DIFF
--- a/src/aes.rs
+++ b/src/aes.rs
@@ -3,8 +3,7 @@
 
 use std::fmt::Debug;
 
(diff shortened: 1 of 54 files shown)


simo5:
This is again a mostly cosmetic refactoring to improve the ergonomics of the code, but not substantial changes, so I will once again self-approve this one.

Issue #105, opened 2024-10-21
It would be nice if the configuration file and its directives was documented in README.md


simo5:
It would be nice if the configuration file and its directives was documented in README.md

_Originally posted by @rcritten in https://github.com/latchset/kryoptic/issues/100#issuecomment-2427527607_

simo5:
May end up adding a doc string to the config crate.
Then you'd be able to get the info after cargo doc is generated

PR #106, opened 2024-10-23
Refactor Storage Layer

simo5:
This splits out the storage spefcific knowledge from the token to the actual storage backend.

The storage backend is restructured so that multiple storage types can use a common information organizational schema for non standardized objects like user pins and token info and non standardized authenticationand encryption layers for the actual data.

This is in preparation for adding code to read databases written by other software tokens that follow different storage philosophies.

While woring on this refactoring a few minor issues with how pin/login were handled was found and have been fixed.

Additionally the code now uses a common Key encryption Key that both the Ueer and SO Pins can unlock allowing recovery of the database via SO pin if the user pin is lost. This means the SO pin needs to be set to a non-default value on initialization to maintain the confidentaility of the token.
It can be set to a random value if recovery is never desired.


simo5:
Note, this commit also drops the special use code available only in tests that was disabling encryption as it turns out that now that we do not decrypt every object on search  but only on explicit fetch the test are not that slow anymore.

simo5:
Let me know if you think this is too slow:
```
$ cargo test --features slow
[..]
test result: ok. 176 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 258.00s
```
I think it is fine given:
```
$ cargo test
[..]
test result: ok. 176 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.19s
```

src/storage/json.rs
```diff
@@ -1,210 +1,87 @@
...
     }
 }
 
-pub fn json() -> Box<dyn Storage> {
-    Box::new(JsonStorage {
-        filename: String::from(""),
-        cache: memory::memory(),
-    })
+#[derive(Debug)]
+pub struct JsonDBInfo {
```
Jakuje, line 55: just a random though, I think there should be some warning that the json db is not encrypted, which might not be obvious on the first sight. Not sure whether this would be good place to have it in the source code comments or start writing some manual page or just a paragraph in readme describing the current format/configuration.
I will probably open a follow-up issue or MR with the draft.

Jakuje (approved):

src/storage/json.rs
Jakuje: Actually the documentation is tracked as #105, but the storage differences and limitations should be mentioned somewhere somehow.

simo5: Frankly, my intention is to eventually drop support for jsondb, the only reason I keep it is that the code is used to load some of the testing data, and allows for some testing of different storage layers.

But if you look at the latest patch it is optional and I have not changed CI to build with jsondb enabled ...

src/config.rs
```diff
@@ -129,59 +129,98 @@ impl Config {
...
+        let mut slotnum: u32 = 0;
+        /* if there are any slot missing a valid slot number
+         * we are going to allocate slots numbers after the highest
+         * properly configured one. Note that the config file format
+         * requires slot numbers, so this generally happens for legacy
+         * or init args configurations only, ie a single slot */
+        let mut missing = false;
+        for slot in &self.slots {
+            if slot.slot != u32::MAX {
+                if slotnum <= slot.slot {
```
Jakuje, line 150: just a nit that this condition is awkward to read and I would naturally like to read the condition the other way round as, where I compare the "iterator" variable to some other value, not the other way round:
```
                if slot.slot >= slotnum  {
```
This also sounds to me like the test `if (0 == function(...))`.

src/config.rs
```diff
@@ -129,59 +129,98 @@ impl Config {
...
+                    slotnum += slot.slot + 1;
```
Jakuje, line 151: I think this should be just
```
                    slotnum = slot.slot + 1;
```
otherwise you do not use the first unused, but sum of all unused + some.

src/config.rs
simo5: My reading is reversed, here the "subject" is slotnum, and I am comparing it to the many slot.slot that we are parsing through to increment it as needed ...

Jakuje (approved):
lgtm

simo5:
Thank you!

PR #107, opened 2024-10-29
bindings: Create 32b compatible bindings

Jakuje:
It turns out the bindgen assumes 64b architecture by default when converting constants, which causes warnings like
```
warning: literal out of range for `u32`
 --> src/pkcs11/bindings.rs:7:50
  |
7 | pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = 18446744073709551615;
  |                                                  ^^^^^^^^^^^^^^^^^^^^
  |
  = note: the literal `18446744073709551615` does not fit into the type `u32` whose range is `0..=4294967295`
  = note: `#[warn(overflowing_literals)]` on by default
```
but more important, most of the tests keep failing as the comparisons around this value wont work.

Rather than inventing a way to convert the constant to c-type ulong, I thought setting specific value known as to be the MAX would be more obvious.

Builds run in rawhide now:

https://copr.fedorainfracloud.org/coprs/jjelen/kryoptic/build/8186035/

Unfortunately, this did not fix it and we are still getting error during intialization with:
```
---- tests::token::test_token_sql stdout ----
thread 'tests::token::test_token_sql' panicked at src/tests/mod.rs:97:43:
called `Result::unwrap()` on an `Err` value: Error { kind: CkError, origin: None, errmsg: None, reqsize: 0, ckrv: 19 }
```
(`CKR_ATTRIBUTE_VALUE_INVALID`), which sounds like some attributes pulled from DB have different bit size than expected by the 32b build. Not sure if there is something in #106, which might fix it -- I will try after it will get merged.

The Fedora 41 builds seems to be failing for unrelated reasons. Sounds like the bindings were not generated correctly and the types with blocklisted prefixes are not generated as they should on the Fedora 41 version of bindgen? -- still investigating.

### DIFF
--- a/src/pkcs11/interface.rs
+++ b/src/pkcs11/interface.rs
@@ -6,6 +6,9 @@ include!("bindings.rs");
 // types that need different mutability than bindgen provides
 pub type CK_FUNCTION_LIST_PTR = *const CK_FUNCTION_LIST;
 pub type CK_FUNCTION_LIST_3_0_PTR = *const CK_FUNCTION_LIST_3_0;
+// this is wrongly converted on 32b architecture to too large value
+// which can not be represented in CK_ULONG.
+pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = CK_ULONG::MAX;
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
 pub struct CK_INTERFACE {
--- a/build.rs
+++ b/build.rs
@@ -162,6 +162,7 @@ fn main() {
         .blocklist_type("CK_FUNCTION_LIST_PTR")
         .blocklist_type("CK_FUNCTION_LIST_3_0_PTR")
         .blocklist_type("CK_INTERFACE")
+        .blocklist_var("CK_UNAVAILABLE_INFORMATION")
         .parse_callbacks(Box::new(Pkcs11Callbacks))
         .generate()
         .expect("Unable to generate bindings")


src/pkcs11/interface.rs
```diff
@@ -6,6 +6,14 @@ include!("bindings.rs");
...
 pub type CK_FUNCTION_LIST_PTR = *const CK_FUNCTION_LIST;
 pub type CK_FUNCTION_LIST_3_0_PTR = *const CK_FUNCTION_LIST_3_0;
+// this is wrongly converted on 32b architecture to too large value
+// which can not be represented in CK_ULONG. Docs says it can't have
+// any other value so enumerate them here:
+// https://doc.rust-lang.org/std/os/raw/type.c_ulong.html
+#[cfg(target_pointer_width = "64")]
+pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = u64::MAX;
+#[cfg(target_pointer_width = "32")]
+pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = u32::MAX;
```
simo5, line 16: have you tried to just use CK_ULONG::MAX ?
Or maybe (-1 as CK_ULONG) ?

simo5: To be clear, I think on windows unsigned long is 32 bit while pointer is 64 bit, which is why your test is not correct.

Jakuje: Yes, the warning is gone, but the other issues are still present.

simo5:
@Jakuje given you are updating bindgen related stuff, would you mind updating to bindgen 0.70 (desirable as it drops a bunch of dependencies), and see if the problem persists with 0.70?

If it does your solution is ok with me for now.

simo5 (changes requested):
See comment above

Jakuje:
> @Jakuje given you are updating bindgen related stuff, would you mind updating to bindgen 0.70 (desirable as it drops a bunch of dependencies), and see if the problem persists with 0.70?

This might be an issue for the Fedora packaging, as the Fedora package is still on 0.69, see https://src.fedoraproject.org/rpms/rust-bindgen.

Anyway, I opened rust-lang/rust-bindgen#2965 as it reproduces with both 0.69 and 0.70. I would prefer to keep it as proposed for now. Maybe I will just add a comment to the commit/code to make sure to remove the workaround as we will have fix.

simo5 (approved):

simo5:
Thanks, this works for now!

PR #113, opened 2024-11-12
Fix issues when compiling on i686

simo5:
This should address all remaining issues found by @Jakuje when trying to build on Fedora on the i686 arch.


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [ossl3, fips, release, dynamic]
+        name: [ossl3, fips, release, dynamici, i686]
```
Jakuje, line 17:
```suggestion
        name: [ossl3, fips, release, dynamic, i686]
```

src/ossl/aes.rs
```diff
@@ -276,8 +276,7 @@ impl AesOperation {
                 let l = 15 - params.ulNonceLen;
                 if params.ulDataLen == 0
                     || params.ulDataLen > (1 << (8 * l))
-                    || (params.ulDataLen + params.ulMACLen)
-                        > CK_ULONG::try_from(u64::MAX)?
+                    || params.ulDataLen > CK_ULONG::MAX - params.ulMACLen
```
Jakuje, line 279: I assume the operator precedence in rust is sensible (unlike in some other insane languagues), but I would prefer using braces around the expression to be explicit here:
```suggestion
                    || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
```
same in the next chunk

.github/workflows/build.yml
```diff
@@ -100,6 +104,9 @@ jobs:
           if [ "${{ matrix.name }}" = "dynamic" ]; then
             cargo test --release --features dynamic
           fi
+          if [ "${{ matrix.name }}" = "i686" ]; then
+            cargo test --target i686-unknown-linux-gnu --features dynamic
+          fi
```
Jakuje, line 109: I think we should add also an appropriate condition to the `Build` stage. Right now it works (unlike the `dynamici` case which passes without executing anything) as the build is happening in the `Test` stage, but it makes the output a bit confusing.

I would probably suggest to make the `if` chain more explicit so the typos like this can get caugth with something like:
```
if X; then
 ...
elif Y; then
  ...
else
   fail unknown
fi
```

src/ossl/aes.rs
simo5: the compiler will complain of unneeded parenthesis if I do that, remember I do not establish the rule, `make fix-format` and `rustc` reigns here.

.github/workflows/build.yml
simo5: build and test are frankly mostly redundant, they only help to distinguish at a glance if the issue is in the code or in the tests themselves, this i why for i686 I intentionally only build tests, as that covers all.

Jakuje (approved):

PR #116, opened 2024-11-20
Refactor and Centralize ECC helpers

### LINKED ISSUE #112: Make key objects of carying type-specific methods
Until #111, every key type had 1-n mapping to one or more mechanisms, which could encapsulate all operations on the given key type.

The #111 introduced EC_MONTGOMERY keys,  that share the ECDH mechanism with the standard EC(DSA) key objects. The ECDH Operation would be more readable, if we could convert the key object to pkey and ask the object itself for the output length, instead of calling the key specific functions from different modules.

              Instead of hardcoding this here I think we should dynamically check at run time (by looking at the mechanisms available perhaps during registration) whether montgomery curves are available.

These functions should probably be exported by the relative mechanism and the mechanism should be sourced via the key type.

We may need to add a new trait to key objects that can be used to extract this info in a more abstract way.
Ie key.get_key_len() or key.get_field_len() or key.get_modulo_len() or some such.

The trait can return Err(CKR_KEY_TYPE_INCONSISTENT)? in the default implementation and only keys that can return a reasonable value will implement this.

If you think it is too much for this PR pleae file an issue and we can make a follow-up later.

_Originally posted by @simo5 in https://github.com/latchset/kryoptic/pull/111#discussion_r1833420662_

simo5:
This PR reworks the various ECC files to use common routines to simplify and make uniform, as much as possible all ECC helper code.

Fixes #112

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,34 +78,34 @@ jobs:
       - name: Build
         run: |
           if [ "${{ matrix.name }}" = "fips" ]; then
-            cargo build -vv --no-default-features --features fips
+            cargo build -vv --features fips
           fi
           if [ "${{ matrix.name }}" = "ossl3" ]; then
-            cargo build -vv
+            cargo build -vv --features standard
           fi
           if [ "${{ matrix.name }}" = "release" ]; then
-            cargo build -vv --release
+            cargo build -vv --release --features standard
           fi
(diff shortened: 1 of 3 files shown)


Jakuje (approved):
did not notice anything odd. Looks good!

simo5:
Thanks!

PR #119, opened 2024-11-22
Ensure ECC key attributes are aligned with the PKCS#11 Spec

### LINKED ISSUE #115: Review and fix once and for all CKA_EC_POINTand CKA_VALUE for EC curves
The code is confused and in some place contradictory on what the encoding should be for CK_ECC or CK_EC_MONTGOMERY objects. Consolidate all handling in helper functions and add reefrences to the relevant standards that establish what the encoding should really be.

simo5:
The different ECC key types accept values encoded in different ways.

Add code to ensure values are correctly formed on import and generation according to what the spec prescribes.

Fixes #115

### DIFF
--- a/src/ec/eddsa.rs
+++ b/src/ec/eddsa.rs
@@ -68,24 +68,13 @@ impl ObjectFactory for EDDSAPubFactory {
         /* According to PKCS#11 v3.1 6.3.5:
          * CKA_EC_POINT, Byte array,
          * Public key bytes in little endian order as defined in RFC 8032 */
-        let point = get_ec_point_from_obj(&obj).map_err(|e| {
+        check_ec_point_from_obj(&oid, &mut obj).map_err(|e| {
             if e.attr_not_found() {
                 Error::ck_rv_from_error(CKR_TEMPLATE_INCOMPLETE, e)
-            } else if e.rv() != CKR_ATTRIBUTE_VALUE_INVALID {
-                Error::ck_rv_from_error(CKR_ATTRIBUTE_VALUE_INVALID, e)
             } else {
-                general_error(e)
+                e
             }
         })?;
-        let rawsize = ec_point_size(&oid)?;
-        if point.len() != rawsize {
-            /* For compatibility with applications that use DER encoding */
-            if point.len() == rawsize + 2 {
-                convert_ec_point_from_30(&mut obj)?;
-            } else {
-                return Err(CKR_ATTRIBUTE_VALUE_INVALID)?;
-            }
-        }
 
         Ok(obj)
     }
--- a/src/ec/mod.rs
+++ b/src/ec/mod.rs
@@ -134,17 +134,6 @@ pub fn get_ec_point_from_obj(key: &Object) -> Result<Vec<u8>> {
     Ok(octet.to_vec())
 }
 
-pub fn convert_ec_point_from_30(key: &mut Object) -> Result<()> {
-    match key.get_attr_as_ulong(CKA_KEY_TYPE).map_err(general_error)? {
-        CKK_EC_EDWARDS | CKK_EC_MONTGOMERY => (),
-        _ => return Err(CKR_GENERAL_ERROR)?,
-    }
-    let point = key.get_attr_as_bytes(CKA_EC_POINT)?;
-    let decoded = asn1::parse_single::<&[u8]>(point).map_err(device_error)?;
-    key.set_attr(Attribute::from_bytes(CKA_EC_POINT, decoded.to_vec()))
-        .map_err(general_error)
-}
-
 #[cfg(test)]
 pub fn curvename_to_key_size(name: &str) -> Result<usize> {
     ec_key_size(&curvename_to_oid(name)?)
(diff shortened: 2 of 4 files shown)


simo5:
Addressing one key at a time, first patch addressed CKK_EC Public key, other to come as I implement and add test.

src/ec/mod.rs
```diff
@@ -47,6 +47,40 @@ pub const EDWARDS448: &str = "edwards448";
 pub const CURVE25519: &str = "curve25519";
 pub const CURVE448: &str = "curve448";
 
+fn bits_to_bytes(bits: usize) -> usize {
+    (bits + 7) / 8
+}
+
+fn ec_ws_point_size(bits: usize) -> usize {
```
Jakuje, line 54: in the function name, its not clear to me what the `ws` means on the first read so I would propose to either choose a better name or do explaining comment.

Jakuje (approved):
the formatter is not happy though.

src/ec/mod.rs
simo5: changed ws ->secp

simo5:
This PR is ready for review, however, being spec compliant means a lot of tools will just fail to deal with Edwaurds and Montgomery public curves as they expect a DER encoded OCTET STRING for the public EC_POINT for these curves.

However a DER OCTET STRING is incorrect even if we look at the 3.0 spec that specified a DER BIT STRING vs the 3.1 spec which specifies just a raw byte buffer.

Fixing all tools to expect both encodings will take a while ...

Jakuje (approved):

src/ec/mod.rs
```diff
@@ -75,28 +102,32 @@ fn curvename_to_oid(name: &str) -> Result<asn1::ObjectIdentifier> {
...
-    let point = key.get_attr_as_bytes(CKA_EC_POINT).map_err(general_error)?;
-    /* [u8] is an octet string for the asn1 library */
-    let octet = asn1::parse_single::<&[u8]>(point).map_err(device_error)?;
+    let point = key.get_attr_as_bytes(CKA_EC_POINT)?;
+    let octet = match key.get_attr_as_ulong(CKA_KEY_TYPE)? {
+        CKK_EC => {
+            /* [u8] is an octet string for the asn1 library */
+            asn1::parse_single::<&[u8]>(point).map_err(device_error)?
+        }
+        CKK_EC_EDWARDS | CKK_EC_MONTGOMERY => point.as_slice(),
```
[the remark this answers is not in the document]
simo5: Yeah,
so I can easily add support for allowing to get 3.0 inputs, but the big issue is the outputs. Applications that expect only DER formatted EC_POINT will fail.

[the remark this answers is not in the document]
Jakuje: Right. Do we have, right now, any application that accepts non-DER encoded EC_POINT? So should we do some quirks or configuration for this to not break the whole world, but not to do bad thing by default?

simo5:
So the last commit adds a feature to enable 3.1 compatibility and handles 3.0 compatibility as default.

I am not entirely sure I like it this way, most applications and most tokens will probably fail to deal with this appropriately anyway, so perhaps we should have a runtime configuration switch rather than a build time feature. A runtime switch will make it easier to deal with.

The issue though is where to put this switch. One option is the config file, or we could have both a config file and an environment variable ... what a mess...

src/ec/mod.rs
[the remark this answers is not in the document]
simo5: We have pkcs11-provider now, but see the latest commits, I added a way to set this in config or via env var.

src/lib.rs
```diff
@@ -890,8 +911,73 @@ extern "C" fn fn_get_attribute_value(
...
+                    let buf: &mut [u8] = unsafe {
+                        std::slice::from_raw_parts_mut(
+                            a.pValue as *mut u8,
+                            buflen,
+                        )
+                    };
+                    let out = res_or_ret!(point_buf_to_der(buf, bufsize));
+                    if let Some(v) = out {
+                        if v.len() > bufsize {
+                            return CKR_GENERAL_ERROR;
```
Jakuje, line 961: should not this be BUFFER_TOO_SMALL error?

simo5: Not really, this means the DER encoding produced something linger than len + 2, and it means there is an internal bug somewhere.

src/ec/eddsa.rs
```diff
@@ -46,10 +46,46 @@ impl EDDSAPubFactory {
...
+        })?;
+        let rawsize = ec_point_size(&oid)?;
+        if point.len() != rawsize {
+            /* For compatibility with applications that use DER encoding */
+            if point.len() == rawsize + 2 {
+                convert_ec_point_from_30(&mut obj)?;
+            } else {
+                return Err(CKR_ATTRIBUTE_VALUE_INVALID)?;
+            }
+        }
```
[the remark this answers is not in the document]
simo5: Ok, I took this and made one step forward by unifying the check in a single place so we have less duplication.

src/lib.rs
simo5: The case were the buffer is smaller than needed is handled by the internal token routines as usual, as with any other attribute. The only case we need to handle is if we have to DER encode the attribute and we have a buffer that is keysize < buflen < keysize+2 and point_buf_to_der() handles that case.

Jakuje (approved):
Looks good!

Issue #118, opened 2024-11-22
Extend test coverage matrix


Jakuje:
> Given how the complexity of the possible features expands, I would probably propose to add some more targets, such as `minimal` with minimal set of features to make sure all tests and uses of conditional mechanisms are correctly gated. Then maybe minimal + nssdb or something similar. Having full matrix of all combinations does not make sense, I think.

_Originally posted by @Jakuje in https://github.com/latchset/kryoptic/pull/110#discussion_r1854474036_

PR #122, opened 2024-11-28
Extend test matrix to cover more possible build combinations of features

Jakuje:
With the amount of features the kryoptic supports, keeping track that they are self contained and all of the dependencies is quite tedious work.

This extends the test matrix and updates the tests to be able to run with different storages. For now, `jsondb`, `sqlitedb` and `nssdb` work, but the `memorydb` does not. Additionally, I encountered the issue that the memory db is not possible to provision, unless encryption is turned on, because it tries to find PIN counters, which are not in non-encrypted storage.

Fixes: #118


Jakuje:
This is mostly ready for review. The only thing to figure out is the issue with the memorydb and why the jsondb fails in CI, but works locally.
If needed, changes can be merged and CI changes could land later after we will figure out last details.

src/tests/aes.rs
```diff
@@ -500,14 +514,10 @@ fn test_aes_operations() {
...
-            Err(ret) => return assert_eq!(ret, CKR_OK),
-        };
-        let plaintext = match get_test_data(session, testname, "plaintext") {
-            Ok(vec) => vec,
-            Err(ret) => return assert_eq!(ret, CKR_OK),
-        };
+        let ciphertext = hex::decode("4154c0be71072945d8156f5f046d198d")
+            .expect("Failed to decode ciphertext");
+        let plaintext = hex::decode("8b2b1b22f733ac09d1196d6be6a87a72")
+            .expect("Failed to decode plaintext");
```
simo5, line 520: What about turning get_test_data to be a function like you did for get_gcm_test_data() ?
It can be a function that return data from an array of test data base on an index or something.
Also in the commit message CKA_DATA -> CKO_DATA

src/tests/init.rs
```diff
@@ -41,7 +41,12 @@ fn test_init_token() {
         pin_value.len() as CK_ULONG,
         std::ptr::null_mut(),
     );
-    assert_eq!(ret, CKR_PIN_INCORRECT);
+    if testtokn.dbtype == "nssdb" {
```
simo5, line 44: I would skip the fn_init_otken entirely if no nssdb, rather than matching just he result ?

src/tests/login.rs
```diff
@@ -82,7 +82,12 @@ fn test_login() {
     let mut token_info = CK_TOKEN_INFO::default();
     let ret = fn_get_token_info(testtokn.get_slot(), &mut token_info);
     assert_eq!(ret, CKR_OK);
-    assert_eq!(token_info.flags & pin_flags_mask, CKF_USER_PIN_COUNT_LOW);
+    if testtokn.dbtype == "nssdb" {
```
simo5, line 85: Here too, I would just skip the operation, rather than matching to a fixed value.

src/tests/nssdb.rs
```diff
@@ -175,6 +175,9 @@ fn test_nssdb_token() {
     let ret = fn_find_objects_final(session);
     assert_eq!(ret, CKR_OK);
 
+    #[cfg(not(feature = "rsa"))]
```
simo5, line 178: Given nssdb always requires HMAC, what about changing some of these operations to use/generate a SECRET value usable by HMAC and defer RSA testing to the general RSA testing?

RSA is not really key here, what is ey is testing that encryption.decryption of a secret attribute works and fetching works as expected.

simo5 (approved):
Looks largely ok, only minor nits.
Thanks this is exactly where I wanted to see things headed!

.github/workflows/build.yml
```diff
@@ -75,39 +76,41 @@ jobs:
...
-          fi
-          if [ "${{ matrix.name }}" = "dynamic" ]; then
-            cargo build -vv --release --features standard,dynamic
+          FEATURES=""
+          OPTS=""
+
+          if [ "${{ matrix.db }}" = "memorydb" ]; then
+            FEATURES="memorydb"
+          elif [ "${{ matrix.db }}" = "jsondb" ]; then
+            FEATURES="memorydb"
```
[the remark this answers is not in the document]
Jakuje: Indeed! Fixing.

.github/workflows/build.yml
```diff
@@ -75,39 +76,41 @@ jobs:
...
-          fi
-          if [ "${{ matrix.name }}" = "release" ]; then
-            cargo build -vv --release --features standard
-          fi
-          if [ "${{ matrix.name }}" = "dynamic" ]; then
-            cargo build -vv --release --features standard,dynamic
+          FEATURES=""
+          OPTS=""
+
+          if [ "${{ matrix.db }}" = "memorydb" ]; then
```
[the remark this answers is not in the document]
Jakuje: Indeed. Did not think about that initially. Changing

src/tests/mod.rs
```diff
@@ -129,8 +140,8 @@ impl TestToken<'_> {
         file.write(data.as_bytes()).unwrap();
     }
 
-    fn make_init_string(&self) -> String {
-        let confname = format!("{}.conf", self.filename);
+    fn make_init_string(&mut self) -> String {
```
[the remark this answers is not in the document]
Jakuje: Sounds like this is no longer needed and is some leftover from some initial attempts.

src/tests/mod.rs
```diff
@@ -167,13 +178,40 @@ impl TestToken<'_> {
...
 
+    #[allow(unreachable_code)]
+    fn get_default_db(name: &str) -> (&'static str, String) {
+        use std::fmt::Write;
+
+        /* in order of preference, the first wins */
+        #[cfg(feature = "nssdb")]
+        {
+            let ret = format!("configDir={}/{}/", TESTDIR, name);
+            return ("nssdb", ret);
```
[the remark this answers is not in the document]
Jakuje: Good suggestion. Changed.

src/tests/objects.rs
```diff
@@ -115,17 +115,19 @@ fn test_create_objects() {
...
-        CKO_CERTIFICATE,
-        &[(CKA_CERTIFICATE_TYPE, CKC_X_509)],
-        &[
-            (CKA_CHECK_VALUE, "ignored".as_bytes()),
-            (CKA_SUBJECT, "subject".as_bytes()),
-            (CKA_VALUE, "value".as_bytes())
-        ],
-        &[(CKA_TOKEN, true), (CKA_TRUSTED, false)],
-    ));
+    if testtokn.dbtype != "nssdb" {
```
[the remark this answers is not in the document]
Jakuje: Probably misread. Sorry. It should have been on the DATA import if I see right.

src/tests/aes.rs
Jakuje: I did not like either approach. With the `get_test_data`, the hex blocks get very indented, but huge blobs inside of function make the reading harder. In case of AES, the values are quite short so I considered this ok.

src/tests/init.rs
Jakuje: Fine by me. I though keeping the test around and checking return value of the nssdb will make sure we do not regress even in these corner cases, but skipping is probably ok. Changing.

Jakuje:
> The idea is to be comprehensive in the coverage, or just have at least some test coverage?
> 
> because I has some experience with [all-pairs testing](https://en.wikipedia.org/wiki/All-pairs_testing) and extension of it to triplets, quadruplets, etc....

Right now, we have around 25 features:

https://github.com/latchset/kryoptic/blob/main/Cargo.toml#L49

They have some internal dependencies and then there is the "default" feature which we want or dont want to build. I was worried that doing all pairs (or triplets) would explode or would be even more work to exclude meaningless ones.

tomato42:
> > The idea is to be comprehensive in the coverage, or just have at least some test coverage?
> > because I has some experience with [all-pairs testing](https://en.wikipedia.org/wiki/All-pairs_testing) and extension of it to triplets, quadruplets, etc....
> 
> Right now, we have around 25 features:
> 
> https://github.com/latchset/kryoptic/blob/main/Cargo.toml#L49
> 
> They have some internal dependencies and then there is the "default" feature which we want or dont want to build. I was worried that doing all pairs (or triplets) would explode or would be even more work to exclude meaningless ones.

all pairs (n-tuples) testing is especially effective if you have a lot of features, as then you avoid the combinatorial explosion

and no, having meaningless combinations is not a problem, when you generate the combinations you can exclude combinations that are either impossible to support (like testing ECDH on build without ECC) or combinations we explicitly don't want to support (for whatever reason)

src/tests/objects.rs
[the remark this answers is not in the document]
Jakuje: Ok, this was just in combination with FIPS mode for some reason
```
thread 'tests::objects::test_create_objects' panicked at src/tests/objects.rs:123:13:
```
I will have to dig into this deeper to figure out why it fails only in FIPS + NSSDB

simo5:
> > > The idea is to be comprehensive in the coverage, or just have at least some test coverage?
> > > because I has some experience with [all-pairs testing](https://en.wikipedia.org/wiki/All-pairs_testing) and extension of it to triplets, quadruplets, etc....
> > 
> > 
> > Right now, we have around 25 features:
> > https://github.com/latchset/kryoptic/blob/main/Cargo.toml#L49
> > They have some internal dependencies and then there is the "default" feature which we want or dont want to build. I was worried that doing all pairs (or triplets) would explode or would be even more work to exclude meaningless ones.
> 
> all pairs (n-tuples) testing is especially effective if you have a lot of features, as then you avoid the combinatorial explosion
> 
> and no, having meaningless combinations is not a problem, when you generate the combinations you can exclude combinations that are either impossible to support (like testing ECDH on build without ECC) or combinations we explicitly don't want to support (for whatever reason)

In most cases each algorithm is fully independent from each other algorithm, so that testing them in pairs makes little sense.

We have a core that is always involved and therefore is always tested.

Then we have the algorithms which just operate on specific data, regardless of the core or anything else.

The only major variability is the databases, because what they contain may influence the result of specific tests.

So perhaps testing each "true leaf" algorithm on their own with each database may make some sense. Mostly to ensure there is no unexpected dependency between algorithms, but that is rather simple to verify so I would not want to spend a lot of CPU on that, and definitely not make CI much slower due to this.

Perhaps we can have matrix testing done once a week on a scheduled CI run just for completeness.

src/tests/objects.rs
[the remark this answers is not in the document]
Jakuje: Ok, I tried to debug this with gdb and I have no idea why the combination of NSSDB and FIPS makes this panic (unless some weird optimization is going on there).

If I see right, we never reach the `fn_create_object()`, but the `import_object()` test helper does not do anything suspicious related to NSSDB nor FIPS.

src/tests/nssdb.rs
Jakuje: Indeed, I can do that so we can move forward with this change.

Jakuje:
> warning: method `dbsuffix` is never used

With recent changes the dbsuffix is not used anywhere. I kept it for now because it might be useful in the future, even though it works only for SQL and JSON. If you do not agree, I can remove it altogether.

src/tests/objects.rs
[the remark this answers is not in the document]
simo5: Do you get back a specific error number?
I wonder if the pbkdf2 used to generate the key to encrypt the data fails because the salt is to short or something like that.

src/tests/nssdb.rs
simo5: Up to you.

simo5:
> > warning: method `dbsuffix` is never used
> 
> With recent changes the dbsuffix is not used anywhere. I kept it for now because it might be useful in the future, even though it works only for SQL and JSON. If you do not agree, I can remove it altogether.

We can always resurrect from git, I am ok dropping anything that is not used anymore.

src/tests/objects.rs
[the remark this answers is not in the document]
Jakuje: Not sure, if the 32 is the specific error:
```
thread 'tests::objects::test_create_objects' panicked at src/tests/objects.rs:123:13:
32
```
https://github.com/latchset/kryoptic/actions/runs/12236641232/job/34130555468?pr=122

That would map back to `CKR_DATA_INVALID`. Are there some parts of the certificates encrypted at all?

src/tests/nssdb.rs
Jakuje: I will drop it then. The current solution with if feature + return is ugly.

Jakuje: Or try to move it to RSA test if possible.

src/tests/mod.rs
```diff
@@ -167,13 +179,38 @@ impl TestToken<'_> {
         }
     }
 
+    #[allow(unreachable_code)]
+    fn get_default_db(name: &str) -> (&'static str, String) {
+        /* in order of preference, the first wins */
+        #[cfg(feature = "nssdb")]
+        {
+            let ret = format!("configDir={}/{}/", TESTDIR, name);
+            return (storage::nssdb::DBINFO.dbtype(), ret);
```
simo5, line 188: Do we actually need to get this string by making modules public ?
At the end the dbtype() is nothing else than the db name which is what we are using to select which storage option we are going to fetch this same string from...
I'd like to keep the modules private to the storage tree, and accessory function would work as well if needed()

src/tests/objects.rs
```diff
@@ -107,13 +107,15 @@ fn test_create_objects() {
...
     testtokn.login();
 
-    let _ = ret_or_panic!(import_object(
-        session,
-        CKO_DATA,
-        &[],
-        &byte_values,
-        &bool_values,
-    ));
+    if testtokn.dbtype != "nssdb" {
```
simo5, line 113: This should not be necessary anymore

simo5 (changes requested):
Only a couple of nits, but great job with this, thanks!!

src/tests/objects.rs
Jakuje: The CKO_DATA object type is not posible to store in the NSS DB if I understand it right.

src/tests/init.rs
```diff
@@ -69,7 +69,7 @@ fn test_init_token() {
     );
 
     #[cfg(feature = "fips")]
-    {
+    if testtokn.dbtype != "nssdb" {
```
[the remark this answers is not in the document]
Jakuje: Yes, it is. This attempts to create the CKO_VALIDATION objects, that are not supported by the NSS DB either. Not sure if this is feature or bug though.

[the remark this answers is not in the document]
simo5: Oh right, currently they are synthetic objects in softoken, can't store them.

src/tests/mod.rs
simo5: Yeah by accessor I meant a function exposed in mod.rs ... but no biggie, it is ok as is for now

src/tests/objects.rs
simo5: Doh! Of course.. I spoke with Bob and we could actually store them, but to be able to that the nssdb.rs code will need to be changed, so nothing to do in this PR.

simo5 (approved):

simo5:
Thanks for this work, really appreciated!

PR #123, opened 2024-12-04
build.rs: fix Cargo calls

sahilnxp:
Fixed following compilation error

root@evk:~/kryoptic# CONFDIR=/etc cargo build
   Compiling kryoptic v0.1.0 (/root/kryoptic)
error: unsupported output in build script of `kryoptic v0.1.0 (/root/kryoptic)`: `cargo::rerun-if-changed=pkcs11_headers/3.1/pkcs11.h` Found a `cargo::key=value` build directive which is reserved for future use. Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust. See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -82,7 +82,7 @@ fn build_ossl() {
             providers_path.to_string_lossy()
         );
         println!("cargo:rustc-link-lib=static=fips");
-        println!("cargo::rerun-if-changed={}", libfips);
+        println!("cargo:rerun-if-changed={}", libfips);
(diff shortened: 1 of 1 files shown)


simo5 (commented):
LGTM,
thanks a lot!

simo5 (approved):

PR #124, opened 2024-12-04
Make sqlite database upgradabale and crypto-agile

### LINKED ISSUE #109: Change decryption key storage format
Currently the decryption key stored in the database is assumed to have a specific format and there is no algorithmic agility.
Furthermore individual encrypted entries do not have any indicator about what key generation was used to encrypt them so it is not easy to perform rolling key changes.

We should change the stored wrapped key to have more structure.
An outer layer should indicate what algorithm(or mechanism version) needs to used to unwrap the stored key.

An inner layer should add known data to the payload to ensure that if a mechanism that does not perform authentication of the encryption is used we still have a way to verify we obtained back a proper key and not just random garbage. And it should add a key generation id to be used for decryption.
If the key generations is formed as a UUID then this can be stored on the key object (as CKA_LABEL) and used as the inner data to confirm proper decryption.

For encrypted entries we should add a key generation id to their outer layer so that the correct key can be used.

Finally a way to store multiple key generations should be added, we have two options:
- Change the inner layer of the wrapped keys to hold multiple generations by treating the payload as an array.
- Allow to store multiple entries per user with historic keys.

The latter would allow to more easily purge older key generations.

simo5:
Add metadata to the sqlite database that allows for easier crypto agility as well as allow future upgrades. Move all non-object information to a metadata table.

Fixes #109


src/storage/sqlite.rs
```diff
@@ -368,6 +461,160 @@ impl StorageRaw for SqliteStorage {
...
+            Some(&flags32.to_le_bytes() as &[u8]),
+        )?;
+
+        tx.commit().map_err(bad_storage)
+    }
+
+    fn fetch_user(&self, uid: &str) -> Result<StorageAuthInfo> {
+        let conn = self.conn.lock()?;
+        let mut stmt = conn
+            .prepare("SELECT name, data from meta WHERE ((name=? OR name=? OR name=?) AND value=?)")?;
```
Jakuje, line 549: `name in (?, ?, ?)` SQL would be more appropriate here:

https://www.techonthenet.com/sqlite/in.php

Not sure if the rustqlite have some even better shorthands for the construction of list to make it even simpler.

doc/storage-encryption.md
```diff
@@ -0,0 +1,115 @@
...
+decrypt the storage Encryption key. The Encryption key is versioned and is used
+to encrypt each secret attribute individually by using a key derivation step to
+obtain an object specific key. The Key Derivation algorithms and paramters are
+felxibly defined by wrapping the keys and encrypted values using PKCS#5 like
+ASN.1 structures.
+
+This OID has been defined to identify the HKDF-Expand Key Derivation function:
+1.3.6.1.4.1.2312.20.1.1
+
+KKBPS - Kryoptic Key Based Protection Scheme v1
```
Jakuje, line 29: Every time I read the shortcut `kbps` in lower case I read it as unit of transfer speed kB/s, which is a bit confusing. But I do not have better naming suggestion ...

src/storage/format.rs
```diff
@@ -231,45 +158,24 @@ impl StdStorageFormat {
...
-        let auth_obj = self.aci.make_auth_object(
-            facilities,
-            &get_pin_uid(CKU_SO)?,
-            checked_pin(&[]),
-        )?;
-        self.store.store_obj(auth_obj)
+        let data =
+            self.aci
+                .key_to_user_data(facilities, SO_ID, checked_pin(&[]))?;
+        /* 1 to mark this is the default value */
```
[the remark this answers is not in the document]
simo5: yup

src/storage/sqlite.rs
simo5: no only has parameters, it is a very plain shim on top if libsqlite3

doc/storage-encryption.md
simo5: I know, there isn't much we can do

src/storage/aci.rs
```diff
@@ -50,19 +451,15 @@ impl StorageACI {
...
-        let mut template = CkAttrs::with_capacity(3);
-        template.add_ulong(CKA_CLASS, &class);
-        template.add_ulong(CKA_KEY_TYPE, &keytyp);
-        template.add_ulong(CKA_VALUE_LEN, &keylen);
-        template.add_bool(CKA_EXTRACTABLE, &truebool);
-        let aes = facilities.mechanisms.get(CKM_AES_KEY_GEN)?;
-        self.kek = Some(aes.generate_key(
+        /* Need to use a generic secret here, because the secret key
+         * is used together with HKDF to derive the actual encryption
+         * key and HKDF allows only CKK_HKDf or CKK_GENERIC_SECRET */
```
Jakuje, line 456:
```suggestion
         * key and HKDF allows only CKK_HKDF or CKK_GENERIC_SECRET */
```

Jakuje (approved):
Looks good.
By any chance, did you test this change with pkcs11-provider locally to make sure this works correctly and we wont break the CI?

simo5:
> Looks good. By any chance, did you test this change with pkcs11-provider locally to make sure this works correctly and we wont break the CI?

Yes I always make a run with pkcs11-provider locally to ensure changes actually work in a backwards compatible way.
pkcs11-provider CI more than once caught things.

PR #125, opened 2024-12-09
Fix handing of public objects with nssdb

simo5:
The fetch side was unconditionally applying sensitivity object parsing which meant it would try to decrypt any CKA_VALUE attribute regardless of object type. However the store() function would encrypt it only for the private database objects.

Fix the asymmetry and add a test to ensure writing public object works as expected.


simo5:
@Jakuje this pr should fix the nssdb+fips issue you see in your PR

Jakuje (approved):
lgtm. Came up with the same solution in #122.

PR #126, opened 2024-12-10
nssdb: Fix object search without template

Jakuje:
Enumerating all objects on the token leads to calling search on NSS DB without any search arguments, which resulted in invalid SQL queries. This allows me to move on a bit further with the testing of NSS DB


simo5 (approved):
Obviously correct!

PR #132, opened 2024-12-12
Properly set PIN initialization flag in NSSDB

simo5:
This patch should unblock https://github.com/latchset/pkcs11-provider/pull/487

### DIFF
--- a/src/storage/nssdb/mod.rs
+++ b/src/storage/nssdb/mod.rs
@@ -151,6 +151,14 @@ impl NSSStorage {
         if self.config.password_required {
             info.flags |= CKF_LOGIN_REQUIRED;
         }
+        match self.fetch_password() {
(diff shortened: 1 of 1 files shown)


src/storage/nssdb/mod.rs
```diff
@@ -803,6 +811,10 @@ impl Storage for NSSStorage {
         if ret != CKR_OK {
             return Err(ret)?;
         }
+        /* wnsure we drop the lock here, otherwise we deadlock inside
```
Jakuje, line 814:
```suggestion
        /* ensure we drop the lock here, otherwise we deadlock inside
```

Jakuje (approved):

src/storage/nssdb/mod.rs
simo5: fixed

Jakuje (approved):

PR #133, opened 2024-12-13
Add a caching layer for NSSDB per-attribute keys

### LINKED ISSUE #130: NSSDB Performance issues
The NSSDB is quite slow because each attrib ute is encyrpted with a key that requires PBDF2 with 10000 rounds to derive the encryption key.

NSS "solves" this by caching such keys, so we should probably do the same in nssdb/ci.rs

simo5:
Adds a caching layer to NSSDB to reduce the performance impact of the key derivation operations.
NSSDB uses a particularly costly operation (pbkdf2) to derive a key for each of the attributes being encrypted.

This PR also adds a test to demonstrate the issue.
This test takes a relatively long time to execute without the cache:
```
$ time cargo test --features standard,nssdb test_nssdb_key_cache
   Compiling kryoptic v0.1.0 (/home/remote/ssorce/devel/git/kryoptic)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 4.96s
     Running unittests src/lib.rs (target/debug/deps/kryoptic_pkcs11-0639a8933718ff9b)

running 1 test
test tests::nssdb::test_nssdb_key_cache has been running for over 60 seconds
test tests::nssdb::test_nssdb_key_cache ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 186 filtered out; finished in 110.29s


real	1m55.282s
user	1m57.233s
sys	0m0.649s

```

It is much faster with the cache:
```
$ time cargo test --features standard,nssdb test_nssdb_key_cache
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running unittests src/lib.rs (target/debug/deps/kryoptic_pkcs11-0639a8933718ff9b)

running 1 test
test tests::nssdb::test_nssdb_key_cache ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 186 filtered out; finished in 0.11s


real	0m0.166s
user	0m0.144s
sys	0m0.020s

```

Fixes #130

### DIFF
--- a/src/storage/nssdb/ci.rs
+++ b/src/storage/nssdb/ci.rs
@@ -84,11 +84,17 @@ impl KeysWithCaching {
     }
 
     pub fn set_key(&mut self, key: Vec<u8>) {
+        if let Some(ref mut oldkey) = &mut self.enckey {
+            oldkey.zeroize();
+        }
         self.enckey = Some(key);
     }
 
     pub fn unset_key(&mut self) {
-        self.enckey = None;
+        if let Some(ref mut key) = &mut self.enckey {
+            key.zeroize();
+            self.enckey = None;
+        }
     }
 
     fn get_cached_key(&self, id: &[u8; SHA256_LEN]) -> Option<LockedKey> {
--- a/src/storage/nssdb/ci.rs
+++ b/src/storage/nssdb/ci.rs
@@ -1,6 +1,9 @@
 // Copyright 2024 Simo Sorce
 // See LICENSE.txt file for terms
 
+use std::collections::BTreeMap;
+use std::sync::{RwLock, RwLockReadGuard};
+
 use crate::attribute::CkAttrs;
 use crate::error::Result;
 use crate::interface::*;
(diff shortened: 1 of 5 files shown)


simo5:
Added a patch to constraint the size of the cache, otherwise it could grow unbounded.
I chose an arbitrary limit of 128 keys. The logic can be improved later if it turns out there is some use case that can consistently cause thrashing

simo5:
Immortalizing an epic failure:
```
---- tests::login::test_login stdout ----
thread 'tests::login::test_login' panicked at src/tests/login.rs:78:9:
assertion `left != right` failed
  left: 0
 right: 0
```

src/storage/nssdb/ci.rs
```diff
@@ -12,6 +15,124 @@ use crate::token::TokenFacilities;
...
+            None => false,
+        }
+    }
+
+    pub fn set_key(&mut self, key: Vec<u8>) {
+        self.enckey = Some(key);
+    }
+
+    pub fn unset_key(&mut self) {
+        self.enckey = None;
```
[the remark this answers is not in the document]
simo5: yes it would, I think

src/storage/nssdb/ci.rs
```diff
@@ -12,6 +15,130 @@ use crate::token::TokenFacilities;
...
+            id: *id,
+            l: read_lock,
+        })
+    }
+
+    fn set_cached_key(&self, id: &[u8; SHA256_LEN], key: Object) -> Result<()> {
+        match self.cache.write() {
+            Ok(mut w) => {
+                if w.len() > MAX_KEY_CACHE_SIZE {
+                    let _ = w.pop_last();
```
Jakuje, line 121: This is basically dropping a random item from the BTreeMap as it is ordered. Is there some way to be able to drop the oldest item without too much complication, or this is the best approximation we can get?

src/storage/nssdb/mod.rs
```diff
@@ -1197,15 +1197,16 @@ impl Storage for NSSStorage {
...
-        )?;
+        let mut encdata =
+            encrypt_data(facilities, &newkeys, iterations, NSS_PASS_CHECK)?;
 
         /* FIXME: need to re-encode all encrypted/integrity protected attributes */
 
+        /* now that the pin has changed all cached keys are invalid, replace the lot */
+        self.keys = newkeys;
+        /* changing the pin does not leave the token logged in */
+        self.keys.unset_key();
```
Jakuje, line 1208: This sounds weird. We get the new keys, we replace them in `self` and then throw away as with logout. Would not it be more straightforward just throw away both instead of doing the indirect removal of old keys by assigning new keys and then flushing them too?

simo5: No because I want to preserve the cache of the pin which is created when we store it.

If you look carefully at the code you'll see caches are never dropped except when we whole sale replace all keys in this function, which makes sense because a change of PIN will change all keys so all caches are useless.

Jakuje (approved):
lgtm

PR #134, opened 2024-12-16
Add docs and some minor cleanups

### LINKED ISSUE #105: It would be nice if the configuration file and its directives was documented in README.md

simo5:
This PR adds documentation for the kryoptic_pkcs11 crate.

Additionally add documentation for internal functions too.

Do some cosmetic build cleanups as well.

Fixes #105

### DIFF
--- a/src/aes.rs
+++ b/src/aes.rs
@@ -7,10 +7,9 @@ use crate::attribute::Attribute;
 use crate::error::Result;
 use crate::interface::*;
 use crate::mechanism::*;
-use crate::misc::zeromem;
+use crate::misc::{cast_params, zeromem};
 use crate::object::*;
 use crate::ossl::aes::*;
-use crate::{attr_element, cast_params};
 
 use once_cell::sync::Lazy;
 
--- a/src/aes.rs
+++ b/src/aes.rs
@@ -13,9 +13,15 @@ use crate::ossl::aes::*;
 
 use once_cell::sync::Lazy;
 
+/// Smallest AES Key Size (128 bits)
 pub const MIN_AES_SIZE_BYTES: usize = 16; /* 128 bits */
+/// Medium AES Key size (192 bits)
 pub const MID_AES_SIZE_BYTES: usize = 24; /* 192 bits */
+/// Largest AES Key Size (256 bits)
 pub const MAX_AES_SIZE_BYTES: usize = 32; /* 256 bits */
+
+/// The AES block size is 128 bits (16 bytes) for all currently implemented
+/// variants
 pub const AES_BLOCK_SIZE: usize = 16;
 
 pub(crate) fn check_key_len(len: usize) -> Result<()> {
@@ -24,6 +30,19 @@ pub(crate) fn check_key_len(len: usize) -> Result<()> {
         _ => Err(CKR_KEY_SIZE_RANGE)?,
     }
 }
+
+/// This is a specialized factory for objects of class CKO_SECRET_KEY
+/// and CKA_KEY_TYPE of value CKK_AES
+///
+/// [AES secret key objects](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203476)
+/// (Version 3.1)
+///
+/// Derives from the generic ObjectFactory, CommonKeyFactory and
+/// SecretKeyFactory
+///
+/// This is used to store the list of attributes allowed for an AES Key object, as well as provide
+/// methods for generic manipulation of AES key objects (generation, derivation, wrapping ...)
+///
 
 #[derive(Debug)]
 pub struct AesKeyFactory {
@@ -65,6 +84,9 @@ impl AesKeyFactory {
 }
 
 impl ObjectFactory for AesKeyFactory {
+    /// Creation of AES keys use the default generic secret creation
+    /// code and additionally ensures the key size is one of the AES allowed
+    /// sizes (currently 128, 192 or 256 bits).
     fn create(&self, template: &[CK_ATTRIBUTE]) -> Result<Object> {
         let mut obj = self.default_object_create(template)?;
         let len = self.get_key_buffer_len(&obj)?;
@@ -117,6 +139,8 @@ impl ObjectFactory for AesKeyFactory {
         SecretKeyFactory::import_from_wrapped(self, data, template)
     }
 
+    /// The AES derive adds key length checks on top of the generic secret
+    /// derive helper
     fn default_object_derive(
         &self,
         template: &[CK_ATTRIBUTE],
@@ -148,6 +172,8 @@ impl SecretKeyFactory for AesKeyFactory {
         ObjectFactory::default_object_unwrap(self, template)
     }
 
+    /// Helper to set key that check the key is correctly formed for an
+    /// AES key object
     fn set_key(&self, obj: &mut Object, key: Vec<u8>) -> Result<()> {
         let keylen = key.len();
         check_key_len(keylen)?;
@@ -156,6 +182,8 @@ impl SecretKeyFactory for AesKeyFactory {
         Ok(())
     }
 
+    /// AES recommends very specific size preferring the largest key size
+    /// first and going down from there
     fn recommend_key_size(&self, max: usize) -> Result<usize> {
         if max >= MAX_AES_SIZE_BYTES {
             Ok(MAX_AES_SIZE_BYTES)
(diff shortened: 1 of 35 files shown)


src/config.rs
```diff
@@ -26,6 +26,33 @@ const DEFAULT_CONF_DIR: &str = "test";
...
+/// idealized hardware slot where a token can be inserted at any time to
+/// execute operations.
+///
+/// In Kryoptic we use slots to allow to provide multiple independent
+/// tokens with their own storage separate from any other slot. Slots
+/// can't share the same storage.
+///
+/// Each slot is identified by a slot number (a u32 quantity) and can
+/// optionally have a customized description and manufacturer string.
+/// If not description or manufacturer strings are provided then default
```
Jakuje, line 41:
```
/// If no description nor manufacturer strings are provided then default
```

src/config.rs
```diff
@@ -130,12 +203,19 @@ impl Config {
         }
     }
 
+    /// Generates a configuration structure from the named file which must
+    /// be a properly formatted confgiuation file in toml format.
```
Jakuje, line 207:
```
    /// be a properly formatted configuration file in toml format.
```

src/config.rs
```diff
@@ -156,6 +236,9 @@ impl Config {
         Ok(conf)
     }
 
+    /// Ensure all slot numbers are consistents, and allocates new slot
```
Jakuje, line 239:
```
    /// Ensure all slot numbers are consistent, and allocates new slot
```

src/ossl/aes.rs
```diff
@@ -1144,6 +1204,7 @@ impl AesOperation {
         }
     }
 
+    /// Instantiates a messaged-based encryption operation
```
Jakuje, line 1207:
```suggestion
    /// Instantiates a Message Based encryption operation
```
the same for the following docs entries. Using the version without the dash as this is used in the previous doc entries and would be good to have the naming systematic. Actually I do not care if it will be "message-based" or "Message Based", but it should be systematic.

src/ossl/aes.rs
```diff
@@ -1572,6 +1647,11 @@ impl Encryption for AesOperation {
         Ok(outlen)
     }
 
+    /// Provides the expect buffer size for the provided input plaintext length
```
Jakuje, line 1650:
```suggestion
    /// Provides the expected encrypted output buffer size for the provided input plaintext length
```

src/ossl/aes.rs
```diff
@@ -1967,6 +2057,11 @@ impl Decryption for AesOperation {
         Ok(outlen)
     }
 
+    /// Provides the expect buffer size for the provided input ciphertext length
```
Jakuje, line 2060:
```suggestion
    /// Provides the expected decrypted output buffer size for the provided input ciphertext length
```
the same clarification should come to the message-based functions below

src/ossl/aes.rs
```diff
@@ -2077,6 +2177,9 @@ impl MsgEncryption for AesOperation {
         self.msg_encrypt_new(param, paramlen, aad)
     }
 
+    /// Feed the next plaintext buffer to be encrypted
```
Jakuje, line 2180:
```suggestion
    /// Feed in the next plaintext buffer to be encrypted
```

src/aes.rs
```diff
@@ -7,16 +7,21 @@ use crate::attribute::Attribute;
...
 use crate::ossl::aes::*;
-use crate::{attr_element, cast_params};
 
 use once_cell::sync::Lazy;
 
+/// Smallest AES Key Size (128 bits)
 pub const MIN_AES_SIZE_BYTES: usize = 16; /* 128 bits */
+/// Medium AES Key size (192 bits)
 pub const MID_AES_SIZE_BYTES: usize = 24; /* 192 bits */
+/// Biggest AES Key Size (256 bits)
```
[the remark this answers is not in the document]
simo5: Potato, tomato :-)

src/lib.rs
```diff
@@ -1,6 +1,12 @@
 // Copyright 2023 Simo Sorce
 // See LICENSE.txt file for terms
 
+#![warn(missing_docs)]
+
+//! This is Kryoptic
+//!
+//! A cryptographic software token using the PKCS#11 standard API
```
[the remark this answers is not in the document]
simo5: I am ok with PKCS#11, I do not think the space adds anything and as you noted it makes some parsing iffy.

src/lib.rs
```diff
@@ -492,10 +512,18 @@ fn set_ec_point_encoding(val: config::EcPointEncoding) -> CK_RV {
     CKR_OK
 }
 
+/// Implementation of C_Finalize function
+///
+/// Version 3.1 Specification: [https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203256](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203256)
```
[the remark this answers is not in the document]
simo5: If you look at the generated docs you will see they are very useful as you can click them to go straight to the part of the spec you need to read to understand the function, woithout haveing to search for it.

As for links updates, that is necessary only if the 3.2 spec is different, which is very unlikley (basically only the TLS bits (where we fixed mistakes) will change, all the other links will remain relevant.

Net new 3.2 functions will point to the 3.2 standard.

src/lib.rs
```diff
@@ -677,13 +745,23 @@ extern "C" fn fn_get_session_info(
     }
     CKR_OK
 }
+
+/// Implementation of C_GetOperationState function (Not Implemented Yet)
```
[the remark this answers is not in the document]
simo5: same difference, I do not see it as awkward, it is just factual.

src/lib.rs
```diff
@@ -3334,6 +3777,18 @@ static INTERFACE_SET: Lazy<Vec<InterfaceData>> = Lazy::new(|| {
     v
 });
 
+/// Provides access to the list of interfaces defined by this implementation
+///
+/// Starting with PKCS#11 version 3.0.0 tokens provide a list of interfaces
```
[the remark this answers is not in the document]
simo5: ack

simo5:
I fixed the reported mistakes, and I think this PR is already large enough to throw it in as is, and use followups to add more.

src/ossl/aes.rs
```diff
@@ -2065,6 +2164,7 @@ impl MsgEncryption for AesOperation {
         self.msg_encrypt_final(param, paramlen, plain, cipher)
     }
 
+    /// Begin a new message based encryption
```
Jakuje, line 2167: nit: Some of the docs mention the Message Based with upper and some with lower first case. Would be good to have this systematic too.

src/ossl/aes.rs
```diff
@@ -2221,6 +2327,8 @@ impl MsgEncryption for AesOperation {
         Ok(outlen)
     }
 
+    /// Provides the expect buffer size for the provided input plaintext length
```
Jakuje, line 2330: I mentioned this in some similar function that it is not clear what buffer size it is. Also expect -> expected. Please check similar length functions in this file if it has matching clear wording:
```suggestion
    /// Provides the expected output ciphertext buffer size for the provided input plaintext length
```

src/object.rs
```diff
@@ -1153,19 +1264,33 @@ pub trait SecretKeyFactory {
...
+    /// this function to ensure the key is appropriately formed for the type.
+    /// For example by checking the buffer length.
     fn set_key(&self, obj: &mut Object, key: Vec<u8>) -> Result<()> {
         let keylen = key.len();
         obj.set_attr(Attribute::from_bytes(CKA_VALUE, key))?;
         self.set_key_len(obj, keylen)?;
         Ok(())
     }
 
+    /// Helper to allow other mechanisms select the correct key length.
```
Jakuje, line 1277:
```suggestion
    /// Helper to allow other mechanisms to select the correct key length.
```

src/lib.rs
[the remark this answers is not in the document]
Jakuje: As in the other part of the PR, I am ok with the links, but I would rather than showing them as clickable urls, have them with a section title (I mixed up the parameters in the original comment):

/// [Version 3.1 Specification: 5.4.2 C_Finalize](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203256)

what do you think?

src/lib.rs
```diff
@@ -3334,6 +3777,18 @@ static INTERFACE_SET: Lazy<Vec<InterfaceData>> = Lazy::new(|| {
     v
 });
 
+/// Provides access to the list of interfaces defined by this implementation
+///
+/// Starting with PKCS#11 version 3.0 tokens provide a list of interfaces
+/// that can be obtained from the token. Each interface provides a name, and
```
Jakuje, line 3783: I would say here `module` instead of token as the token is a term for the thing inside of the slot in pkcs11 terminology.
```suggestion
/// Starting with PKCS#11 version 3.0 modules provide a list of interfaces
/// that can be obtained from the module. Each interface provides a name, and
```

src/lib.rs
```diff
@@ -3369,6 +3824,14 @@ pub extern "C" fn C_GetInterfaceList(
     CKR_OK
 }
 
+/// Returns a specific interface identified by name and version
+///
+/// Applications that wants to immediately access a specific interface name,
+/// optionally a speific version too.
```
Jakuje, line 3830:
```suggestion
/// optionally a specific version too.
```

src/object.rs
simo5: There is no need to add a to, it's correct either way.

Jakuje (approved):
looks good!

PR #136, opened 2024-12-19
Update packaging template

Jakuje:
The packaging template stayed behind, not including changes from the features refactoring. I regenerated it also using the new rust2rpm and tested all works. Updated the copr

https://copr.fedorainfracloud.org/coprs/jjelen/kryoptic/build/8414994/

and package review request:

https://bugzilla.redhat.com/show_bug.cgi?id=2325933


simo5 (approved):
LGTM

PR #137, opened 2024-12-20
More performance improvements with nssdb

simo5:
Using valgrind's callgrind tool I am analyzing the behavior of with the nssdb storage backend which shows particularly bad performance in pkcs11-provider's CI.

The first easy pick was to remove the Zeroize crate, which single-handedly accounted for a 50% of the time spent in pbkdf2 operations.

The second easy pick was to rewrite the HAM initialization to use fixed slices instead of vectors for the internal state.

The nssdb is still much slower than the others, due to higher use of the pbkdf2 function, yet nssdb uses the same and it is much faster.

Some playing around shows the "native" version is about 20/30 % slower than the OpenSSL one we use in fips mode ... but there is definitely more work to do to improve performance here.

### DIFF
--- a/src/misc.rs
+++ b/src/misc.rs
@@ -6,7 +6,7 @@ use crate::attribute::{Attribute, CkAttrs};
 use crate::error::Result;
 use crate::interface::*;
 use crate::object::{Object, ObjectFactories, ObjectType};
-
+use crate::ossl::common::zeromem as ossl_zeromem;
 pub const CK_ULONG_SIZE: usize = std::mem::size_of::<CK_ULONG>();
 
 #[macro_export]
@@ -190,4 +190,8 @@ pub fn copy_sized_string(s: &[u8], d: &mut [u8]) {
         d[..slen].copy_from_slice(&s[..slen]);
         d[slen..].fill(0x20); /* space in ASCII/UTF8 */
     }
+}
+
+pub fn zeromem(mem: &mut [u8]) {
+    ossl_zeromem(mem);
 }
--- a/src/native/hmac.rs
+++ b/src/native/hmac.rs
@@ -12,6 +12,12 @@ use crate::misc::zeromem;
 
 use constant_time_eq::constant_time_eq;
 
+/* max algo right now is SHA3_224 with 144 bytes blocksize,
+ * use slightly larger for good measure (and alignment) */
+const MAX_BSZ: usize = 160;
+const IPAD_INIT: [u8; MAX_BSZ] = [0x36; MAX_BSZ];
+const OPAD_INIT: [u8; MAX_BSZ] = [0x5c; MAX_BSZ];
+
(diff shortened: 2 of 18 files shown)


src/misc.rs
```diff
@@ -6,7 +6,7 @@ use crate::attribute::{Attribute, CkAttrs};
 use crate::error::Result;
 use crate::interface::*;
 use crate::object::{Object, ObjectFactories, ObjectType};
-
+use crate::ossl::common::zeromem as ossl_zeromem;
```
Jakuje, line 9: This pulls openssl into the common, which might be an issue if we would like to support some different backend. Likely not an issue now, but more like a question/though, if this could be done in some crypto agnostic manner (memset?).

simo5:
Ok I nailed the second biggest performance issue.
Amazing how much just the "native" HMAC initialization impacted performance.
Switching to slices shaved off another 50% of performance drop in the HMAc function.
Tested the code with a pkcs11-provider CI run and now kryoptic.nss tests are at most 4 times slower than NSS ones in the worst case, and generally close or at most twice as slow, including setup.
The rest of performance issues will probably need to be found in the actual storage layer, and will be handle in a future PR.

simo5:
Eh, sounds like I broke something :-D
Hold on

simo5:
Ok found a copy/paste bug in the reinit() function that caused the failures.
Btw I tested with a release build, and the performance in pkcs11-provider CI is very close to that of NSS, in the worst case I observed only 30% slower tests, but generally performance is on par and within margin of error of measurement.

src/native/hmac.rs
```diff
@@ -8,9 +8,14 @@ use crate::hash;
 use crate::hmac::*;
 use crate::interface::*;
 use crate::mechanism::*;
+use crate::misc::zeromem;
 
 use constant_time_eq::constant_time_eq;
-use zeroize::Zeroize;
+
+/* max algo right now is SHA3_224 with 144 byts blocksize,
```
Jakuje, line 15:
```suggestion
/* max algo right now is SHA3_224 with 144 bytes blocksize,
```

src/native/hmac.rs
```diff
@@ -8,9 +8,14 @@ use crate::hash;
...
+ * use slightly larger for good measure (and alignment) */
+const IPAD_INIT: [u8; 160] = [0x36; 160];
+const OPAD_INIT: [u8; 160] = [0x5c; 160];
```
Jakuje, line 18: should we make the number 160 constant instead of using it on at least 5 different places here and below?

Jakuje (approved):
just nits.

src/native/hmac.rs
simo5: Yeah will do

simo5:
Nits fixed, and tests pass. Merging.

PR #138, opened 2025-01-03
Improve output length checking for FIPS indicators in AES/HMAC code

Jakuje:
There were several issues:
 * The `check_mac_fips_indicators()` won't do anything for the CMACs, as the cmac implementation in FIPS provider does not implement the fips indicator OSSL_PARAM.
* The AES key with SIGN/VERIFY flags was also not marked correctly as FIPS valid so any CMAC operation on that given key caused the FIPS indicator failure.
* Adding explicit tag/mac length checks for CMAC and GCM
* Ignoring CCM and KMAC as I think we do not certify these

I am still not sure about the GCM as the specs are vague. Putting 64b for now.

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -1394,6 +1394,18 @@ impl AesOperation {
             _ => return Err(self.op_err(CKR_GENERAL_ERROR)),
         };
 
+        /*
+         * NIST SP 800-38D: 5.2.1.2 Output Data
+         * > t may be any one of the following five values: 128, 120, 112,
+         * > 104, or 96. For certain applications, t may be 64 or 32;
+         *
+         * We assume here that 64b (8B) is still acceptable value and since
+         * we take the length from user in bytes, we do not have to bother
+         * about values non-dividable by 8.
+         */
+        if self.params.taglen < 8 {
+            self.fips_approved = Some(false);
+        }
         Ok(())
     }
 }
@@ -2632,9 +2644,22 @@ impl AesCmacOperation {
         zeromem(&mut buf);
 
         #[cfg(feature = "fips")]
-        {
-            self.fips_approved = check_mac_fips_indicators(&mut self.ctx)?;
-        }
+        self.fips_approval_cmac()?;
+
+        Ok(())
+    }
+
+    #[cfg(feature = "fips")]
+    fn fips_approval_cmac(&mut self) -> Result<()> {
+        /*
+         * NIST SP 800-38B A.2:
+         * > For most applications, a value for Tlen that is at least 64
+         * > should provide sufficient protection against guessing attacks.
+         *
+         * 64b == 8B
+         */
+        self.fips_approved = Some(self.maclen >= 8);
+
         Ok(())
     }
 }
--- a/src/fips/indicators.rs
+++ b/src/fips/indicators.rs
@@ -510,6 +510,8 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
             restrictions: [restrict!(CKK_AES), restrict!()],
             genflags: CKF_ENCRYPT
                 | CKF_DECRYPT
+                | CKF_SIGN // for CMAC
+                | CKF_VERIFY
                 | CKF_WRAP
                 | CKF_UNWRAP
                 | CKF_DERIVE,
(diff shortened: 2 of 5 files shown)


src/ossl/aes.rs
```diff
@@ -2632,9 +2636,20 @@ impl AesCmacOperation {
...
+    }
+
+    #[cfg(feature = "fips")]
+    fn fips_approval_cmac(&mut self) -> Result<()> {
+        /* The MAC len needs to have specific length for FIPS
+         * NIST SP 800-38B A.2:
+         * > For most applications, a value for Tlen that is at least 64
+         * should provide sufficient protection against guessing attacks.
+         */
+        self.fips_approved = Some(self.maclen >= (32 / 8));
```
jvdsn, line 2651: Is it intended to use 32 or 64 bits here? The comment seems to reference 64 bits but the code uses 32 bits.

jvdsn:
For what it's worth, if you're ever thinking about certifying KMAC, OpenSSL < 3.4 doesn't have a check. OpenSSL 3.4+ should have a check.

src/ossl/aes.rs
Jakuje: Indeed. Thanks for double-checking.

src/ossl/aes.rs
```diff
@@ -1394,6 +1394,10 @@ impl AesOperation {
             _ => return Err(self.op_err(CKR_GENERAL_ERROR)),
         };
 
+        /* The Tag len needs to have specific length for FIPS */
+        if self.params.taglen < 64 / 8 {
```
simo5, line 1398: why 64 / 8 and not just .. 8 ?

src/ossl/aes.rs
```diff
@@ -2632,9 +2636,20 @@ impl AesCmacOperation {
...
+    }
+
+    #[cfg(feature = "fips")]
+    fn fips_approval_cmac(&mut self) -> Result<()> {
+        /* The MAC len needs to have specific length for FIPS
+         * NIST SP 800-38B A.2:
+         * > For most applications, a value for Tlen that is at least 64
+         * should provide sufficient protection against guessing attacks.
+         */
+        self.fips_approved = Some(self.maclen >= (64 / 8));
```
simo5, line 2651: sounds like you use 64/8 to indicate we need to check for 64bit length but the quantity is in bytes.

I would see us either definintg a const or using:
   64 / sizeof!(u8) 

sizeof is a macro in src/misc.c

simo5: ah but sizeof() returns the result in CK_ULONG, so I guess you'll have to do:
 64 / std::mem::size_of::<u8>()
instead

simo5 (approved):
Only nits, otherwise this looks good to me, any reason why it is still in draft ?

src/ossl/aes.rs
Jakuje: Correct. I wanted to express the bytes/bits conversion here for readability. The standards talk about bits, but we have lengths in bytes.

The sizeof will return the number of bytes and not bits that I would need here. Getting number of bits from some constant (especially from u8) sounds to me quite redundant.

Moreover, I think everything in the world now counts on 8 bits in byte.

simo5: bah I am not thinking straight yet :)

Maybe provide a macro?
bits_to_bytes!(64)

?

Jakuje: That sounds to me over engineering the task of getting a number 8 from 64. Maybe some appropriate comment would server better?

simo5: A comment and just the number 8 would be fine for this case.

Jakuje: After trying to find at least some reference to put into the comment, I do not see anything specific in
https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf
It is still not clear to me if even the 64b should be approved or 32 should, as the wording is as vague as Stephan's:

> t may be any one of the following five values: 128, 120, 112, 104, or 96. For certain applications, t may be 64 or 32;

Jakuje: Changed to plain numbers and added appropriate comments.

simo5:
Looks good, feel free to push once tests pass

Jakuje:
All green. Merged.

PR #139, opened 2025-01-07
Add pragma to ensure secure delete is on

simo5:
In many distributions this seem to be on by default already, although the official sqlite documentations ays it is off by default.

Also by experimentation it seem this setting is not saved to the db file, so ensure secure_delete is turned on whenever we open the database.


simo5:
In FIPS mode we even need to scrub public keys in some cases ... so I went for the nuclear option which is to scrub everything (it is the default in Fedora sqlite apparently anyway, so no performance loss was recorded).

PR #142, opened 2025-01-15
Increase test coverage of ALLOWED_MECHANISMS usage

Jakuje:
Currently, `ALLOWED_MECHANISMS` were tested only with AES keys. This add test with RSA keys to make sure they are effective also with different key types.


simo5 (approved):
LGTM, feel free to merge once rebased

PR #143, opened 2025-01-20
Add integration test with pkcs11-provider

Jakuje:
This inspects the pkcs11 provider keeps working with kryoptic, which is mostly symmetric to the kryoptic test in the pkcs11-provider repo in:

https://github.com/latchset/pkcs11-provider/blob/main/.github/workflows/kryoptic.yml


simo5 (approved):
Great addition to our CI!
Thanks a lot.

PR #144, opened 2025-01-20
Padding bug mismatch

simo5:
Found while messing with some openssl test code and pkcs11-provider

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -749,12 +749,20 @@ impl AesOperation {
         if res != 1 {
             return Err(self.op_err(CKR_DEVICE_ERROR))?;
         }
-        if self.mech == CKM_AES_CBC_PAD {
-            let res =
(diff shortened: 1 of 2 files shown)


Jakuje (approved):
lgtm.
The only though I had is that the `decrypt_initialize()` and `encrypt_initialize()` have almost all repeated code so I though if we should factor it out to some `crypt_initialize()` and use the `EVP_CipherInit_ex()` and similar functions in here. But I can do it in separate PR if you would find it useful.

simo5:
> lgtm. The only though I had is that the `decrypt_initialize()` and `encrypt_initialize()` have almost all repeated code so I though if we should factor it out to some `crypt_initialize()` and use the `EVP_CipherInit_ex()` and similar functions in here. But I can do it in separate PR if you would find it useful.

I am not against, but it is also minor impact work, so can be done later as needed.

Jakuje:
ok. Will make it into an issue so it wont be lost and do it once I will have some time.

Issue #108, opened 2024-10-30
Support SoftHSMv2 sql database format


abbra:
Allow reading/writing in SoftHSMv2 sql database format to provide transparent migration path.

This would help FreeIPA to migrate existing environments where we use SoftHSM for DNSSEC to use kryoptic without forcing data migration.

PR #146, opened 2025-01-28
Add a tool to migrate data from a softhsm token into kryoptic

simo5:
I evaluated whether to implement a softhsm driver as requested in #108 however ultimately I decided that supporting writing to a sub-specified file-format dependent on things like casting C++ enums to integers and file based locks was not really something I wanted to do.

Instead this tools solves the problem by allowing wholesale migration of the data to our database format which we know how to handle properly.

Incidentally the tool allows importing softhsm data into any token that exposes a C_CreateObject API and supports all the object types that the softhsm stored.


simo5:
Tests are still missing, will have to add some in CI.

simo5:
Added a test that can be run with make tests to verify that the data read exist in the token and can be used.

Unfortunately this cannot be done in cargo test due to cargo limitation in managing cdylib crates.

Cargo.toml
```diff
@@ -16,6 +16,10 @@ name = "conformance"
 path = "src/conformance/main.rs"
 test = false
 
+[[bin]]
+name = "sh_migrate"
```
Jakuje, line 20: I am wondering if this is a good name for the executable. The `sh` is ambiguous so I would prefer `softhsm_migrate` to be more clear. I think being it short is really not an issue as it should be one-off tool and not something to use on daily basis.

Makefile
```diff
@@ -18,3 +20,6 @@ fix-format:
 
 check-spell:
 	@.github/codespell.sh
+
+tests: build
+	src/tools/softhsm/test.sh
```
Jakuje, line 25: should we run it with the `check` targets we have instead of reinventing a new one?

src/tools/softhsm/migrate.rs
```diff
@@ -0,0 +1,1068 @@
...
+            session: CK_INVALID_HANDLE,
+        })
+    }
+
+    fn initialize(&self, initargs: &CStr) -> Result<(), Error> {
+        unsafe {
+            match (*self.fntable).C_Initialize {
+                None => {
+                    return Err(format!(
+                        "Broken pkcs11 token, non Initialize function in list"
```
Jakuje, line 121:
```suggestion
                        "Broken pkcs11 module, no C_Initialize function in function list"
```

src/tools/softhsm/migrate.rs
```diff
@@ -0,0 +1,1068 @@
...
+
+    fn open_session(&mut self) -> Result<CK_SESSION_HANDLE, Error> {
+        if self.session != CK_INVALID_HANDLE {
+            return Ok(self.session);
+        }
+        self.session = unsafe {
+            match (*self.fntable).C_OpenSession {
+                None => {
+                    return Err(format!(
+                        "Broken pkcs11 token, non Initialize function in list"
```
Jakuje, line 163:
```suggestion
                        "Broken pkcs11 module, no C_OpenSession function in function list"
```

src/tools/softhsm/migrate.rs
```diff
@@ -0,0 +1,1068 @@
...
+        };
+        Ok(self.session)
+    }
+
+    fn login(&mut self, pin: &str) -> Result<CK_SESSION_HANDLE, Error> {
+        unsafe {
+            match (*self.fntable).C_Login {
+                None => {
+                    return Err(format!(
+                        "Broken pkcs11 token, non Initialize function in list"
```
Jakuje, line 196:
```suggestion
                        "Broken pkcs11 module, no C_Login function in function list"
```

src/tools/softhsm/migrate.rs
```diff
@@ -0,0 +1,1068 @@
...
+
+    let session = pkcs11.open_session()?;
+
+    /* Derive key from pin.
+     * SoftHSM2 indicated a Key Derivation specified in RFC4800 is used,
+     * but not exactly which if the variants, so here is a description
+     * of what it does:
+     *
+     * Salt s must be 8 bytes
+     * Iteation counter c is 1500 + the last byte of the salt as a u8 number
```
Jakuje, line 753:
```suggestion
     * Iteration counter c is 1500 + the last byte of the salt as a u8 number
```
found by codespell:
```
src/tools/softhsm/migrate.rs:753: Iteation ==> Iteration
```

src/tools/softhsm/migrate.rs
```diff
@@ -0,0 +1,1068 @@
...
+
+    let encryptionkey = match encryption_key(pkcs11, &pin_attr.value, pin) {
+        Ok(k) => k,
+        Err(e) => {
+            eprintln!("Failed to get key: {}", e);
+            return 0xFE;
+        }
+    };
+
+    let mut error = 0u8;
```
Jakuje, line 936: we can have only 256 errors? Is it enough?

Makefile
Jakuje: Or ad this to the CI so we can see it works in different configurations.

Cargo.toml
simo5: ack

Makefile
simo5: I have another commit I forgot for the CI, we can't really run as part of check because check does a cargo test, and we need a cargo build, and I do not want to do both for the common check case.

src/tools/softhsm/migrate.rs
simo5: plenty for now

.github/workflows/migrate.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Install Dependencies
+        run: |
+          dnf -y install git cargo clang-devel openssl-devel \
+            'perl(FindBin)' 'perl(lib)' 'perl(File::Compare)' \
+            'perl(File::Copy)' 'perl(bigint)' 'perl(Time::HiRes)' \
+            'perl(IPC::Cmd)' 'perl(Pod::Html)' 'perl(Digest::SHA)' \
+            'perl(Module::Load::Conditional)' 'perl(File::Temp)' \
+            'perl(Test::Harness)' 'perl(Test::More)' 'perl(Math::BigInt)' \
+            zlib-devel sed sqlite-devel
```
Jakuje, line 41: you don't need all the openssl dependencies if you build the dynamic feature

simo5: true, copy/paste remnants

Jakuje (approved):

PR #147, opened 2025-01-29
Add TLS PRF selftests in FIPS mode

Jakuje:
Second version does only the TLS PRF test with @simo5's suggested approach, that will invoke dynamically (and only once using the Lazy construct), as well as for the `CKM_TLS_MAC` mechanisms, that internally use TLS PRF.

### DIFF
--- a/src/fips/indicators.rs
+++ b/src/fips/indicators.rs
@@ -245,7 +245,7 @@ struct FipsMechanism {
 
 struct FipsChecks {
     keys: [FipsKeyType; 15],
-    mechs: [FipsMechanism; 67],
+    mechs: [FipsMechanism; 71],
 }
 
 /* TODO: double check the values, this is just an initial
@@ -782,6 +782,15 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
             ],
             genflags: 0,
         },
+        FipsMechanism {
+            mechanism: CKM_TLS_MAC,
+            operations: CKF_SIGN | CKF_VERIFY,
+            restrictions: [
+                restrict!(CKK_GENERIC_SECRET, range!(112, 512)),
+                restrict!(),
+            ],
+            genflags: 0,
+        },
         /* Key gen, gen/derive */
         FipsMechanism {
             mechanism: CKM_PKCS5_PBKD2,
@@ -796,10 +805,10 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
             mechanism: CKM_GENERIC_SECRET_KEY_GEN,
             operations: CKF_GENERATE,
             restrictions: [
-                restrict!(CKK_GENERIC_SECRET, range!(112, 256)),
+                restrict!(CKK_GENERIC_SECRET, range!(112, 255 * 64 * 8)),
                 restrict!(),
             ],
-            genflags: CKF_DERIVE,
+            genflags: CKF_SIGN | CKF_VERIFY | CKF_DERIVE,
         },
         FipsMechanism {
             mechanism: CKM_HKDF_KEY_GEN,
@@ -839,6 +848,45 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
                 | CKF_DECRYPT
                 | CKF_WRAP
                 | CKF_UNWRAP
+                | CKF_DERIVE,
+        },
+        FipsMechanism {
+            mechanism: CKM_TLS12_MASTER_KEY_DERIVE,
+            operations: CKF_DERIVE,
+            restrictions: [
+                restrict!(CKK_GENERIC_SECRET, step!(48 * 8)),
+                restrict!(),
+            ],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
+                | CKF_DERIVE,
+        },
+        FipsMechanism {
+            mechanism: CKM_TLS12_KEY_AND_MAC_DERIVE,
+            operations: CKF_DERIVE,
+            restrictions: [
+                restrict!(CKK_GENERIC_SECRET, range!(112, 255 * 64 * 8)),
+                restrict!(),
+            ],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
+                | CKF_DERIVE,
+        },
+        FipsMechanism {
+            mechanism: CKM_TLS12_KEY_SAFE_DERIVE,
+            operations: CKF_DERIVE,
+            restrictions: [
+                restrict!(CKK_GENERIC_SECRET, range!(112, 255 * 64 * 8)),
+                restrict!(),
+            ],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
                 | CKF_DERIVE,
         },
         FipsMechanism {
@@ -1104,5 +1152,6 @@ pub fn is_approved(
         }
     }
 
+    /* mechanism not found in the indicators table -- not allowed */
     false
 }
--- a/src/tlskdf.rs
+++ b/src/tlskdf.rs
@@ -276,10 +276,12 @@ impl TLSKDFOperation {
     fn new(mech: &CK_MECHANISM) -> Result<TLSKDFOperation> {
         match mech.mechanism {
             CKM_TLS12_MASTER_KEY_DERIVE => Self::new_tls12_mk_derive(mech),
-            CKM_TLS12_KEY_AND_MAC_DERIVE => Self::new_tls12_keymac_derive(mech),
-            CKM_TLS12_KEY_SAFE_DERIVE => Self::new_tls12_keymac_derive(mech),
-            CKM_TLS12_KDF => Self::new_tls_generic_key_derive(mech),
-            CKM_TLS_KDF => Self::new_tls_generic_key_derive(mech),
+            CKM_TLS12_KEY_AND_MAC_DERIVE | CKM_TLS12_KEY_SAFE_DERIVE => {
+                Self::new_tls12_keymac_derive(mech)
+            }
+            CKM_TLS12_KDF | CKM_TLS_KDF => {
+                Self::new_tls_generic_key_derive(mech)
+            }
             _ => return Err(CKR_MECHANISM_INVALID)?,
         }
     }
@@ -770,10 +772,7 @@ impl Derive for TLSKDFOperation {
             CKM_TLS12_MASTER_KEY_DERIVE => {
                 self.derive_master_key(key, template, mechanisms, objfactories)
             }
-            CKM_TLS12_KEY_AND_MAC_DERIVE => {
-                self.derive_mac_key(key, template, mechanisms, objfactories)
-            }
-            CKM_TLS12_KEY_SAFE_DERIVE => {
+            CKM_TLS12_KEY_AND_MAC_DERIVE | CKM_TLS12_KEY_SAFE_DERIVE => {
                 self.derive_mac_key(key, template, mechanisms, objfactories)
             }
             CKM_TLS12_KDF | CKM_TLS_KDF => {
--- a/src/tlskdf.rs
+++ b/src/tlskdf.rs
@@ -3,7 +3,7 @@
 
 use std::fmt::Debug;
 
-use crate::attribute::CkAttrs;
+use crate::attribute::{Attribute, CkAttrs};
 use crate::error::Result;
 use crate::hmac::{hash_to_hmac_mech, register_mechs_only};
 use crate::interface::*;
@@ -11,6 +11,9 @@ use crate::mechanism::*;
 use crate::misc::CK_ULONG_SIZE;
 use crate::object::{Object, ObjectFactories};
 use crate::{bytes_to_slice, bytes_to_vec, cast_params};
+
+#[cfg(feature = "fips")]
+use crate::hmac::test_get_hmac;
 
 use constant_time_eq::constant_time_eq;
 use once_cell::sync::Lazy;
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -2242,6 +2242,29 @@ extern "C" fn fn_derive_key(
         _ => return CKR_MECHANISM_INVALID,
     };
 
+    #[cfg(feature = "fips")]
+    match mechanism.mechanism {
+        CKM_TLS12_MASTER_KEY_DERIVE
+        | CKM_TLS12_KEY_AND_MAC_DERIVE
+        | CKM_TLS12_KEY_SAFE_DERIVE
+        | CKM_TLS12_KDF
+        | CKM_TLS_KDF => {
```
jvdsn, line 2251: What are the use cases for CKM_TLS12_KDF and CKM_TLS_KDF? We wouldn't consider those mechanisms as approved for FIPS unless there's a TLS protocol usecase that can't be done using MASTER_KEY_DERIVE, KEY_AND_MAC_DERIVE, or KEY_SAFE_DERIVE.

src/token.rs
```diff
@@ -76,6 +76,8 @@ pub struct Token {
     storage: Box<dyn Storage>,
     session_objects: HashMap<CK_OBJECT_HANDLE, Object>,
     logged: CK_USER_TYPE,
+    #[cfg(feature = "fips")]
+    kdf_selftest_ran: bool,
```
jvdsn, line 80: Should this variable name be more specific?

src/lib.rs
simo5: Are you saying CKM_TLS[12]_KDF should be made unavailable in FIPS mode? (we can easily do that at build time).

src/token.rs
simo5: we are changing how (all of) this will be handled after a discussion with @Jakuje so wait a bit for reviewing details

src/lib.rs
jvdsn: SP 800-135r1 approves the KDFs only "in the context of the TLS protocol". So if there is no usecase for CKM_TLS[12]_KDF that is not already covered by the other mechanisms, then it should be non-approved, yes.

src/tlskdf.rs
```diff
@@ -210,6 +267,11 @@ impl Mechanism for TLSPRFMechanism {
             return Err(CKR_MECHANISM_INVALID)?;
         }
 
+        #[cfg(feature = "fips")]
```
simo5, line 270: I think this can be moved inside of TLSKDFOperation::new() (as the very first check) so you do not have to repeat it three times.

simo5: You can also just do:
```
if TLS_PRF_SELFTEST.result != CKR_OK {
    return Err(TLS_PRF_SELFTEST.result)?;
}
```

src/tlskdf.rs
```diff
@@ -200,6 +203,60 @@ impl TLSPRFMechanism {
     }
 }
 
+#[cfg(feature = "fips")]
+struct FIPSSelftest {
+    result: CK_RV,
+}
+
+#[cfg(feature = "fips")]
+static TLS_PRF_SELFTEST: Lazy<FIPSSelftest> = Lazy::new(|| {
```
simo5, line 212: Exactly what I expected

simo5 (approved):
LGTM

src/lib.rs
Jakuje: I think this will be a follow-up change in the indicators waters. For now, we have just HKDF in the fips indicators file so I assume we should also list some of the TLS KDFs mechanisms that we want to allow:
https://github.com/latchset/kryoptic/blob/ff6b251a6ab87956f2eef1440bead7d80032a4e3/src/fips/indicators.rs#L813

src/token.rs
Jakuje: Yeah. This was mostly PoC. All revamped now.

src/tlskdf.rs
Jakuje: I was thinking about that, but did not want to indent the whole function. Will figure out something.

simo5: no re-indenting necessary ?

Jakuje: Doing it in `TLSKDFOperation()` won' t cover the `TLSMACOperation` then. I can move it into these two, but then we are just on 2 from 3 and one level deeper in the call stack.

It could be moved also into the `TLSPRF::new()`, but I was worried that it would be too deep to easily understand.

src/lib.rs
Jakuje: I am a bit lost in the PKCS#11 specs about what is really needed for the TLS operations. I think you are right and we do not need the `CKM_TLS_KDF`. The specs talks about implementation of `key material exporters`, which does not sounds like something FIPS approved :)
The `CKM_TLS12_KDF` looks like generic algorithm too so lets mark them as unapproved in the indicators and see how it will go. I can do it in this PR once I already touched this. So far we do not have any tls kdf in the fips indicators list.

src/tlskdf.rs
Jakuje: Looks like we do not have any test coverage for the `CKM_TLS12_MAC` if I see right. That would be another TODO/next issue.

Jakuje: Done.

src/lib.rs
Jakuje: Done also the indicators, including the `CKM_TLS12_MAC`. But I am not exactly sure what sizes/operations limitation should be on all of them, mostly `CKM_TLS12_MAC`. Why the TLS has to be so complicated?

Jakuje:
Trying to rerun the ` Integration tests / Integration test with pkcs11-provider (pull_request) Queued — Waiting to run this check... ` which failed on the first attempt in `imported.kryoptic.nss` with
```
 openssl pkeyutl -sign
              -in ${TMPPDIR}/sha256.bin
pkeyutl: Error initializing context
8002CCC1517F0000:error:03000096:digital envelope routines:evp_pkey_signature_init:operation not supported for this keytype:crypto/evp/signature.c:575:
```

src/lib.rs
simo5: This is a bug I brought up for solution in PKCS#11 3.2 specification where we'll deprecate a bunch of duplicated "12" mechanisms that were never needed an never implemented by anybody :(
CKM_TLS_KDF == CKM_TLS12_KDF and CK_TLS_MAC == CKM_TLS12_MAC ...

simo5: This is an example of the language that there will be in the new spec: "Note: CKM_TLS12_MAC is provided only for historical reasons and should be considered deprecated. This mechanism shares the same behavior with CKM_TLS_MAC."

So we should mark only CKM_TLS_MAC and not CKM_TLS12_MAC as approved as we'll phase out CKM_TLS12_MAC later on once we add a "deprecated" feature.

simo5: Also this:
```CKM_TLS12_KDF and CKM_TLS12_MAC are considered aliases of the corresponding CKM_TLS_KDF and CKM_TLS_MAC mechanisms and behave identically. These mechanisms are available with a different identifier for historical reasons and are considered deprecated.```

Comment (Jakuje) [reply src/lib.rs]: Oh, yeah. It looked confusing at places. So the `CKM_TLS_MAC` is the one to use and it only differs from the 1.0/1.1 in the way how the key is derived initially?
In any case, I will update it to use the `CKM_TLS_MAC`. That is tested actually.

Comment (simo5) [reply src/lib.rs]: Yes please

Comment (Jakuje) [reply src/lib.rs]: I see we use the `CKM_TLS12_MAC` in the `TLS_MASTER_SECRET_ALLOWED_MECHS`. This should be then amended either to support both of them or just the non-deprecated one as the deprecation is not yet public and some people not close to the standardization group might have implemented this.

Comment (simo5) [reply src/lib.rs]: We should accept both mechanism numbers in input but only use the correct variants internally.

Comment (Jakuje) [reply src/lib.rs]: Seems like the indicators will need some more love too. With current changes, all of the mechanism from tlskdf are considered unapproved, because the test generates the SECRET_KEY of 48b, while the FIPS table has only the keygen approved only for the following size:
```
restrict!(CKK_GENERIC_SECRET, range!(112, 256)),
```
But the keys themselves are approved for larger range:
```
sizes: range!(112, 255 * 64 * 8), 
```
Is there any reason not to allow generic secret keygen of larger sizes?

Comment (simo5) [reply src/lib.rs]: I think GENERIC_SECRET key can be any size

**Comment by Jakuje** [comment]
Added also some fips_validation that the flags work for both derivation and MAC sign/verify.

Just to clarify, the generic `CKM_TLS12_KDF` is not allowed now.

For the `CKM_TLS_MAC` operation, I had to make the sign/verify operation explicitly approved after completion to report the right results. Are there some possible input/parameters combination for this mechanism that should not be approved (except for the generic checks done based on the key and mechanism tables)?

File: src/fips/indicators.rs [hunk 5]
```diff
@@ -841,6 +850,51 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
...
+            restrictions: [
+                restrict!(CKK_GENERIC_SECRET, step!(48 * 8)),
+                restrict!(),
+            ],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
+                | CKF_WRAP
+                | CKF_UNWRAP
```
simo5, line 865: Are we sure that keys derived from the TLS master key can be used to wrap/unwrap or further derive other keys ?
Mac and Encryption are obviously ok.

simo5, line 865: Are we sure that keys derived from the TLS master key can be used to wrap/unwrap or further derive other keys ?
Mac and Encryption are obviously ok.

src/fips/indicators.rs
```diff
@@ -841,6 +850,51 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
...
+                restrict!(CKK_GENERIC_SECRET, range!(112, 255 * 64 * 8)),
+                restrict!(),
+            ],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
+                | CKF_WRAP
+                | CKF_UNWRAP
+                | CKF_DERIVE,
```
simo5, line 881: Same Q as above

simo5, line 881: Same Q as above

src/tlskdf.rs
```diff
@@ -858,6 +930,12 @@ impl TLSMACOperation {
 
         let out = self.tlsprf.finish(&self.seed, self.outputlen)?;
         output.copy_from_slice(out.as_slice());
+
+        #[cfg(feature = "fips")]
+        {
```
simo5, line 935: are these braces actually needed?

simo5 (approved):
Minor nits, but overall LGTM

src/fips/indicators.rs
Jakuje: The PKCS#11 specs are vague and probably written by people who understand the TLS 1.2. The notes have the following wording (hope it is not for 1.0 and 1.1):

> The mechanism also contributes the CKA_ALLOWED_MECHANISMS attribute consisting only of CKM_TLS12_KEY_AND_MAC_DERIVE, CKM_TLS12_KEY_SAFE_DERIVE, CKM_TLS12_KDF and CKM_TLS12_MAC.

From there I read that the master key derived keys can be used for all sign/verify/encrypt/decrypt/derive. The wrap/unwrap really does not make sense, right.

src/fips/indicators.rs
Jakuje: The pkcs11 specs describe this that resulting keys have same usage as above, sign/verify/encrypt/decrypt/derive. The problem is that it generates two different keys, each with different usage, and this is something we can not catch here, I think:

>  The two MACing keys ("client_write_MAC_secret" and "server_write_MAC_secret") (if present) are always given a type of CKK_GENERIC_SECRET. They are flagged as valid for signing and verification.

> The other two keys ("client_write_key" and "server_write_key") are typed according to information found in the template sent along with this mechanism during a C_DeriveKey function call.  By default, they are flagged as valid for encryption, decryption, and derivation operations.

The wrap/unwrap is nonsense. Will remove.

Jakuje: the last one, `CKM_TLS12_KEY_SAFE_DERIVE`, should be treated the same way:

> CKM_TLS12_KEY_SAFE_DERIVE is identical to CKM_TLS12_KEY_AND_MAC_DERIVE except that it shall never produce IV data, and the  ulIvSizeInBits field of CK_TLS12_KEY_MAT_PARAMS is ignored and treated as 0.

src/tlskdf.rs
Jakuje: Yes:
```
error[E0658]: attributes on expressions are experimental
   --> src/tlskdf.rs:934:9
    |
934 |         #[cfg(feature = "fips")]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `kryoptic` (lib test) due to 1 previous error
```

src/fips/indicators.rs
```diff
@@ -1104,5 +1152,6 @@ pub fn is_approved(
         }
     }
 
+    /* the mechanism not found in the indicators table -- not allowed */
```
simo5, line 1155: nit: drop the leading "the"

simo5 (approved):

jvdsn:
I don't think we as a lab currently enforce lower/upper bounds on the inputs for the TLS PRFs (since it's a pretty big mess to determine the right values). In any case the limits seem reasonable enough to me.

jvdsn:
I would personally be more concerned about the allowed hash functions, but maybe that's checked somewhere else that I didn't see.

src/tlskdf.rs
```diff
@@ -200,6 +205,60 @@ impl TLSPRFMechanism {
...
+    key.set_attr(Attribute::from_bool(CKA_DERIVE, true))
+        .unwrap();
+
+    let mech = test_get_hmac(prf);
+
+    let mut tlsprf = match TLSPRF::init(&key, &mech, prf) {
+        Ok(a) => a,
+        Err(_) => return status,
+    };
+    let out = match tlsprf.finish(&seed, expect.len()) {
```
jvdsn, line 252: Is there any particular reason why TLSPRF::finish takes label and seed as one Vec instead of two parameters as defined in RFC 5246?

Jakuje:
> I would personally be more concerned about the allowed hash functions, but maybe that's checked somewhere else that I didn't see.

That is a good point. So I assume SHA2 and better are ok and SHA1 not or do we have some other limitations?

jvdsn:
> That is a good point. So I assume SHA2 and better are ok and SHA1 not or do we have some other limitations?

Only SHA-256, SHA-384 or SHA-512 are approved at this point.

simo5:
@Jakuje I guess we'll need to add a fips specific check at invocation that only those hashes are accepted.

simo5:
> @Jakuje I guess we'll need to add a fips specific check at invocation that only those hashes are accepted.

Probably the easiest way is to create a tlskdf.rs helper to use to source the Hash mechanism from the params, and in case of fips build will change the fips_approved boolean to false if a non approved hash is requested.

This also means we can't unconditionally set fips_approved to Some(true) at the end, but will have to check first if it hasn't been already set first.

jvdsn:
You might also need to think about hash algorithms for CKM_TLS_MAC/CKM_TLS12_MAC, CKM_PKCS5_PBKD2, CKM_RSA_PKCS_OAEP and PRFs for CKM_SP800_108_*_KDF. Also, do you support the extended master secret?

simo5:
For PBKDF2 and RSA I think the underlying openssl calls already take that in account, if not we'll have to fix it there anyway.

The extended secret mechanism is not yet supported because the mechanism is only available in PKCS#11 3.2 which is not released yet. In OASIS we are close to release, so I will shortly begin working on a PR to add the 3.2 pre-release headers and then we can add implementation of extended master secret

Issue #145, opened 2025-01-21
Rewrite AES encrypt and decrypt to reuse code


Jakuje:
> The only though I had is that the `decrypt_initialize()` and `encrypt_initialize()` have almost all repeated code so I though if we should factor it out to some `crypt_initialize()` and use the `EVP_CipherInit_ex()` and similar functions in here. But I can do it in separate PR if you would find it useful.

_Originally posted by @Jakuje in https://github.com/latchset/kryoptic/pull/144#pullrequestreview-2563960713_

PR #148, opened 2025-01-30
aes: Unify encrypt/decrypt initialization

Jakuje:
The functions encrypt_initialize() and decrypt_initialize() have most of the duplicated code except for the OpenSSL function names. This unfifies the use on functions that work in both direction, keeping the code simpler and needing to fix bugs only in one place.

There were slight differences already in both of the implementations so I tried to take the better variant with more descriptive comments and checking.

Fixes: #145

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -691,7 +691,7 @@ impl AesOperation {
         }
     }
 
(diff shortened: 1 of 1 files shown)


simo5 (approved):
LGTM

PR #149, opened 2025-02-03
Pkcs11 3.2 Draft

### LINKED ISSUE #141: FIPS Validation flags revisited
Discussed at length with Bob the meaning of the validation flags on stored objects, and the result is that we should not store flags on objects,a s flags are mostly meant for ephemeral keys generated in the token as part of complex chains of operations (like TLS key exchange).

Any key already stored in the database should be considered Ok to use, and the first operation that uses a key will re-validate that it meets the criteria for the currently implemented FIPS restrictions.

This allows upgrades where the FIPS requirement changes and keys already stored can be re-checked for validity (and are not considered valid just because eons earlier they passed test on generation with a different set of FIPS requirements).

simo5:
We have now the 3.2 draft, no additional changes that break the API are expected at this point.

This PR adds a conditional feature called pkcs11_3_2 so that the new changes are gated until we have a the finalized spec.

Fixes #141


src/object.rs
```diff
@@ -244,52 +244,56 @@ bitflags! {
...
 
         /* The attribute can only be change False -> True */
-        const ChangeToTrue         = 0b01010000000000;
+        const ChangeToTrue         = 0x00001400;
 
         /* The attribute can be changed only during a Copy Operation */
-        const ChangeOnCopy         = 0b10010000000000;
+        const ChangeOnCopy         = 0x00002400;
+
+        /* The attribute is epehmeral and should not be stored on
```
Jakuje, line 294:
```suggestion
        /* The attribute is ephemeral and should not be stored on
```

src/ec/montgomery.rs
```diff
@@ -79,35 +82,39 @@ impl ObjectFactory for ECMontgomeryPubFactory {
...
-        data.attributes.append(&mut data.init_common_key_attrs());
-        data.attributes
-            .append(&mut data.init_common_private_key_attrs());
-        data.attributes.push(attr_element!(
+        let mut factory: ECMontgomeryPrivFactory = Default::default();
+
+        factory.add_common_object_attrs();
+        factory.add_common_storage_attrs();
+        factory.add_common_key_attrs();
+        factory.add_common_private_key_attrs();
```
Jakuje, line 110: Could the `add_common_private_key_attrs()` add all the above attributes at once without the need to call all 4 functions for each private key factory?

src/storage/nssdb/attrs.rs
```diff
@@ -262,12 +262,18 @@ pub fn is_db_attribute(attr: CK_ATTRIBUTE_TYPE) -> bool {
...
+pub const NSS_SKIP_SIZE: usize = 4;
+
+pub static NSS_SKIP_ATTRIBUTES: [CK_ATTRIBUTE_TYPE; NSS_SKIP_SIZE] = [
     CKA_UNIQUE_ID,
     CKA_COPYABLE,
     CKA_DESTROYABLE,
-    CKA_VALIDATION_FLAGS,
     CKA_ALLOWED_MECHANISMS,
+    #[cfg(feature = "pkcs11_3_2")]
+    CKA_OBJECT_VALIDATION_FLAGS,
```
Jakuje, line 276: do you need to skip it here when it is already marked as ephemeral?

src/ec/montgomery.rs
simo5: I tried, but it doesn't work because each of these function is froma  different trait, and they do no derive from each other, lemme see if I can fix this with some clever trait magic

src/storage/nssdb/attrs.rs
simo5: well nssdb does not use the ephemeral stuff, but I guess it should ...

simo5: Maybe we should do this before even calling the storage layer.
But then there is some asymmetry with adding this value in nssdb on object retrieval ... hmmm

Jakuje (approved):
thanks! Just one more typo and I think we are good!

PR #150, opened 2025-02-11
Add FIPs fault indicator checks

simo5:
In FIPS mode we need to consistently check (and set) the general fault indicator which is turned permanently on when a self test fails.

As part of this change move the native implementation of tlskdf in the proper place under src/native


Jakuje (approved):
lgtm

PR #154, opened 2025-02-25
Fix RSA_X_509 mechanism

### LINKED ISSUE #153: Signature and decryption error with `CKM_RSA_X_509`
When running `pkcs11-tool --test` on kryoptic with one generated RSA 2048 key, the test fails with

```
Verify (currently only for RSA)
  testing key 0 (RSA2048)
    RSA-X-509:   ERR: C_Sign() returned CKR_DEVICE_ERROR (0x30)
Decryption (currently only for RSA)
  testing key 0 (RSA2048)
    RSA-PKCS: OK
    RSA-X-509: resulting cleartext doesn't match input
    Original: 00 7e 64 ff 81 eb 61 fd fe c3 9b 67 bf 0d e9 8c 7e 4e 32 bd f9 7c 8c 6a c7 5b a4 3c 02 f4 b2 ed 72 16 ec f3 01 4d f0 00 10 8b 67 cf 99 50 5b 17 9f 8e d4 98 0a 61 03 d1 bc a7 0d be 9b bf ab 0e d5 98 01 d6 e5 f2 d6 f6 7d 3e c5 16 8e 21 2e 2d af 02 c6 b9 63 c9 8a 1f 70 97 de 0c 56 89 1a 2b 21 1b 01 07 0d d8 fd 8b 16 c2 a1 a4 e3 cf d2 92 d2 98 4b 35 61 d5 55 d1 6c 33 dd c2 bc f7 ed de 13 ef e5 20 c7 e2 ab dd a4 4d 81 88 1c 53 1a ee eb 66 24 4c 3b 79 1e a8 ac fb 6a 68 f3 58 46 06 47 2b 26 0e 0d d2 eb b2 1f 6c 3a 3b c0 54 2a ab ba 4e f8 f6 c7 16 9e 73 11 08 db 04 60 22 0a a7 4d 31 b5 5b 03 a0 0d 22 0d 47 5d cd 9b 87 78 56 d5 70 4c 9c 86 ea 0f 98 f2 eb 9c 53 0d a7 fa 5a d8 b0 b5 db 50 c2 fd 5d 09 5a 2a a5 e2 a3 fb b7 13 47 54 9a 31 63 32 23 4e ce 76 5b 75 71 b6 4d
    Decrypted: b3 73 a9 18 6d c8 57 4c 7e e0 4d f5 9a 19 8f 2f dc 6d ba c6 85 12 d6 fe aa 42
```

After further investigation, it looks like the problem with `C_Sign` is connected to the length of signed data, which is 256 for the failed test in the `pkcs11-tool`. Up to 245 bytes, the signature is created without errors, but for larger data, the  `C_Sign` returns `CKR_DEVICE_ERROR`.

```
$ head -c 245 </dev/urandom > data
$ pkcs11-tool --module=kryoptic/target/debug/libkryoptic_pkcs11.so --token-label="Kryoptic Token" --login --pin=123456 --id 01 -s -m RSA-X-509 --input-file data --output-file data.sig
Using signature algorithm RSA-X-509
```

Here, the error in `C_SignUpdate` is reported just because the `C_Sign` failed at first and `pkcs11-tool` tries the `C_SignUpdate` afterwards:
```
$ head -c 246 </dev/urandom > data
$ pkcs11-tool --module=kryoptic/target/debug/libkryoptic_pkcs11.so --token-label="Kryoptic Token" --login --pin=123456 --id 01 -s -m RSA-X-509 --input-file data --output-file data.sig
Using signature algorithm RSA-X-509
error: PKCS11 function C_SignUpdate failed: rv = CKR_OPERATION_NOT_INITIALIZED (0x91)
Aborting.
```

However, I haven't found the cause of the decryption error.

The setup script used: [kryoptic-reproducer.txt](https://github.com/user-attachments/files/18961843/kryoptic-reproducer.txt)

Jakuje:
Turns out this mechanism was not handled correctly and when we do not specify padding to OpenSSL, it uses the PKCS#1.5 by default so both signature and verification seems to work, unless the input is too large for this padding.

Setting the explicit padding mode solves the problem and allows us to accept also input up to the length of the RSA modulus.

Fixes #153


simo5 (approved):
LGTM

PR #155, opened 2025-02-25
Another round of spec file updates after review  + RPM rebuild in CI

Jakuje:
I got few more comments in the Fedora review bug:

https://bugzilla.redhat.com/show_bug.cgi?id=2325933

It also integrates the RPM rebuild in CI, which we did not have

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,9 @@
 // Copyright 2023 Simo Sorce
 // See LICENSE.txt file for terms
 
+use std::env;
+use std::path::PathBuf;
+
 #[derive(Debug)]
 pub struct Pkcs11Callbacks;
 
@@ -32,6 +35,8 @@ impl bindgen::callbacks::ParseCallbacks for Pkcs11Callbacks {
 }
 
 fn ossl_bindings(header: &str, args: &[&str]) {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+
     bindgen::Builder::default()
         .header(header)
         .clang_args(args)
@@ -49,7 +54,7 @@ fn ossl_bindings(header: &str, args: &[&str]) {
         .allowlist_item("LN_aes.*")
         .generate()
         .expect("Unable to generate bindings")
-        .write_to_file("src/ossl/bindings.rs")
+        .write_to_file(out_path.join("ossl_bindings.rs"))
         .expect("Couldn't write bindings!");
 }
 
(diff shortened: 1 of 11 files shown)


Jakuje:
https://bugzilla.redhat.com/show_bug.cgi?id=2312825

> I'm sorry, building Rust packages (or any other packages that use generated BuildRequires for that matter) outside of mock (i.e. "fedpkg local" or "rpmbuild -bb") will just never be supported. So I don't see a reason to make package changes here.

Ugh ... will need some more fighting with rpmbuild

packaging/kryoptic.spec
```diff
@@ -38,27 +39,29 @@ A PKCS #11 software token written in Rust.}
...
-CONFDIR=%{_sysconfdir} %cargo_build -f dynamic,nssdb,standard
+export CONFDIR=%{_sysconfdir}
+%cargo_build -f dynamic,nssdb,standard
 %{cargo_license_summary -f dynamic,nssdb,standard}
 %{cargo_license -f dynamic,nssdb,standard} > LICENSE.dependencies
 
 %install
 %cargo_install -f dynamic,nssdb,standard
-install -Dp target/rpm/libkryoptic_pkcs11.so $RPM_BUILD_ROOT/%{_libdir}/libkryoptic_pkcs11.so
+install -Dp target/rpm/libkryoptic_pkcs11.so $RPM_BUILD_ROOT/%{_libdir}/pkcs11/libkryoptic_pkcs11.so
```
simo5, line 60: shouldn't the dir be something like %{_libdir}/kryoptic/libkryoptic_pkcs11.so ?

simo5 (commented):
some nits

packaging/kryoptic.spec
Jakuje: The PKCS#11 packaging guidelines say that the directory should be `pkcs11`. It makes it easier at least for the p11-kit, but also for discoverability:
https://docs.fedoraproject.org/en-US/packaging-guidelines/Pkcs11Support/
We could also consider shipping the `p11-kit` module to load kryoptic automatically, which might be also useful, but I was too lazy to add it so far. I can add it to this PR too

Jakuje:
> hmm. The `getrandom` crate is needed only in FIPS mode and the update is not compatible. Will have to figure out this yet.

The API is quite trivial and our use is also only in one place so I think we should be good to update to 0.3. It will keep the Fedora package simpler as we will not have to depend on older/compat version of this crate.

build.rs
```diff
@@ -49,7 +54,7 @@ fn ossl_bindings(header: &str, args: &[&str]) {
         .allowlist_item("LN_aes.*")
         .generate()
         .expect("Unable to generate bindings")
-        .write_to_file("src/ossl/bindings.rs")
+        .write_to_file(out_path.join("ossl_bindings.rs"))
```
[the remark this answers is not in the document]
simo5: yes, but no need to do it in this PR, I may try to just ad a Make target later that is smart enough to figure it out on its own

simo5 (approved):
LGTM

PR #156, opened 2025-02-26
Adding truncated HMACs support

### LINKED ISSUE #55: Implement PBKDF2 with truncated hashes such as `CKP_PKCS5_PBKD2_HMAC_SHA512_224`
> No truncated prf's? such as `CKP_PKCS5_PBKD2_HMAC_SHA512_224` defined in the pkcs11 specs? We don't have these as separate truncated hmacs though ...

> But if the semantics is just truncating the hmacs, what would it take to implement them too?

In any case, we will need some test vectors and clarification if they are acceptable or not in FIPS.

_Originally posted by @Jakuje in https://github.com/latchset/kryoptic/pull/52#discussion_r1609597664_

beldmit:
A naive attempt to fix #55

As OpenSSL defines truncated hashes in FIPS provider, I also add them to FIPS indicators


simo5 (approved):
LGTM,
thanks for the contribution

simo5:
So the integration tests are broken because the new hashes can't be initialized.
In src/ossl/hash.rs we do en explicit match of mechanims type to get the openssl hash name string.

simo5 (changes requested):
To fix the fips builds change:
```
struct FipsChecks {
    keys: [FipsKeyType; 15],
    mechs: [FipsMechanism; 71],
}
```
to contain the correct number of mechanism which grew from 71 to 73 with the addition of the 2 new digests

beldmit:
Will do tomorrow

src/ossl/hash.rs
```diff
@@ -37,20 +37,10 @@ unsafe impl Sync for HashState {}
...
-            CKM_SHA512_256 => OSSL_DIGEST_NAME_SHA2_512_256,
-            CKM_SHA3_224 => OSSL_DIGEST_NAME_SHA3_224,
-            CKM_SHA3_256 => OSSL_DIGEST_NAME_SHA3_256,
-            CKM_SHA3_384 => OSSL_DIGEST_NAME_SHA3_384,
-            CKM_SHA3_512 => OSSL_DIGEST_NAME_SHA3_512,
-            _ => return Err(CKR_MECHANISM_INVALID)?,
-        };
+        let alg: *const c_char = mech_type_to_digest_name(mech);
+        if alg.is_null() {
+            return Err(CKR_MECHANISM_INVALID.into());
```
simo5, line 42: change to:
`return Err(CKR_MECHANISM_INVALID)?;`

simo5 (changes requested):
Just one nit and we are good to go.

PR #157, opened 2025-02-27
Some more make targets

beldmit:
+ make clean
+ make tags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,3 +26,9 @@ tests: build
 
 docs:
 	cargo doc --features standard,nssdb,jsondb,fips --document-private-items
+
+tags:
+	ctags -R src/
+
+clean:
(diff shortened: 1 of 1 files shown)


simo5:
A simple make ctags will not encompass bindings, I have a more complicated commit here that I am plannign to add, that said I use "scope" from https://github.com/chrissicool/scope-rs which produces better rust tags normally (because I use cscope integration in my vim setup) and should give you also proper ctags.

beldmit:
I can drop tags part but I think it's better have it now and you can replace it with the proper one when time permits

simo5:
Sorry mistakenly clciked the merge button, can you rebase on main and squash into a single commit?

simo5 (approved):

PR #158, opened 2025-02-27
Fix compilation warnings

beldmit:


Jakuje:
There some rust equivalent to gcc's `-Werror` to make sure we can catch these without manual review of each and every run. I think it would be good for CI as usually every new change brings some warning in and catching them later is painful:

https://stackoverflow.com/questions/29814978/is-there-a-rustc-equivalent-of-wall-werror
https://users.rust-lang.org/t/how-to-fail-a-build-with-warnings/2687

beldmit:
I'm not a big fun of -Werror, speaking frankly. Having it in Makefile gives 100% probability that the build will fail on compiler update

simo5 (approved):
LGTM

simo5:
please squash all fixup commits

Issue #159, opened 2025-02-27
Do not return CKR_ATTRIBUTE_TYPE_INVALID for missing CKA_ALLOWED_MECHANISMS


Jakuje:
Kryoptic currently returns `CKR_ATTRIBUTE_TYPE_INVALID` when querying for attribute  `CKA_ALLOWED_MECHANISMS` on objects that were created without this attribute. This is wrong as it makes the caller think the attribute is invalid for all the objects.

The specs are vague in here what we should return in case we do not have any restrictions here. The SoftHSM returns zero-length buffer. The closest wording in the specs I found is

> The generation of the KCV may be prevented by the application supplying the attribute in the template as a *no-value (0 length) entry*. The application can query the value at any time like any other attribute using C_GetAttributeValue. C_SetAttributeValue may be used to destroy the attribute, by supplying no-value.

Referring to 0-length entry, meaning NULL result pointer and 0 length.

simo5:
I think it is a general bug in kryoptic to return CKR_ATTRIBUTE_TYPE_INVALID just because an attribute is missing
IIRC CKR_ATTRIBUTE_TYPE_INVALID should be returned only if the attribute is invalid for the object type, not if it is merely empty/missing.

simo5:
https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/cs01/pkcs11-spec-v3.1-cs01.html#_Toc111203287

This is the part of the spec to be carefully interpreted, my thinking is that a missing (but valid) attribute should cause CK_UNAVAILABLE_INFORMATION in ulValueLen but no error.

Jakuje:
> https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/cs01/pkcs11-spec-v3.1-cs01.html#_Toc111203287

Looking at this section, I think the closest option would be the point 3:

> 3.     Otherwise, if the pValue field has the value NULL_PTR, then the ulValueLen field is modified to hold the exact length of the specified attribute for the object.

Doing the `CK_UNAVAILABLE_INFORMATION` implies that the attribute is either invalid (this issue) or sensitive (not this case) with respective return codes. What we really want to say is that it is valid attribute, but it is really empty (NULL_PTR). But the second part of the sentence is a bit unclear to me, but I would interpret it as zero-length field ...

simo5:
Point 3 is for the case when the caller does not know how big is the attribute so they are basically asking us to tell them.
[...]

PR #160, opened 2025-02-28
objects: Do not return CKR_ATTRIBUTE_TYPE_INVALID for unavailable attributes

Jakuje:
Fixes: #159


Jakuje:
as expected, the pkcs11-provder integration test fails ... will dig into that later

simo5:
The change looks good to me, hopefully the pkcs11-provider setup failures are easy to solve.

simo5 (approved):
LGTM

PR #161, opened 2025-02-28
Silly way to generate cscope database

simo5:
As we have to use OUT_DIR, there is no simple way to index the generated files as they are stored in a location that changes freuently due to the hash in the target directory.

At build time store the file names of the generate files so that if the file is present they can be indexed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /*.sql
 cscope.*
 tags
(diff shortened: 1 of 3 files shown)


simo5:
Aaand this breaks because we are now writing a file that has nothing to do with modifying the source directory :/

Issue #163, opened 2025-03-03
CloseSession does not seem to Logout user


ZoltanFridrich:
When I do C_OpenSession, C_Login, C_CloseSession and then again C_OpenSession, C_Login, the login fails with CKR_USER_ALREADY_LOGGED_IN.
From pkcs11 spec: CloseSession: "If this function is successful and it closes the last session between the application and the token, the login state of the token for the application returns to public session"
It seems that kryoptic does not logout the user after C_CloseSession.

Jakuje:
I confirm that this is an issue. With the following reproducer, I can see what you describe:
```
diff --git a/src/tests/login.rs b/src/tests/login.rs
index 0bbc2f4..1218d30 100644
--- a/src/tests/login.rs
+++ b/src/tests/login.rs
@@ -138,3 +138,41 @@ fn test_login() {
 
     testtokn.finalize();
 }
+
+#[test]
+#[parallel]
+fn test_login_close() {
+    let mut testtokn = TestToken::initialized("test_login_close", None);
+
+    /* Run this twice and make sure the second call does not return ALREADY_LOGGED_IN */
+    for _ in 0..2 {
+        let session = testtokn.get_session(true);
+
+        let mut info = CK_SESSION_INFO {
+            slotID: CK_UNAVAILABLE_INFORMATION,
+            state: CK_UNAVAILABLE_INFORMATION,
+            flags: 0,
+            ulDeviceError: 0,
+        };
+        let ret = fn_get_session_info(session, &mut info);
+        assert_eq!(ret, CKR_OK);
+        assert_eq!(info.state, CKS_RW_PUBLIC_SESSION); <<<<<< error on this line shows the session is already CKS_RW_USER_FUNCTIONS
+
+        /* login */
+        let pin = "12345678";
+        let ret = fn_login(
+            session,
+            CKU_USER,
+            pin.as_ptr() as *mut _,
+            pin.len() as CK_ULONG,
+        );
+        assert_eq!(ret, CKR_OK);
+
+        let ret = fn_get_session_info(session, &mut info);
+        assert_eq!(ret, CKR_OK);
+        assert_eq!(info.state, CKS_RW_USER_FUNCTIONS);
+
+        /* close session should reset the login state */
+        testtokn.close_session();
+    }
+}
```
With error:
```
thread 'tests::login::test_login_close' panicked at src/tests/login.rs:159:9:
assertion `left == right` failed
  left: 3
 right: 2
[...]

PR #165, opened 2025-03-03
Logout when last session is closed on given token

Jakuje:
> If this function is successful and it closes the last session between
> the application and the token, the login state of the token for
> the application returns to public sessions. Any new sessions to the token
> opened by the application will be either R/O Public or R/W Public sessions.

https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203273

Fixes: #163

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -256,6 +256,10 @@ impl State {
         };
         self.get_slot_mut(slot_id)?.drop_session(handle);
         self.sessionmap.remove(&handle);
+        /* The specs requires the last session to logout the token */
+        if !self.has_sessions(slot_id)? {
+            self.get_token_from_slot_mut(slot_id)?.logout();
+        }
         Ok(())
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -731,6 +731,8 @@ extern "C" fn fn_close_all_sessions(slot_id: CK_SLOT_ID) -> CK_RV {
     for handle in dropped_sessions {
         token.drop_session_objects(handle);
     }
+    /* The spec requires the token to logout once the last session is closed */
+    token.logout();
     CKR_OK
 }
 
--- a/src/tests/attrs.rs
+++ b/src/tests/attrs.rs
@@ -171,10 +171,24 @@ fn test_set_attr_rsa() {
         fn_set_attribute_value(session, handle, template.as_ptr() as *mut _, 1);
     assert_eq!(ret, CKR_SESSION_READ_ONLY);
 
-    let session = testtokn.get_session(true);
+    /* Open second, read-write, session. Keep the first around to make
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -256,6 +256,9 @@ impl State {
         };
         self.get_slot_mut(slot_id)?.drop_session(handle);
         self.sessionmap.remove(&handle);
+        if !self.has_sessions(slot_id)? {
+            self.get_token_from_slot_mut(slot_id)?.logout();
+        }
```
simo5, line 261: You need to add a call to logout also in drop_all_sessions_slot() I think. Otherwise calling C_CloseAllSessions will leave the token logged in.

Probably we should add a test for that one too.

simo5 (changes requested):

src/lib.rs
simo5: Note that it may make sense to put this into a helper function to be called at the end of fn_close_session() and fn_close_all_sessions() instead of in drop_session(), as that will make the policy behavior more evident.

Jakuje: I consider this more like a specification than a policy that should need to be evident from reading the code in the lib.rs. Especially in case of CloseAllSessions, we do not need to check if there are some leftover sessions, making the logout unconditional.
For the CloseAllSessions, I put the logout directly into lib though as this was the place where I was able to easily access the token. In the CloseSession, I would have to pull much more stuff (slot_id, slot) to check there are some other sessions, which was already available in the `drop_session()` and would duplicate a lot of code in there.

simo5 (changes requested):
small changes

src/lib.rs
```diff
@@ -730,6 +730,7 @@ extern "C" fn fn_close_all_sessions(slot_id: CK_SLOT_ID) -> CK_RV {
     for handle in dropped_sessions {
         token.drop_session_objects(handle);
     }
+    token.logout();
```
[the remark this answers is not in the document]
Jakuje: Added comment to both of the changes.

src/tests/attrs.rs
```diff
@@ -184,8 +198,12 @@ fn test_set_attr_rsa() {
...
-        fn_set_attribute_value(session, handle, template.as_ptr() as *mut _, 1);
+    let ret = fn_set_attribute_value(
+        session2,
+        handle,
+        template.as_ptr() as *mut _,
+        1,
+    );
     assert_eq!(ret, CKR_ATTRIBUTE_READ_ONLY);
 
     testtokn.finalize();
```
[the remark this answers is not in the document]
Jakuje: Yes. Thank you. Done.

simo5 (approved):

PR #166, opened 2025-03-03
Change default feature

### LINKED ISSUE #162: Default build supports no PK mechanisms
Trying to create a new CKO_PUBLIC_KEY object of key type CKK_RSA fails with CKR_ATTRIBUTE_VALUE_INVALID.
To build kryoptic, I used the same instructions that are written in README.md of this project, however it seems that the default build has no support for any PK mechanisms and thats why my attempt to create an RSA key fails. It would be nice to have a better default build.

simo5:
The README.md was not updated when we last change defaults.

This commit restores the default build to point to the "standard" feature. This makes it slightly harder to build non standard configurations, but makes it easier for new-comers.

Fixes #162

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,20 +69,20 @@ tlskdf = []
 # Databases
 jsondb = ["memorydb"]
 memorydb = ["aes", "hkdf", "pbkdf2"] # for encryption
-sqlitedb = [ "dep:rusqlite", "aes", "hkdf", "pbkdf2" ]
-nssdb = [ "dep:rusqlite", "aes", "hmac", "pbkdf2" ]
+sqlitedb = ["dep:rusqlite", "aes", "hkdf", "pbkdf2"]
+nssdb = ["dep:rusqlite", "aes", "hmac", "pbkdf2"]
 
-default = [ "sqlitedb" ]
+default = ["standard"]
 
 ecc_all = ["ecdsa", "ec_montgomery", "eddsa", "ecdh"]
 hash_all = ["hash", "hmac"]
-kdf_all = [ "hkdf", "pbkdf2", "sp800_108", "sshkdf", "tlskdf" ]
+kdf_all = ["hkdf", "pbkdf2", "sp800_108", "sshkdf", "tlskdf"]
 
-standard = [ "ecc_all", "hash_all", "kdf_all", "rsa"]
+standard = ["sqlitedb", "ecc_all", "hash_all", "kdf_all", "rsa"]
 
 ecc_fips = ["ecdsa", "ecdh"]
-fips = [ "rusqlite/bundled", "aes", "ecc_fips", "hash_all", "kdf_all", "rsa"]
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -21,6 +23,16 @@ For FIPS module, you need to generate hmac checksum:
 
     $ ./hmacify.sh target/release/libkryoptic_pkcs11.so
 
+The default build specifies "standard" as the default feature for
+ease of use. "Standard" pulls in all the standard algorithms and the
+sqlitedb storage backedn.
```
Jakuje, line 28:
```suggestion
sqlitedb storage backend.
```

Cargo.toml
```diff
@@ -72,16 +72,16 @@ memorydb = ["aes", "hkdf", "pbkdf2"] # for encryption
...
 ecc_all = ["ecdsa", "ec_montgomery", "eddsa", "ecdh"]
 hash_all = ["hash", "hmac"]
 kdf_all = [ "hkdf", "pbkdf2", "sp800_108", "sshkdf", "tlskdf" ]
 
-standard = [ "ecc_all", "hash_all", "kdf_all", "rsa"]
+standard = [ "sqlitedb", "ecc_all", "hash_all", "kdf_all", "rsa" ]
 
 ecc_fips = ["ecdsa", "ecdh"]
-fips = [ "rusqlite/bundled", "aes", "ecc_fips", "hash_all", "kdf_all", "rsa"]
+fips = [ "aes", "ecc_fips", "hash_all", "kdf_all", "rsa"]
```
Jakuje, line 84: nit: Do we have style rule if we should keep whitespace before closing square brace? I see half of the cases have, the other does not and you add one in the `standard` case. It would be good to have this systematic

README.md
```diff
@@ -21,6 +23,16 @@ For FIPS module, you need to generate hmac checksum:
...
+
+In order to make a different selection you need to use the cargo
+switch to disable default features (--no-default-features) and then
+specify the features you want to build with:
+
+eg: cargo build --no-default-features --features fips,sqlitedb,nssdb
```
Jakuje, line 34: this example line does not match what is execuded in CI nor in the Makefile, where the `rusqlite/bundled` is used too. I see you removed it from the `fips` feature but it is not clear to me why it can not stay under the "fips" feature to make the feature selection easier for the user.

Cargo.toml
simo5: We never set style rules for toml, I wonder if there is a simple tool for enforcing

README.md
simo5: the reason I had removed it is that in theory you could have fips with a non-sqlite db, but currently we support only sqlite or nssdb ... so ... I'll put it back

Jakuje: My point was that make has the following line:
```
cargo build --no-default-features --features fips,rusqlite/bundled,sqlitedb,nssdb
```
but the readme this one:
```
cargo build --no-default-features --features fips,sqlitedb,nssdb
```
I did not test either of them, but either one is needlessly complicated or the other is missing something.

simo5: Yeah, I forgot to change the Makefile and tests with the last push :)

PR #167, opened 2025-03-04
ci: Cache DNF downloads

Jakuje:
Sounds like the latest fedora changed the location of metadata and package downloads.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            /var/cache/dnf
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            /var/cache/dnf
+            /var/cache/libdnf5
```
simo5, line 31: should we be more aggressive and use something like /var/cache/*dnf* ?

simo5 (approved):
Nit but feel free to merge, LGTM

PR #169, opened 2025-03-04
Fix DNF cache again

Jakuje:
* changes the path in the remaining workflows
 * Store the cache only from the workflow that has the most dependencies to make it effective


simo5 (approved):
LGTM!

PR #171, opened 2025-03-04
Reduce the number of unimplemnetd functions

### LINKED ISSUE #168: Handle parallel session operations
The current code allows only one operation per session, however the PKCS#11 spec implies that you can have one operation of each type (base on CK_FLAGS) in flight at the same time. This is evident ion the text of C_CancelSession for example where you can explicitly cancel a specific operation type by using appropriate flags.

Change Kryoptic to allow parallel operations of different type.

Note that this change is also precondition for implementing combine operations like C_SignEncryptUpdate which need to be able to run two opearations in parallel

simo5:
Also refactors how operations are handled to fix #168

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2743,20 +2743,22 @@ extern "C" fn fn_generate_random(
     ret_to_rv!(get_random_data(data))
 }
 
-/// Implementation of C_GetFunctionStatus function (Not Implemented Yet)
+/// Implementation of C_GetFunctionStatus function
+/// (Legacy function. Always returns `CKR_FUNCTION_NOT_PARALLEL`)
 ///
 /// Version 3.1 Specification: [https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203361](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203361)
 
 extern "C" fn fn_get_function_status(_session: CK_SESSION_HANDLE) -> CK_RV {
-    CKR_FUNCTION_NOT_SUPPORTED
+    CKR_FUNCTION_NOT_PARALLEL
 }
 
-/// Implementation of C_CancelFunction function (Not Implemented Yet)
+/// Implementation of C_CancelFunction function
+/// (Legacy function. Always returns `CKR_FUNCTION_NOT_PARALLEL`)
 ///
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -2743,20 +2625,20 @@ extern "C" fn fn_generate_random(
...
 }
 
-/// Implementation of C_CancelFunction function (Not Implemented Yet)
+/// Implementation of C_CancelFunction function
 ///
 /// Version 3.1 Specification: [https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203362](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203362)
 
 extern "C" fn fn_cancel_function(_session: CK_SESSION_HANDLE) -> CK_RV {
-    CKR_FUNCTION_NOT_SUPPORTED
+    CKR_FUNCTION_NOT_PARALLEL
```
Jakuje, line 2641: I think for these two it might make sense to add a comment that these are legacy function really not doing anything useful. From reading the code, it is clear, but it might not be from reading just the doc text. Some note like this could do instead of the previously mention of  "not Implemented".

> Legacy function. Always returns `CKR_FUNCTION_NOT_PARALLEL`

src/lib.rs
```diff
@@ -2990,15 +2872,67 @@ extern "C" fn fn_login_user(
...
+    let mut session = res_or_ret!(rstate.get_session_mut(s_handle));
+    let mut res = CKR_OK;
+
+    if flags & CKF_MESSAGE_ENCRYPT != 0 {
+        if session.cancel_operation::<dyn Encryption>().is_err() {
+            res = CKR_OPERATION_CANCEL_FAILED;
+        }
+    }
+    if flags & CKF_MESSAGE_DECRYPT != 0 {
+        if session.cancel_operation::<dyn Decryption>().is_err() {
```
Jakuje, line 2893:
```suggestion
        if session.cancel_operation::<dyn MsgEncryption>().is_err() {
            res = CKR_OPERATION_CANCEL_FAILED;
        }
    }
    if flags & CKF_MESSAGE_DECRYPT != 0 {
        if session.cancel_operation::<dyn MsgDecryption>().is_err() {
```
?

simo5: oops

Jakuje (approved):
Just one typo in commit  message: `s/signle/single/`. Otherwise lgtm

PR #173, opened 2025-03-07
Combined functions support

simo5:
The code for this functions was all already available, just a matter of wiring up the calls.

Implements digest/encrypt, decrypt/digest, sign/encrypt, decrypt/verify.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,7 @@ mod storage;
 mod token;
 
 use config::Config;
-use error::Result;
+use error::{arg_bad, Result};
 use interface::*;
 use mechanism::*;
 use rng::RNG;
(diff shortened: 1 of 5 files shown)


simo5:
> just typos for now. I will have to wrap my head around the operations next week

Note that I may add some de-duplication via a helper function for encryption/decryption, but haven't done it so far because just copy/paste looked cleaner.

src/lib.rs
```diff
@@ -2094,31 +2143,22 @@ extern "C" fn fn_decrypt_digest_update(
...
-    let outlen = res_or_ret!(operation.decrypt_update(enc, dpart));
-    let retlen = cast_or_ret!(CK_ULONG from outlen);
-    unsafe { *pul_part_len = retlen };
-    let retpart: &mut [u8] =
-        unsafe { std::slice::from_raw_parts_mut(part, outlen) };
-    let operation = res_or_ret!(session.get_operation::<dyn Digest>());
-    ret_to_rv!(operation.digest_update(retpart))
+
+    let part_len: CK_ULONG = unsafe { *pul_part_len };
+    ret_to_rv!(internal_digest_update(&mut session, part, part_len,))
```
Jakuje, line 2161: trailing comma in inline argument list is ok from both functional point of view as well as rust-fmt, but distracting to read. Would be good to get rid of that. The same in `fn_digest_encrypt_update()` and other dual functions. Is there some linter that could enforce these? I saw clippy in the rust-openssl, but did not look into that much further ...

src/lib.rs
```diff
@@ -2137,28 +2177,21 @@ extern "C" fn fn_sign_encrypt_update(
...
     res_or_ret!(session.check_op::<dyn Sign>());
-    let operation = res_or_ret!(session.get_operation::<dyn Encryption>());
-    let plen = cast_or_ret!(usize from part_len => CKR_ARGUMENTS_BAD);
+
+    res_or_ret!(internal_encrypt_update(
+        &mut session,
+        part,
+        part_len,
+        encrypted_part,
+        pul_encrypted_part_len
```
Jakuje, line 2188: nit: on the other hand, I would expect trailing comma here

src/lib.rs
simo5: I with rust-format would better handle these, it is its doing ...
I write freestyle then call `make fix-format` before pushing ...

Jakuje (approved):
regardless of the formatting changes, looks good!

simo5:
Fixed style issues and merged, thanks!

PR #175, opened 2025-03-10
Add a release script in preparation for v1.0

simo5:
Create a misc directory for stuff that is useful but that we do not want to clutter the root tree.

Add a release script that can be used after a release version has been created.

This script checks the version we are trying to make a release of is correctly set in Cargo.toml, and then proceeds to create matching signed tag and signed tarballs for the source and all the dependencies which are retrieved via cargo vendor.

Requires that the person running it has a properly setup gpg key suitable for signing releases.

Currently does not push the tag to github and requires that the release manager has permission to push the tag directly to the tree.


misc/release.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo "Please provide the release version"
+fi
+
+VERSION="$1"
+TAG="v${VERSION}"
+
+#Ensure the release is correct
+git grep "version = \"${VERSION}\"" || (echo "version mismatch, check Cargo.toml" && false)
+
+echo "Craring version tag"
```
Jakuje, line 13: wondering why the codespell is so bad at caching such simple typos ....
```suggestion
echo "Creating version tag"
```

Jakuje (approved):
just a typo. Otherwise lgtm.

PR #177, opened 2025-03-11
Tooling improvements

simo5:
Improve non code aspects of the project

- Better release script
- PRs template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,26 @@
+#### Description
+
+<!--  The description should give a general overview of the goal of the PR.
+
+Individual commit message should highlight important changes that people may
+want to know about when looking at the commit years later
+-->
+
+#### Checklist
+
+<!-- replace [ ] with [x] to select -->
+<!-- (strike not applicable items with ~~ around the text) -->
+
+- [ ] Test suite updated with functionality tests
+- [ ] Test suite updated with negative tests
+- [ ] Documentation was updated
+- [ ] This is not a code change
+
+#### Reviewer's checklist:
+
+- [ ] Any issues marked for closing are fully addressed
+- [ ] There is a test suite reasonably covering new functionality or modifications
+- [ ] This feature/change has adequate documentation added
+- [ ] A changelog entry is added if the change is significant
+- [ ] Code conform to coding style that today cannot yet be enforced via the check style test
+- [ ] Commits have short titles and sensible text
--- a/misc/release.sh
+++ b/misc/release.sh
@@ -1,7 +1,24 @@
 #!/bin/bash -e
 
-if [ -z "$1" ]; then
-    echo "Please provide the release version"
+SIGN=0
+VERSION=
+
+while getopts 'sh' opt; do
+  case "$opt" in
+    s) SIGN=1
+      ;;
+
+    ?|h)
+      echo "Usage: $(basename $0) [-s] <version>"
+      exit 1
+      ;;
+  esac
+done
+shift "$(($OPTIND -1))"
+
+if [ -z "$1"]; then
+    echo "Usage: $(basename $0) [-s] <version>"
+    exit 1
 fi
 
 VERSION="$1"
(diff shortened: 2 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
+#### Description
+
+<!--  The description should give a general overview of the goal of the PR.
+
+Individual commit message should highlight important changes that people may
+want to know about year later when sorting thorough the code.
```
Jakuje, line 6:
```suggestion
want to know about year later when sorting through the code.
```
?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+-->
+
+#### Checklist
+
+<!-- replace [ ] with [x] to select -->
+<!-- (strike not applicable items with ~~ around the text) -->
+
+- [ ] Code was modified
```
Jakuje, line 14:
```suggestion
- [ ] Code was modified for feature?
```
Without the addition this checkbox really sounds to me like meaningless checkbox ...

Jakuje (approved):

.github/PULL_REQUEST_TEMPLATE.md
simo5: I copied it from pkcs11-provider and butchered it by mistake :)
Will fix

misc/release.sh
```diff
@@ -10,14 +27,18 @@ TAG="v${VERSION}"
...
+    echo "Creating version tag"
+    git tag -s ${TAG} -m "Release \"${VERSION}\""
+fi
 
 echo "Creating archives"
-git archive --prefix kryoptic-${VERSION}/ -o kryoptic-${VERSION}.tar.xz ${TAG}
+git archive --format=tar.gz --prefix kryoptic-${VERSION}/ -o kryoptic-${VERSION}.tar.gz ${TAG}
 cargo vendor
-tar --transform "s#^vendor#kryoptic-${VERSION}/vendor#" -czf kryoptic-vendor-${VERSION}.tar.xz vendor
+tar --transform "s#^vendor#kryoptic-${VERSION}/vendor#" -czf kryoptic-vendor-${VERSION}.tar.gz vendor
```
Jakuje, line 38: I think the vendor tarball should be named as `%{name}-%{version}-vendor.tar.gz` (that is at least what we have in RHEL's rust-sequoia-sq package now). I am thinking about adding also a rebuild in centos to test this code path and make sure the vendor tarball also works.

Jakuje: at least that is what the instructions in the spec file say:
```
cargo download %%{crate}==%%{version} > %%{crate}-%%{version}.crate
tar xf %%{crate}-%%{version}.crate
cargo update ... # optional, if you want to update specific dependencies
pushd %%{crate}-%%{version}
cargo vendor-filterer --platform x86_64-unknown-linux-gnu \
                      --platform powerpc64le-unknown-linux-gnu \
                      --platform aarch64-unknown-linux-gnu \
                      --platform i686-unknown-linux-gnu \
                      --platform s390x-unknown-linux-gnu \
                      --all-features=true
```
Should we do the vendoring manually then or try to stick the the cargo tooling?

simo5: 1) tarball names:
   do we have any guideline anywhere? Otheriwise it looks like an arbitrary choice and I am ok with vendor before version, it is more natiral to have the version at the end.

2) tooling:
   I would use cargo tooling as much as possible to generate the right stuff, but I do not want to depend on crates.io which is why I generate the vendor directory, someone with the two tarballs (and the system dependencies necessary) should be able to make a hermetic build (no access to the netwrok).

3) transform:
   If you look into the tarball, if you do not do the transform you get "vendor" as the tarball root directory ... it sounds wrong to me, I think someone that just want to manually inspect the two tarballs and simply -xf them, should get everything under /kryoptic-<version> and not one under that and the other under an unversioned /vendor. However if the tools can't deal with that packaging I can begrudgingly change that.

misc/release.sh
```diff
@@ -1,7 +1,24 @@
...
+
+    ?|h)
+      echo "Usage: $(basename $0) [-s] <version>"
+      exit 1
+      ;;
+  esac
+done
+shift "$(($OPTIND -1))"
+
+if [ -z "$1"]; then
```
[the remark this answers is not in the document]
simo5: This is just a style issue right ?
Because my testing shows that the script works correctly.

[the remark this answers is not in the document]
Jakuje: it works, but throws error (which is ignored as the exit code of the condition does not make the failure of the whole script):
```
./misc/release.sh: line 19: [: missing `]'
```

misc/release.sh
Jakuje: Naming: I did not find the naming explained anywhere, but what rust2rpm generates is as follows `"{project_name}-{project_version}-vendor.tar.xz"`:

https://pagure.io/fedora-rust/rust2rpm/blob/main/f/rust2rpm/vendor.py#_44

Tooling: We could use rust2rpm to generate the vendor tarball.

I think its not that the tools would not be able to handle that, but it will be more complicated. If we take the tarball as it is, the tooling will make it quite straightforward, see #178.

simo5: It seems you have a better handle than I have with this aspects.
Do you want to take over this task in #178 and we close this PR ?

Jakuje: All what I learned this afternoon while experimenting with CI rpmbuild ;)

Sure, we can take over this into the #178 if you are ok with that approach (I pulled your commits into that branch).  I can see if we can get the vendoring simplified with rust2rpm tool.

simo5: it seem those changes are required to get the RHEL rpm builds right, and while the lack of prefix in the vendor tarball irks me, it is not a big deal.

simo5:
Obsoleted by #178

PR #178, opened 2025-03-11
Update packaging template with final release

Jakuje:
Mostly adding gpgcheck and links to the final release.


Jakuje:
will have to figure out the source for the rpm build as for now it is using the crate. Likely based on the script from #177.

packaging/kryoptic.spec
```diff
@@ -109,8 +109,12 @@ echo "module: %{soname}.so" > $RPM_BUILD_ROOT%{_datadir}/p11-kit/modules/kryopti
 
 %if %{with check}
 %check
+%if 0%{?rhel}
+%cargo_test -f dynamic,nssdb,standard -- -- --exact --skip tests::signatures::test_rsa_signatures
```
simo5, line 113: I deally we should fix the sha1 tests instead of skipping them all, but that would take more time, tagging this comment to issue #181 I filed to fix this in general so that we do not forget to change this back.

simo5 (approved):
LGTM

PR #180, opened 2025-03-12
Rebase openssl submodule to 3.5 (prerelease)

simo5:
Rebases the openssl submodule from a version based on Openssl 3.2 to a pre-release of the code that will go in Openssl 3.5.0 with additional patches from Red Hat for fips compliance nd sha1 hardening.

The fips indicator stuff had to be overhauled.

Note:
Two test do not correctly report unapproved fips indicators and will need more investigation.
It may be due to incomplete patching on the openssl side wrt indicators we used to check previously. Keeping PR as draft until openssl branches upstream and thse two test are investigated.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -62,12 +62,25 @@ fn build_ossl(out_file: &Path) {
         .canonicalize()
         .expect("cannot canonicalize path");
 
+    let str_defines;
+
     #[cfg(feature = "fips")]
     let (libpath, bldargs, header) = {
         let providers_path = openssl_path
             .join("providers")
             .canonicalize()
             .expect("OpenSSL providers path unavailable");
+
+        let version = env!("CARGO_PKG_VERSION");
+        let ver_define =
+            format!("-DKRYOPTIC_FIPS_VERSION=\\\"{}-test\\\"", version);
+        let defines = [
+            "-DDEVRANDOM=\\\"/dev/urandom\\\"",
+            "-DOPENSSL_PEDANTIC_ZEROIZATION",
+            "-DFIPS_VENDOR=\\\"Kryoptic\\\"",
+            &ver_define,
+        ];
+        str_defines = defines.join(" ");
 
         let libfips = format!("{}/libfips.a", providers_path.to_string_lossy());
         let buildargs = [
@@ -80,7 +93,7 @@ fn build_ossl(out_file: &Path) {
             "no-des",
             "no-dsa",
             "no-atexit",
-            "-DDEVRANDOM=\\\"/dev/urandom\\\" -DOPENSSL_PEDANTIC_ZEROIZATION -DFIPS_VENDOR=\\\"Kryoptic\\\" -DKRYOPTIC_FIPS_VERSION=\\\"1.0.0-test\\\"",
+            &str_defines,
         ];
 
         println!(
@@ -97,6 +110,7 @@ fn build_ossl(out_file: &Path) {
     let (libpath, bldargs, header) = {
         let libcrypto =
             format!("{}/libcrypto.a", openssl_path.to_string_lossy());
+        str_defines = format!("-DDEVRANDOM=\\\"/dev/urandom\\\"");
         let buildargs = [
             "--debug",
             "no-mdc2",
@@ -104,7 +118,7 @@ fn build_ossl(out_file: &Path) {
             "no-sm2",
             "no-sm4",
             "no-des",
-            "-DDEVRANDOM=\\\"/dev/urandom\\\"",
+            &str_defines,
         ];
 
         println!("cargo:rustc-link-search={}", openssl_path.to_str().unwrap());
--- a/build.rs
+++ b/build.rs
@@ -77,7 +77,10 @@ fn build_ossl(out_file: &Path) {
             "no-ec2m",
             "no-sm2",
             "no-sm4",
-            "-DREDHAT_FIPS_VERSION=\\\"0.0.1-test\\\"",
+            "no-des",
+            "no-dsa",
+            "no-atexit",
+            "-DDEVRANDOM=\\\"/dev/urandom\\\" -DOPENSSL_PEDANTIC_ZEROIZATION -DFIPS_VENDOR=\\\"Kryoptic\\\" -DKRYOPTIC_FIPS_VERSION=\\\"1.0.0-test\\\"",
         ];
 
         println!(
@@ -94,12 +97,21 @@ fn build_ossl(out_file: &Path) {
     let (libpath, bldargs, header) = {
         let libcrypto =
             format!("{}/libcrypto.a", openssl_path.to_string_lossy());
+        let buildargs = [
+            "--debug",
+            "no-mdc2",
+            "no-ec2m",
+            "no-sm2",
+            "no-sm4",
+            "no-des",
+            "-DDEVRANDOM=\\\"/dev/urandom\\\"",
+        ];
 
         println!("cargo:rustc-link-search={}", openssl_path.to_str().unwrap());
         println!("cargo:rustc-link-lib=static=crypto");
         println!("cargo:rerun-if-changed={}", libcrypto);
 
-        (libcrypto, ["--debug"], "ossl.h")
+        (libcrypto, buildargs, "ossl.h")
     };
 
     match std::path::Path::new(&libpath).try_exists() {
--- a/src/ossl/fips.rs
+++ b/src/ossl/fips.rs
@@ -1,8 +1,9 @@
 // Copyright 2023 Simo Sorce
 // See LICENSE.txt file for terms
 
+use std::cell::Cell;
 use std::ffi::CStr;
-use std::ffi::{c_char, c_int, c_uchar, c_uint, c_void};
+use std::ffi::{c_char, c_int, c_uchar, c_void};
 use std::fs::File;
 use std::io::Read;
 use std::io::Seek;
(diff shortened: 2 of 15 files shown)


build.rs
```diff
@@ -77,7 +77,10 @@ fn build_ossl(out_file: &Path) {
             "no-ec2m",
             "no-sm2",
             "no-sm4",
-            "-DREDHAT_FIPS_VERSION=\\\"0.0.1-test\\\"",
+            "no-des",
+            "no-dsa",
+            "no-atexit",
+            "-DDEVRANDOM=\\\"/dev/urandom\\\" -DOPENSSL_PEDANTIC_ZEROIZATION -DFIPS_VENDOR=\\\"Kryoptic\\\" -DKRYOPTIC_FIPS_VERSION=\\\"1.0.0-test\\\"",
```
Jakuje, line 83: should we take the version from the package version here to avoid the need to update version in different places?

src/ossl/fips.rs
```diff
@@ -859,51 +860,69 @@ impl ProviderSignatureCtx {
...
-                Ok(match indicator {
-                    [<EVP_ $up _REDHAT_FIPS_INDICATOR_UNDETERMINED>] => None,
-                    [<EVP_ $up _REDHAT_FIPS_INDICATOR_APPROVED>] => Some(true),
-                    [<EVP_ $up _REDHAT_FIPS_INDICATOR_NOT_APPROVED>] => Some(false),
-                    _ => return Err(CKR_DEVICE_ERROR)?,
-                })
-            }
-        }
-    };
+/* The Openssl FIPS indicator callback is so dump you can't even
```
Jakuje, line 863:
```suggestion
/* The Openssl FIPS indicator callback is so dumb you can't even
```
?

src/tests/rsa.rs
```diff
@@ -415,7 +415,7 @@ fn test_rsa_operations() {
     );
     assert_eq!(ret, CKR_OK);
 
-    #[cfg(not(feature = "fips"))]
+    #[cfg(all(not(feature = "fips"), feature = "dynamic"))]
```
Jakuje, line 418: It is not clear to me the reasoning behind this change. Can you clarify? This was skipped in FIPS mode because it is PKCS#1.5 signature with SHA1 and raw. Now you exclude it also from the static builds, likely because of the new OpenSSL does not let it through? Would not it have the same meaning if you would keep only the second part? We do not support dynamic fips builds:
```suggestion
    #[cfg(feature = "dynamic")]
```
but I think its just matter of time when we will get openssl 3.5 in Fedora and we will need some other change of this condition to work  outside of the box ..

src/tests/signatures.rs
```diff
@@ -156,29 +157,37 @@ fn test_rsa_signatures() {
...
-        session,
-        pri_key_handle,
-        &mut testcase.value,
-        &mut mechanism,
-    ) {
-        Ok(r) => r,
-        Err(e) => panic!("f{e}"),
-    };
-    assert_eq!(testcase.result, result);
+    /* SHA1 is disabled in out openssl submodule */
```
Jakuje, line 163:
```suggestion
    /* SHA1 is disabled in our openssl submodule */
```

src/tests/signatures.rs
```diff
@@ -156,29 +157,37 @@ fn test_rsa_signatures() {
...
-        session,
-        pri_key_handle,
-        &mut testcase.value,
-        &mut mechanism,
-    ) {
-        Ok(r) => r,
-        Err(e) => panic!("f{e}"),
-    };
-    assert_eq!(testcase.result, result);
+    #[cfg(all(not(feature = "fips"), feature = "dynamic"))]
```
Jakuje, line 165: same here. I think the following should be equivalent:
```suggestion
    #[cfg(feature = "dynamic")]
```

build.rs
simo5: For upstream that is ok, yes, downstream we'll probbaly do something different anyway.

src/tests/rsa.rs
simo5: In the static version I have (ossl 3.5 based) we have patches that completely disable SHA1 signatures, while the dynamic version (in CI) uses Fedora's Openssl version which does not disable SHA1 signatures.

Ideally we check what is available and then conditionally run tests, but I just wanted something quick here to get CI green, and defer the decision till later.

Note that these tests would fail on a RHEL system in all cases for example ...

src/tests/signatures.rs
simo5: ah right ...

simo5:
I added a commit to get the fips version from the crate, also fixed the various other issues.
PTAL.

src/tests/aes.rs
```diff
@@ -443,7 +444,8 @@ fn test_aes_operations() {
         assert_eq!(&enc[..12], enc2.as_slice());
 
         /* test that we can get correct indicators based on inputs */
-        assert_eq!(check_validation(session, 0), true);
+        // FIXME: started failing with openssl rebase
+        // assert_eq!(check_validation(session, 0), true);
```
Jakuje, line 448: would prefer to figure out these two failures before merging. Otherwise lgtm.

src/tests/rsa.rs
```diff
@@ -415,7 +415,7 @@ fn test_rsa_operations() {
     );
     assert_eq!(ret, CKR_OK);
 
-    #[cfg(not(feature = "fips"))]
+    #[cfg(feature = "dynamic")]
```
Jakuje, line 418: looking at these changes, its now correct, but completely unreadable why this is enabled only in the dynamic build so I would prefer a comment about what is discussed #181 and in the previous comments why this works only in dynamic builds.
Likely something like in the signatures test would do:
```
    /* SHA1 is disabled in the openssl submodule */
```

simo5: Yeah, I thought the same and then forgot to put the comment, good catch!

simo5:
Updated the openssl submodule to be a descendant of the new upstream openssl-3.5 branch instead of the openssl/master branch

src/tests/aes.rs
Jakuje: Unresolving then to keep them visible around.

src/ossl/sshkdf.rs
```diff
@@ -137,7 +141,15 @@ impl Derive for SSHKDFOperation {
...
-        self.fips_approved = check_kdf_fips_indicators(&mut kctx)?;
+        #[cfg(feature = "fips")]
+        {
+            if ossl_fips_indicator_is_set() {
+                self.fips_approved = Some(false);
+            }
+            if self.fips_approved.is_none() {
+                self.fips_approved = Some(true);
+            }
+        }
```
Jakuje, line 152: This is repeated in several places so reviewing all the cases do the right thing is pain. Given that the `fips_approved` is set on different objects, making a function from that won't easily work (?) so should we go ahead with some macro?

simo5: macros would be worse imo.
But we could create a function that takes &mut arguments in theory ...
Do we need to centralize this?

simo5:
I added a helper function for fips_approval and rebased to latest ossl35 branch_patches, PTAL

src/ossl/fips.rs
```diff
@@ -859,51 +860,83 @@ impl ProviderSignatureCtx {
...
+    return true;
+}
+
+pub fn fips_approval(fips_approved: &mut Option<bool>, finalize: bool) {
+    if ossl_fips_indicator_is_set() {
+        /* The indicator was set, therefore there was an unapproved use */
+        *fips_approved = Some(false);
+    }
+    if finalize {
+        /* this si the last check, mark approval as tue is not set so far */
```
Jakuje, line 934:
```suggestion
        /* this is the last check, mark approval as true if not set so far */
```

src/ossl/fips.rs
```diff
@@ -859,51 +860,83 @@ impl ProviderSignatureCtx {
...
+}
+
+pub fn ossl_fips_indicator_is_set() -> bool {
+    if FIPS_INDICATOR.get() == 0 {
+        return false;
+    }
+    return true;
+}
+
+pub fn fips_approval(fips_approved: &mut Option<bool>, finalize: bool) {
```
Jakuje, line 928: I find the code now much easier to read than going through the branches of fips in every function. I just had to wrap my head around the arguments as the function call `fips_approval(..., true)` is not completely self-explaining.

My hopefully last nit would be to have a helper function `fips_approval_finalize()` and `fips_approval_check()` that might wrap call this function with the bool parameters inside and use these in the code so it is clear on the first sight what is the function doing with the fips approval without the need to look what arguments names it has.

src/tests/aes.rs
```diff
@@ -478,7 +478,7 @@ fn test_aes_operations() {
         assert_eq!(&enc[..12], enc2.as_slice());
 
         /* test that we can get correct indicators based on inputs */
-        assert_eq!(check_validation(session, 0), true);
+        assert_eq!(check_validation(session, 1), true);
```
Jakuje, line 481: the separate issue needs to be filled from this to investigate if this API is really ok to use for FIPS.

src/ossl/fips.rs
simo5: fixed

src/ossl/fips.rs
simo5: done, and added more functions for intialization too

src/tests/aes.rs
simo5: done

Jakuje (approved):

PR #183, opened 2025-03-13
Switch to 2024 edition

simo5:
The migration tool automatically changed some of thcode causing some churn. I manually changed some of the things, like removing expr_2021 from macros as it was not necessary and changing the renaming of the now reserved word gen from r#gen to generator.

A bunch of let Some() {} else () expressions were changed to matches, I let that be, sometime smatch turns out to be idiomatic, sometimes it is about the same readability.

#### Checklist

- [ ] ~Test suite updated with functionality tests~

#### Reviewer's checklist:


Jakuje (approved):
lgmt. I find the matches most of the time easier to read than the `if let Ok() ...` constructs, but at some when there is a lot of code inside, they end up too indented ...

Jakuje:
Should the edition be changed also in Cargo.toml ?

https://github.com/latchset/kryoptic/blob/main/Cargo.toml#L4

simo5:
> Should the edition be changed also in Cargo.toml ?
> 
> https://github.com/latchset/kryoptic/blob/main/Cargo.toml#L4

Well, I am confused, the migration is really about changing it in Cargo.toml which is how I started it ... I am really surprised the MR did not end up having it there, without the Cargo.toml change no change really happened. I will open a followup change to actually do the change then.

simo5:
Filed #187

Issue #179, opened 2025-03-12
`C_EncryptUpdate` fails with `CKR_BUFFER_TOO_SMALL`


xhanulik:
When attempting to encrypt data with `C_EncryptUpdate` multiple-part encryption, it fails in first iteration with `CKR_BUFFER_TOO_SMALL`.
The behavior was spot with `pkcs11-tool`, which reads the input plaintext into `in_buffer` in cycle and provides it to `C_EncryptUpdate`.

```
unsigned char in_buffer[1024], out_buffer[1024];
...
sz = read(fd_in, in_buffer, sizeof(in_buffer));
in_len = sz;
...
rv = p11->C_EncryptInit(session, &mech, key);
do {
	out_len = sizeof(out_buffer);
	rv = p11->C_EncryptUpdate(session, in_buffer, in_len, out_buffer, &out_len);
	if (rv != CKR_OK)
		p11_fatal("C_EncryptUpdate", rv);
	sz = write(fd_out, out_buffer, out_len);
	if ((size_t)sz != out_len)
		util_fatal("Cannot write to %s: %m", opt_output);
	sz = read(fd_in, in_buffer, sizeof(in_buffer));
	in_len = sz;
} while (sz > 0);
```

 For long plaintext, in the first round of multiple-part encryption the input and output lengths are both 1024 bytes. It looks like the kryoptic requires longer output buffer there.

Script with reproducer: [kryoptic-reproducer.txt](https://github.com/user-attachments/files/19207507/kryoptic-reproducer.txt)

simo5:
It depends on the algorithm, what is the mech in this case?

simo5:
Note that CKR_BUFFER_TOO_SMALL is not a fatal error and indicates the buffer is ... too small to complete the operation.

An application that wants to be maximally compatible with modules should call C_EncryptUpdate() with a NULL pointer for out_buffer so that the module can provide what the expected buffer length is.

However I am not saying this bug is invalid, there is probably some cases where there is a mismatch between what some pkcs11 module do for this specific mechanism and what we can do within the limits of our internal calls to openssl that may result in us requesting a buffer slightly larger than what the minimum is, so I need the exact mechanism to be bale to reproduce and verify.

xhanulik:
Sorry, the mechanism is `CKM_AES_CBC_PAD`.

simo5:
@xhanulik does your code expect to get more data when calling C_EncryptFinal ?
[...]

PR #185, opened 2025-03-14
Deal with length query issues

simo5:
#### Description

Fixes various issues with requesting the necessary buffer lengths via behavior 5.2 of the PKCS#11 spec.

#### Checklist

#### Reviewer's checklist:


src/ossl/aes.rs
```diff
@@ -1654,45 +1713,53 @@ impl Encryption for AesOperation {
...
+                    ((self.finalbuf.len() + data_len + AES_BLOCK_SIZE)
+                        / AES_BLOCK_SIZE)
+                        * AES_BLOCK_SIZE
+                }
+                CKM_AES_KEY_WRAP => {
+                    if data_len % 8 != 0 {
+                        return Err(self.op_err(CKR_DATA_LEN_RANGE));
                     } else {
-                        AES_BLOCK_SIZE
+                        data_len + 8
```
Jakuje, line 1746: just a though if we should make a constant out of the 8 here to be clear that this is the AES Key Wrap algorithm block size here and in the above check. And likely also in the KWP mechanism below.
In the decryption case there is already `(AES_BLOCK_SIZE / 2)` used, but having KWP define might make the code easier to read and understand.

Jakuje:
Its much more changes than I expected. But code wise look good. Added one inline nit regarding magic numbers.

There are some typos in the commit messages. We should try to run the spellcheck on that too:
`s/accoutn/account/`
`s/inpout/input/`
`s/ast/as/`
`s/crypton_len/cryption_len/`

src/ossl/aes.rs
simo5: Added a commit with a new constant for this

Jakuje (approved):
lgtm

xhanulik:
For https://github.com/latchset/kryoptic/issues/179 the patch fixes the problem with `pkcs11-tool`.

Issue #182, opened 2025-03-13
Some attributes have defaults, which should return value instead of CK_UNAVAILABLE_INFORMATION


Jakuje:
The generic key objects have the attribute `CKA_START_DATE` with the following description:

> Start date for the key (default empty)

The code in kryoptic defines the attribute correctly with default empty value, which looks like is somewhere converted to unavailable information when queried:

```
11: C_GenerateKey
P:848598; T:0x140243986020032 2025-03-13 18:55:50.499
[in] hSession = 0x2
[in] pMechanism->type = CKM_AES_KEY_GEN              
[in] pMechanism->pParameter[ulParameterLen] NULL [size : 0x0 (0)]
[in] pTemplate[7]: 
    CKA_CLASS             CKO_SECRET_KEY       
    CKA_TOKEN             True
    CKA_SENSITIVE         True
    CKA_VALUE_LEN         00007f8d1408eb70 / 8
    00000000  10 00 00 00 00 00 00 00                          ........        
    CKA_KEY_TYPE          CKK_AES            
    CKA_LABEL             00007f8d14089c10 / 7
    74657374 414553
     t e s t  A E S
    CKA_PRIVATE           True
[out] hKey = 0x1
Returned:  0 CKR_OK

12: C_GetAttributeValue
P:848598; T:0x140243986020032 2025-03-13 18:55:50.500
[in] hSession = 0x2
[in] hObject = 0x1
[in] pTemplate[1]: 
    CKA_END_DATE          0000000000000000 / 0
[out] pTemplate[1]: 
    CKA_END_DATE          0000000000000000 / -1
Returned:  0 CKR_OK
```
Will have to dig into this further to figure out where does it come from.

Jakuje:
The question is, what should the "empty" date look like ...

simo5:
Good question, from the spec (CK_DATE subsection of 3.4 Object Types)

`When a Cryptoki object carries an attribute of this type, and the default value of the attribute is specified to be "empty," then Cryptoki libraries SHALL set the attribute's ulValueLen to 0.`

simo5:
And I think the problem may be that when we specify ulValueLen of 0 we drop the attribute in the storage layer ...

Jakuje:
That was something I would guess could happen. Not sure though where this should be fixed. Should we store empty values? Or should the extracting layer handle that by getting the default values from the factory if not present in the database? I think the second option should be way to go.

simo5:
[...]

Issue #181, opened 2025-03-13
Add way to disble SHA1 entirely


simo5:
When compiled against a static version of openssl that does not support SHA-1 signatures at all we should not offer any SHA-1 signature mechanism.

When dynamically linking against a version of openssl that does not support SHA-1 signatures, similarly, we should also dynamically check that and not offer the mechanisms ...

beldmit:
I can try implementing this

PR #188, opened 2025-03-25
Fix storing and reading default DATE attributes

Jakuje:
#### Description

Turned out the START_DATE and END_DATE attributes had default value defined, but the flag was missing, which lead to not storing the default attributes in DB.

Throughout the chain of storing and reading DATE type, there were several checks for valid dates (which empty string is not) so there were some changes needed to loosen these checks.

Fixes: #182

#### Checklist

- [X] Documentation was updated

#### Reviewer's checklist:

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -414,6 +414,14 @@ const ASCII_DASH: u8 = 0x2D;
 const MIN_ASCII_DIGIT: u8 = 0x30;
 const MAX_ASCII_DIGIT: u8 = 0x39;
 
+fn empty_date() -> CK_DATE {
+    CK_DATE {
+        year: [0x30, 0x30, 0x30, 0x30],
+        month: [0x30, 0x30],
+        day: [0x30, 0x30],
+    }
+}
+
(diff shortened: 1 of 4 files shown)


Jakuje:
sounds like something changed in Fedora dependencies and now we have test failure (only with Fedora build dependencies). The same happens in koschei so I will try to investigate what is going on there:

https://koschei.fedoraproject.org/package/kryoptic

From the update list, it looks like OpenSSL 3.5 does not like the RSA PKCS#1.5 with SHA1 anymore? So related to #181.

Jakuje:
got confirmed from Dima that there are likely some regressions in OpenSSL 3.5 in rawhide so I will ignore the failure for now.

simo5:
LGTM, hopefully we'll fix rawhide's openssl soon, if not we should probably switch to fedora:latest in a future CI update ...

Jakuje:
> LGTM, hopefully we'll fix rawhide's openssl soon, if not we should probably switch to fedora:latest in a future CI update ...

Fedora latest did not have latest rust dependencies. But 42 should be out soon which should have all we need for now.

PR #189, opened 2025-03-25
aes: Restrict AES-GCM to at least 1B tag length

Jakuje:
#### Description

The PKCS#11 specs says 0 is a valid tag length:

> *ulTagBits*
> length of authentication tag (output following cipher text) in bits.
> Can be any value between 0 and 128.

From 6.13.5 AES GCM and CCM Mechanism parameters
https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203499

But OpenSSL's API requires that the tag is at least on byte when
decrypting data:

> EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_AEAD_SET_TAG, taglen, tag)
>   Sets the expected tag to "taglen" bytes from "tag".  The tag length
>   can only be set before specifying an IV.  "taglen" must be between
>   1 and 16 (POLY1305_BLOCK_SIZE) inclusive.  This call is only valid
>   when decrypting data.

When data is encrypted, it is still possible to not read the tag though,
but allowing not to set the tag when decrypting data would probably be
too error prone.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -353,6 +353,11 @@ impl AesOperation {
                 if params.ulTagBits > 128 {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
                 }
+                // The PKCS#11 specification allows 0-length tags, but OpenSSL
+                // does not so we restrict the params to sensible values only
+                if params.ulTagBits < 8 {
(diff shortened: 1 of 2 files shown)


simo5 (approved):
LGTM

Issue #170, opened 2025-03-04
CKF_MULTI_MESSAGE is not set on any mechanism


simo5:
Description:
"True if the mechanism can be used with C_*MessageBegin. One of CKF_MESSAGE_* flag must also be set."

PR #190, opened 2025-03-26
Introduce CKF_MULTI_MESSAGE flag

Jakuje:
#### Description

Fixes: #170

If you think some more tests will be needed, please let me know. I think we support this only for AES-GCM/CCM for now.

#### Checklist
- ~[ ] Test suite updated with negative tests~

#### Reviewer's checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3203,6 +3203,7 @@ extern "C" fn fn_encrypt_message_begin(
 
     let rstate = global_rlock!(STATE);
     let mut session = res_or_ret!(rstate.get_session_mut(s_handle));
+    let slot_id = session.get_slot_id();
 
     #[cfg(feature = "fips")]
     session.reset_fips_indicator();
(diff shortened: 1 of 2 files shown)


simo5:
Should we add a test that ensure this works as expected ?
Ie the GCM mechanism has the flag set and trying to use the Message function with another cipher fails because of lack of the flag ?

Jakuje:
> Should we add a test that ensure this works as expected ? Ie the GCM mechanism has the flag set and trying to use the Message function with another cipher fails because of lack of the flag ?

I was thinking what would be a usable tests, but for now only AES-GCM and AES-CCM works with the message API and both of them have the MULTI_MESSAGE flag. So the only thing I can test is to call init with mechanism that does not have message API (AES-CRT). But this fails already at `fn_message_encrypt_init()` on the missing `CKF_MESSAGE_ENCRYPT` flag. Do you have some other test in you mind that could excercise this code path? Or do you know from top of you head if there is some mechanism that does not support the MULTI_MESSAGE that we can implement/wire for the test?

simo5 (approved):

Issue #193, opened 2025-03-31
Reading attributes from objects should pull more related attributes


Jakuje:
The test that generates non-sensitive key and then tries to read the CKA_VALUE fails:

https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki/tests/basic.rs#L1648

This is because when we query a DB, we pull just small subset of attributes (CKA_CLASS, CKA_KEY_TYPE, CKA_CERTIFICATE_TYPE) and when we filter them, we ask if the object is sensitive. But at that time, we do not have that attribute loaded and it defaults to true.

We should pull more attributes when doing the get_attribute queries to make sure we are able to decide if the attribute is extractable/sensitive ....

simo5:
Uhmm we check only if it is token or private ... sounds like we need to conditionally request also CKA_SENSITIVE and CKA_EXTRACTABLE ...

simo5:
In our test this passes correctly because session objects are returned complete, but for tokens we added optimizations to pull only the attribute explicitly requested, so that then fails the sensitivity check in get_object_attributes() ...

PR #194, opened 2025-03-31
Ensure token store objects can be extracted if the right booleans are set

simo5:
#### Description

The database optimized code was not sourcing the CKA_SENSITIVE/CKA_EXTRACTABLE attributes for token objects, this would cause subsequent checks on these attributes to use the object defaults rather than what was actually stored.
Session objects did not suffer from this issue because all attributes are always available for those.

Fixes: #193

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/storage/format.rs
+++ b/src/storage/format.rs
@@ -227,6 +227,11 @@ impl Storage for StdStorageFormat {
              * not be returned in that case */
             attrs.add_missing_ulong(CKA_KEY_TYPE, &dnm);
             attrs.add_missing_ulong(CKA_CERTIFICATE_TYPE, &dnm);
+            /* We also need to know whether the object is sensible/extractable
+             * so that the token code can decide whether it is ok to return
+             * some attributes or not */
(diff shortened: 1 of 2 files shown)


simo5:
> you were faster :) lgtm. I just wanted to make sure we do not need some more in different code paths, but that can be done later if needed.

Reasonably confident this is all we need, but we can always add more patches later indeed.

Thanks for the review and all these bugs, each fix make kryoptic more robust and compliant, which I like.

Jakuje:
Thinking if this is the right place to add these though. In `get_object_attrs()` we are adding already different attributes for the non-logged in users. I think that would be probably better place ...

simo5:
> Thinking if this is the right place to add these though. In `get_object_attrs()` we are adding already different attributes for the non-logged in users. I think that would be probably better place ...

No, this is the correct place, the one you see in get_objects_attr() is only for the case where a user is not logged in and mixing the use cases would be confusing.

Jakuje:
Ok. Went through the places where else we fetch the data from db and this is the only one where we missed something, I think.

Issue #184, opened 2025-03-14
Calling WrapKey with NULL `pWrappedKey` panics


Jakuje:
The WrapKey operation does not follow the convention described in Section 5.2 (of pkcs#11 specs) on producing output and assumes the wrappedkey pointer is not NULL in 

https://github.com/latchset/kryoptic/blob/606d9ecd9ac0509c737a0c929cd5d3e08dec66a5/src/lib.rs#L2431

```
   6: kryoptic_pkcs11::fn_wrap_key
             at /home/jjelen/devel/kryoptic/src/lib.rs:2431:18
   7: C_WrapKey
             at /home/jjelen/devel/OpenSC/src/pkcs11/pkcs11-spy.c:1505:7
```
We will need to use similar trick as we do in encrypt() to be able to know the length of the output, but it makes it more complicated as this is one-shot operation and we do not have operation state around. This means that we will have to create some `mech.wrap_key_len()` API to bubble inside, do everything except for wrapping and return the length.

Or we will have to change API of `wrap_key` API above to distinguish if this is length query or the wrap operation itself.

simo5:
Yep, I just found this out as I am handling #179 which requires a more invasive change than I initially thought.

But once I am done with it it will yield better and more correct semantics for all encrypt/decrypt operations.

Note: it will take a while to complete.

simo5:
Uhm this one ha an additional gap compared to what is needed for #179 I had not realized, so I will deal with it separately later.

PR #202, opened 2025-04-08
Fix C_WrapKey size query

Jakuje:
#### Description

Fixes `C_WrapKey` size queries that previously crashed kryoptic (#184). 

This is a naive approach, which involves creating 0-length buffer for the size query, doing the export of the key and if the buffer is not large enough, returning.

Better approach would be to store the intermediate state of the operation and continue from there once we will have large enough buffer, but it would involve much more changes.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -937,11 +937,23 @@ impl AesOperation {
                 /* Check the data length in CCM matches the provided data -- this is one-shot
                  * operation only */
                 if op.params.datalen != keydata.len() {
+                    zeromem(keydata.as_mut_slice());
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
                 }
             }
             _ => (),
         }
+
+        let needed_len = op.encryption_len(keydata.len(), true)?;
+        if output.len() == 0 {
+            zeromem(keydata.as_mut_slice());
+            return Ok(needed_len);
+        }
+        if output.len() < needed_len {
+            zeromem(keydata.as_mut_slice());
+            return Err(error::Error::buf_too_small(needed_len));
+        }
+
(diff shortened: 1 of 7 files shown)


src/ossl/rsa.rs
```diff
@@ -431,6 +431,11 @@ impl RsaPKCSOperation {
                 return Err(e);
             }
         };
+        let needed_len = op.encryption_len(keydata.len(), true)?;
+        if output.len() < needed_len {
+            zeromem(keydata.as_mut_slice());
+            return Ok(needed_len);
+        }
```
simo5, line 438: This should return Error::buf_too_small(needed_len);

src/ossl/aes.rs
```diff
@@ -937,11 +937,19 @@ impl AesOperation {
...
                 }
             }
             _ => (),
         }
+
+        let needed_length = op.encryption_len(keydata.len(), true)?;
+        if output.len() < needed_length {
+            zeromem(keydata.as_mut_slice());
+            return Ok(needed_length);
+        }
```
simo5, line 951: This needs to be split in 2 checks,
if output.len() is exactly 0 then returning Ok(needed_length) is correct, for any other value of the output's length see the comment on the rsa file

simo5 (changes requested):
The return errors need to be fixed a bit.

src/ossl/aes.rs
Jakuje: My bad. Fixed now!

src/ossl/rsa.rs
Jakuje: My bad. Fixed now!

simo5 (approved):
LGTM

PR #203, opened 2025-04-11
ecdh: Fix max ECDH output size

Jakuje:
#### Description

For some reason, I set the output length to the 2 * field sizes initially. This does not have any effect on deriving AES keys as they can not be that large, but this could explode for secret keys that could get larger.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/tests/ecdh.rs
+++ b/src/tests/ecdh.rs
@@ -221,6 +221,37 @@ fn test_ecc_derive_plain() {
     assert_eq!(ret, CKR_OK);
     assert_eq!(value, ref_value_full[(ref_value_full.len() - 32)..]);
 
+    /* With GENERIC_SECRET without explicit CKA_VALUE_LEN should
+     * create keys of ECDH derive output (ec group size = 32B) */
+    let derive_template = make_attr_template(
+        &[
+            (CKA_CLASS, CKO_SECRET_KEY),
+            (CKA_KEY_TYPE, CKK_GENERIC_SECRET),
+        ],
+        &[],
+        &[(CKA_SENSITIVE, false), (CKA_EXTRACTABLE, true)],
+    );
+    let mut s_handle = CK_INVALID_HANDLE;
+    let ret = fn_derive_key(
+        session,
+        &mut mechanism,
+        handle,
+        derive_template.as_ptr() as *mut _,
+        derive_template.len() as CK_ULONG,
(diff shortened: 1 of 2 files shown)


Jakuje:
This was caught by the test in https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki/tests/basic.rs#L603 -- I can try to emulate it here if needed  (or integrate rust-cryptoki tests in pipeline once I will fix all of them).

Jakuje:
> You marked that the test suite was updated, but I see no changes to tests, did you mean to say that the test suites already sufficiently cover this change ?

Sorry, misread the title of the checkbox. Fixed. Not covered by any internal tests. Only the ones in rust-cryptoki as mentioned inthe comment. I can add some if you prefer.

simo5:
Yeah I'd prefer at least one test, thanks.

simo5 (changes requested):
Please add 1 test that fails with old code and passes with new, thanks.

src/tests/ecdh.rs
```diff
@@ -221,6 +221,37 @@ fn test_ecc_derive_plain() {
...
+
+    let ret = fn_get_attribute_value(
+        session,
+        s_handle,
+        extract_template.as_mut_ptr(),
+        extract_template.len() as CK_ULONG,
+    );
+    assert_eq!(ret, CKR_OK);
+    assert_eq!(value.len(), ref_value_full.len());
+    assert_eq!(value, ref_value_full[(ref_value_full.len() - 32)..]);
```
simo5, line 253: I am not sure I understand the right hand of the eq here.
If I read the code right ref_value_full is a 32 bytes vector, you should be able to just use ref_value_full ?

simo5 (approved):
LGTM, although a bit confused about why you compare the extracted value this way.

src/tests/ecdh.rs
Jakuje: You are right. This was what I wanted to try and amend and then forgot. Fixed now.

simo5:
Note that the previous test is also doing that odd thing, but I will push this now, we can fix that one at a later time if needed.

Jakuje:
> Note that the previous test is also doing that odd thing, but I will merge this now, we can fix that one at a later time if needed.

Sure. Will open a follow-up PR.

PR #205, opened 2025-04-14
Release 1.1

simo5:
#### Description

Prepares the tree for Release 1.1.0

#### Checklist

- ~[ ] Test suite updated with negative tests~
- [x] Documentation was updated

#### Reviewer's checklist:

- [ ] ~Any issues marked for closing are fully addressed~
- [ ] ~There is a test suite reasonably covering new functionality or modifications~
- [ ] ~This feature/change has adequate documentation added~

### DIFF
--- a/packaging/kryoptic.spec
+++ b/packaging/kryoptic.spec
@@ -20,7 +20,7 @@
 
 
 Name:           kryoptic
-Version:        1.0.0
+Version:        1.1.0
 Release:        %autorelease
 Summary:        PKCS #11 software token written in Rust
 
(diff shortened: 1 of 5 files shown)


packaging/kryoptic.spec
```diff
@@ -20,7 +20,7 @@
 
 
 Name:           kryoptic
-Version:        1.0.0
+Version:        1.1.0
```
Jakuje, line 23: We discussed if we should enable pqc features in the fedora build. The annoying thing is the features list is in 6 places in the spec below. Should we make the change also here as this spec file is used in the CI? Or will it explode with older OpenSSL version we have in CentOS Stream 10 now?

Jakuje (approved):
lgtm

packaging/kryoptic.spec
simo5: it may explode, we'll deal with this later.

PR #212, opened 2025-04-18
tests: Drop pkcs11-tool dependency of migrate test

Jakuje:
#### Description

The Fedora 42 splits opensc to libs and tools, but only libs are
multilib. This means the 64b pkcs11-tool can not work with the 32b
kryoptic in the tests.
    
This is attempt to get rid of the opensc dependency by using the
rust-cryptoki API in the migrate test.

#### Checklist

- [X] This is not a code change

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/migrate.yml
+++ b/.github/workflows/migrate.yml
@@ -32,7 +32,11 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          dnf -y install git cargo clang-devel openssl-devel sqlite-devel opensc
+          dnf -y install git cargo clang-devel openssl-devel sqlite-devel
+          if [ "${{ matrix.name }}" = "i686" ]; then
+            dnf -y install rust-std-static.i686 openssl-devel.i686 \
+              sqlite-devel.i686
+          fi
 
       - name: Checkout Repository
         uses: actions/checkout@v4
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -46,7 +46,7 @@ jobs:
             dnf -y install rustc rpm-build openssl-devel cargo rust-toolset sqlite-devel clang git
           else
             dnf -y install rustc rpm-build cargo-rpm-macros openssl-devel git \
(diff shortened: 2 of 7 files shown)


Jakuje:
Ok, for Fedora we will need a new version of rust-cryptoki in Fedora.

simo5:
If we introduce our test setup tools, shouldn't we use them for kryoptic own tests as well ?

Jakuje:
> If we introduce our test setup tools, shouldn't we use them for kryoptic own tests as well ?

I marked them as tests so I do not think they are production ready. I would like to avoid shipping specific tools to set up kryoptic as the pkcs11 API should really do in most of the cases (unless we need multilib tests).

simo5:
So what do we do for the RPM builds, should we just wait until rawhide is updated ?

Jakuje:
> So what do we do for the RPM builds, should we just wait until rawhide is updated ?

The fast option is to downgrade the version we need to version in Fedora. There is really no incompatibility in what we use now. Just did that.

The other is to wait. I started nudging there so I hope it wont take long. We will need to update the rust-cryptoki anyway to support the PKCS#11 3.0 and then 3.2 API anyway.

Jakuje:
Ok, so it tries to install the test binaries too, which is not good. I would rather have them somehow more coupled with the test itself, but I am not sure what would be the best way.

And installation of cryptoki 0.6.2, which should be available in fedora does not seem to work either. Will have to spend some more tweaking around this ...

https://src.fedoraproject.org/rpms/rust-cryptoki

Jakuje:
One last issue is with the rust-cryptoki in Fedora now. As it is outdated, it requires bindgen 0.68, while the rest takes 0.69 already which explodes during the build. Discussed here:

 * https://bugzilla.redhat.com/show_bug.cgi?id=2292912
 * https://bugzilla.redhat.com/show_bug.cgi?id=2292911
 * https://src.fedoraproject.org/rpms/rust-cryptoki-sys/pull-request/1

simo5 (approved):
LGTM
Thanks!

PR #219, opened 2025-04-24
Add better support for array of attributes

### LINKED ISSUE #217: Arrays and Templates support
We currently do not have an attribute type for template of CK_ATTRIBUTE which we need to properly handle these attributes:
- CKA_WRAP_TEMPLATE
- CKA_UNWRAP_TEMPLATE
- CKA_DERIVE_TEMPLATE

Additional CKA_ALLOWED_MECHANISMS is an array of CKM_MECHANISM_TYPEs.

Currently all of these attributes are of type "ByteArray".

This allows carrying them but unless we know how to transforms them then their storage is endianess and platform CK_ULONG size specific.

Additionally we need to wire these templates in the key generation code so that thay can be actually used in wrap/derive operations.

simo5:
#### Description

Some attributes are actually well defined arrays of other attributes.

The two array types are arrays of CK_ULONGs and arrays of CK_ATTRIBUTEs

The latter is used for "TEMPLATE" attributes.

This PR implements support for more correctly storing and parsing arrays of CK_ULONG attributes.

At the same time we prevent trying to store TEMPLATE attributes.
This is because it is not entirely clear how to deal with the memory pointers referenced by the CK_ATTRIBUTE structure, and currently storing these attributes simply results in a broken attribute referencing arbitrary memory.

Fixes #217

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -13,6 +13,7 @@ pub enum AttrType {
     NumType,
     StringType,
     BytesType,
+    UlongArrayType,
     DateType,
     DenyType,
     IgnoreType,
@@ -110,9 +111,9 @@ static ATTRMAP: [Attrmap<'_>; ATTRMAP_SIZE] = {
     amap[63] = attrmap_element!(CKA_EC_POINT; as BytesType);
     amap[64] = attrmap_element!(CKA_ALWAYS_AUTHENTICATE; as BoolType);
     amap[65] = attrmap_element!(CKA_WRAP_WITH_TRUSTED; as BoolType);
-    amap[66] = attrmap_element!(CKA_WRAP_TEMPLATE; as BytesType);
-    amap[67] = attrmap_element!(CKA_UNWRAP_TEMPLATE; as BytesType);
-    amap[68] = attrmap_element!(CKA_DERIVE_TEMPLATE; as BytesType);
+    amap[66] = attrmap_element!(CKA_WRAP_TEMPLATE; as DenyType);
+    amap[67] = attrmap_element!(CKA_UNWRAP_TEMPLATE; as DenyType);
+    amap[68] = attrmap_element!(CKA_DERIVE_TEMPLATE; as DenyType);
     amap[69] = attrmap_element!(CKA_OTP_FORMAT; as NumType);
     amap[70] = attrmap_element!(CKA_OTP_LENGTH; as NumType);
     amap[71] = attrmap_element!(CKA_OTP_TIME_INTERVAL; as NumType);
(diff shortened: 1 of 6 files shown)


simo5:
@Jakuje looks like there is a crate issue on rawhide now ?

Jakuje:
> @Jakuje looks like there is a crate issue on rawhide now ?

The bindgen was updated in rawhide to 0.71. Bumping the dependency in Cargo.toml should solve the issue, I think.

Jakuje:
See #220.

src/attribute.rs
```diff
@@ -286,6 +287,24 @@ impl Attribute {
         Ok(&self.value)
     }
 
+    pub fn to_ulong_array(&self) -> Result<Vec<CK_ULONG>> {
+        let ulen = std::mem::size_of::<CK_ULONG>();
```
Jakuje, line 291: should we use here the `sizeof!()` macro? Given that it is already imported in this file, I think it would make it more consistent with the remaining code.

src/attribute.rs
```diff
@@ -427,6 +454,20 @@ fn bytes_to_vec(val: Vec<u8>) -> Vec<u8> {
 }
 conversion_from_type! {make from_bytes; from_type_bytes; from_string_bytes; from Vec<u8>; as BytesType; via bytes_to_vec}
 
+/* convert array of ulongs into an array of bytes where ulong is converted to
+ * an array of bytes in native endian order */
+fn ulong_array_to_vec(val: Vec<CK_ULONG>) -> Vec<u8> {
+    let ulen = std::mem::size_of::<CK_ULONG>();
```
Jakuje, line 460: same here. I would prefer the macro `sizeof!()` use

src/storage/sqlite.rs
```diff
@@ -58,12 +58,101 @@ pub struct SqliteStorage {
...
+    }
+
+    fn add_ulong_array(
+        obj: &mut Object,
+        atype: CK_ATTRIBUTE_TYPE,
+        val: ValueRef,
+    ) -> Result<()> {
+        obj.set_attr(match val.as_blob_or_null().map_err(bad_storage)? {
+            Some(blob) => {
+                let ulen = std::mem::size_of::<u64>();
```
Jakuje, line 112: `sizeof!()`

src/attribute.rs
simo5: sizeof!() is for places that need a CK_ULONG, this is a usize.

src/storage/sqlite.rs
simo5: does it really make sense to have a macro just to shorten this a bit?

Would you be ok if I import std::mem::size_of and then just call `size_of::<u64>()` in all these places?

Jakuje: Probably not. I missed the conversion in there. Lets ignore these comments.

Jakuje (approved):

PR #225, opened 2025-04-29
Support building on different OpenSSL versions

### LINKED ISSUE #92: test dynamic build on multiple OS Versions
maybe one more idea that can be put in place later in follow-up PR or in long term is to run this on different OS/versions, not just Fedora as this is no longer staticaly compiled from "known-to-be-good" openssl version.

_Originally posted by @Jakuje in https://github.com/latchset/kryoptic/pull/90#discussion_r1749928384_

### LINKED ISSUE #218: v1.1.0 build fails because it cannot find some OpenSSL bindings
Hello,

When trying to build Kryoptic v1.1.0 release tarball with the default features, within a Debian Bookworm (12) Docker container with both `libssl3` and `libssl-dev` installed, I get the following 2 errors:

```
error[E0425]: cannot find value `OSSL_SIGNATURE_PARAM_CONTEXT_STRING` in this scope
   --> src/ossl/eddsa.rs:220:30
    |
220 |                 name_as_char(OSSL_SIGNATURE_PARAM_CONTEXT_STRING),
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `OSSL_SIGNATURE_PARAM_DIGEST_SIZE`
    |
   ::: /kryoptic/target/debug/build/kryoptic-67f0933e799feefa/out/ossl_bindings.rs:602:1
    |
602 | pub const OSSL_SIGNATURE_PARAM_DIGEST_SIZE: &[u8; 12] = b"digest-size\0";
    | ------------------------------------------------------------------------- similarly named constant `OSSL_SIGNATURE_PARAM_DIGEST_SIZE` defined here

error[E0425]: cannot find value `OSSL_SIGNATURE_PARAM_INSTANCE` in this scope
   --> src/ossl/eddsa.rs:244:22
    |
244 |         name_as_char(OSSL_SIGNATURE_PARAM_INSTANCE),
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `OSSL_SIGNATURE_PARAM_DIGEST`
    |
   ::: /kryoptic/target/debug/build/kryoptic-67f0933e799feefa/out/ossl_bindings.rs:597:1
    |
597 | pub const OSSL_SIGNATURE_PARAM_DIGEST: &[u8; 7] = b"digest\0";
    | -------------------------------------------------------------- similarly named constant `OSSL_SIGNATURE_PARAM_DIGEST` defined here
```

This doesn't happen when I build locally on my machine, however.

simo5:
#### Description

Check which version we are building against and error out if incompatible features are requested.

Add tests to build against specific upstream versions to ensure features are compatible.

Fixes #92 
Fixes #218

#### Checklist

- [ ] Documentation was updated

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/openssl_versions.yml
+++ b/.github/workflows/openssl_versions.yml
@@ -1,0 +1,175 @@
+---
+name: OpenSSL Versions
+
+on:
+  pull_request:
+    branches: ["main"]
+  push:
+    branches: ["main"]
+
+jobs:
+  build:
+    name: Versions
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [ossl30, ossl32, ossl35, debian, centos10]
+        include:
+          - name: ossl30
+            build: true
+            version: "3.0"
+            repotype: "dnf"
+            repocachepath: "/var/cache/libdnf5"
+            container: fedora:latest
+          - name: ossl32
+            build: true
+            version: "3.2"
+            repotype: "dnf"
+            repocachepath: "/var/cache/libdnf5"
+            container: fedora:latest
+          - name: ossl35
+            build: true
+            version: "3.5"
+            repotype: "dnf"
+            repocachepath: "/var/cache/libdnf5"
+            container: fedora:latest
+          - name: debian
+            build: false
+            version: "3.5"
+            repotype: "apt"
+            container: debian:testing
+          - name: centos10
+            build: false
+            version: "3.5"
+            repotype: "dnf"
+            repocachepath: "/var/cache/dnf"
+            container: quay.io/centos/centos:stream10
+    container: ${{ matrix.container }}
+    steps:
+      - name: Get Date for DNF cache entry
+        if: ${{ matrix.repotype == 'dnf' }}
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Restore DNF cache
+        if: ${{ matrix.repotype == 'dnf' }}
+        uses: actions/cache/restore@v4
+        id: cache-dnf
+        with:
+          path: ${{ matrix.repocachepath }}
+          key: ${{ matrix.container }}-dnf-${{ steps.get-date.outputs.date }}
+
+      - name: Install Dependencies
+        run: |
+          if [ "${{ matrix.repotype }}" = "dnf" ]; then
+            if [ "${{ matrix.name }}" = "centos" ]; then
+              dnf_opts="--enablerepo=crb"
+            fi
+            dnf -y install git cargo clang-devel
+            if [ "${{ matrix.build }}" = "false" ]; then
+              dnf -y install openssl-devel sqlite-devel
+            else
+              dnf -y install openssl-devel sqlite-devel \
+                'perl(FindBin)' 'perl(lib)' 'perl(File::Compare)' \
+                'perl(File::Copy)' 'perl(bigint)' 'perl(Time::HiRes)' \
+                'perl(IPC::Cmd)' 'perl(Pod::Html)' 'perl(Digest::SHA)' \
+                'perl(Module::Load::Conditional)' 'perl(File::Temp)' \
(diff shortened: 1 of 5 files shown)


simo5:
Note: I tried to add debian:stable as a target to have a dynamic openssl 3.0.x but Debian stable rust is just too old and I couldn't really figure out how to successfully rustup from sh.rustup.rs ... so I removed the debian target from the matrix, maybe some braver than me can figure it out.

.github/workflows/openssl_versions.yml
```diff
@@ -0,0 +1,153 @@
...
+            if [ "${{ matrix.build }}" = "false" ]; then
+              dnf -y install openssl-devel
+            fi
+          elif [ -f /etc/debian_version ]; then
+            apt-get -q update
+            apt-get -yq install git clang curl
+            if [ "${{ matrix.build }}" = "false" ]; then
+              apt-get -yq install libssl-dev
+            fi
+          fi
```
Jakuje, line 71: if the debian is out of the picture, I would remove these parts to simplify the script.  Same for the caching of rust dependencies, where the debian is also checked.

.github/workflows/openssl_versions.yml
```diff
@@ -0,0 +1,153 @@
...
+on:
+  pull_request:
+    branches: ["main"]
+  push:
+    branches: ["main"]
+
+jobs:
+  build:
+    name: Versions
+    runs-on: ubuntu-22.04
```
Jakuje, line 13: Given that we run everything in containers, should we change to `ubuntu-latest`? The 22.04 will eventually go EOL. This is issue also in the other workflow files. Not a big deal though.

simo5: last time I tried ubuntu:latest things broke :)

Jakuje: Lets keep it for separate PR then.

simo5:
I added Debian testing, it is openssl 3.5 so not hugely helpful but at least it gives us a heads up if something breaks on there.

.github/workflows/openssl_versions.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Restore DNF cache
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            /var/cache/libdnf5
+          key: ${{ runner.os }}-dnf-${{ steps.get-date.outputs.date }}
```
Jakuje, line 51: This should be skipped on debian. But for some reason, it does not restore the cache on fedora either.
Also we should have separate cache for fedora and centos10.

simo5: ok let me check

.github/workflows/openssl_versions.yml
```diff
@@ -0,0 +1,173 @@
...
+        shell: bash
+
+      - name: Restore DNF cache
+        if: ${{ matrix.repotype == 'dnf5' }}
+        uses: actions/cache/restore@v4
+        id: cache-dnf
+        with:
+          path: |
+            /var/cache/libdnf5
+          key: ${{ matrix.container }}-dnf-${{ steps.get-date.outputs.date }}
```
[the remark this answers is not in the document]
simo5: no I want to keep these jobs separate for now, the size of the cache is not a big deal

.github/workflows/openssl_versions.yml
```diff
@@ -0,0 +1,175 @@
...
+            container: fedora:latest
+          - name: debian
+            build: false
+            version: "3.5"
+            repotype: "apt"
+            container: debian:testing
+          - name: centos10
+            build: false
+            version: "3.5"
+            repotype: "dnf5"
```
Jakuje, line 45: the only nit is that the centos10 does not have dnf5 so I would prefer now changing the dnf5 in all the `repotype` to `dnf`  (or just removing it and using the `repocachepath`

Jakuje (approved):

PR #230, opened 2025-05-02
Fix ML-DSA Keygen

Jakuje:
#### Description

Fix ML-DSA Keygen and tests basic use case

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/mldsa.rs
+++ b/src/mldsa.rs
@@ -378,6 +378,12 @@ impl Mechanism for MlDsaMechanism {
         ))? {
             return Err(CKR_TEMPLATE_INCONSISTENT)?;
         }
+        if !privkey.check_or_set_attr(Attribute::from_ulong(
+            CKA_PARAMETER_SET,
+            param_set,
+        ))? {
+            return Err(CKR_TEMPLATE_INCONSISTENT)?;
+        }
 
         mldsa::generate_keypair(param_set, &mut pubkey, &mut privkey)?;
         default_key_attributes(&mut privkey, mech.mechanism)?;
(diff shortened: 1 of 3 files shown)


Jakuje:
For the record, the failure is in the `src/ossl/mldsa.rs` in `generate_keypair()` . We are getting the params, but they surprisingly have only `seed` and `priv` parts even though we ask for `EVP_PKEY_KEYPAIR`.

simo5:
We are probably missing to set some property on the ctx before generation, I will look into it

Jakuje:
> We are probably missing to set some property on the ctx before generation, I will look into it

did not find anything that would be needed to set for key generation to get usable keypair, except for `OSSL_PKEY_PARAM_PROPERTIES` which does not have any useful information inline:

https://docs.openssl.org/master/man7/EVP_PKEY-ML-DSA/

simo5:
Let's leave this here for posterity about some of the preposterous choices of upstream: https://github.com/openssl/openssl/blob/3513a830cc159c913b6c9bf1cbaf1577a1f47808/providers/implementations/keymgmt/ml_dsa_kmgmt.c#L373-L376

beldmit:
https://github.com/openssl/openssl/issues/27542

vdukhovni:
Anyone expecting a particular representation of key data from "export" is misguided, all that can be expected from the export form is that it round-trips into "import". It is not a public format, for a stable external format, use d2i_PUBKEY or d2i_PrivateKey (in master, i.e. 3.6 or later, also d2i_PKCS8PrivateKey() which avoids legacy forms even with RSA, ...).

PR #233, opened 2025-05-02
Fixes for ML-KEM and Encapsulate API

Jakuje:
#### Description

Few issues noticed while playing around the ML-KEM
 * Keygen did not copy parameters set from public key template to private key, similarly as in #230
 * The Encapsulate API did not support size query. This is underspecified in current version of specs so we opened oasis-tcs/pkcs11#15
 * Test coverage for both of the issues above

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/tests/mlkem.rs
+++ b/src/tests/mlkem.rs
@@ -236,11 +236,7 @@ fn test_mlkem_operations() {
         ],
         &[],
         &[(CKA_ENCAPSULATE, true)],
-        &[
-            (CKA_CLASS, CKO_PRIVATE_KEY),
-            (CKA_KEY_TYPE, CKK_ML_KEM),
-            (CKA_PARAMETER_SET, CKP_ML_KEM_512)
-        ],
+        &[(CKA_CLASS, CKO_PRIVATE_KEY), (CKA_KEY_TYPE, CKK_ML_KEM),],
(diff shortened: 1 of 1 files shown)


src/mlkem.rs
```diff
@@ -268,6 +273,21 @@ impl Mechanism for MlKemMechanism {
         &self.info
     }
 
+    /// Get expected length of the encapsulated ciphertext for given key
+    ///
+    /// Returns the size in bytes or Err if the key is not ML-KEM
+    fn encapsulate_ciphertext_len(&self, key: &Object) -> Result<usize> {
+        match key.get_attr_as_ulong(CKA_PARAMETER_SET) {
```
simo5, line 280: You should check that CKA_KEY_TYPE matches CKK_ML_KEM first.

ML-DSA/SLH-DSA also have CKA_PARAMETER_SET and the CKP values overlap, so this function may not error out on the wrong key type.

simo5 (changes requested):
LGTM, just a minor nit.

src/mlkem.rs
Jakuje: Good point! I thought this is already done in the C_Encapsulate, but it is only done in the `encapsulate`. Done.

PR #235, opened 2025-05-05
Fix Public Ec Montgomery point storage

### LINKED ISSUE #234: Montgomery EC keypair generation results in wrongly formatted `CKA_EC_POINT`
Hello!

When generating ECC keypairs using the `CKM_EC_MONTGOMERY_KEY_PAIR_GEN` mechanism, the `CKA_EC_POINT` value on the public key ends up being stored as a DER-encoded ASN.1 `OctetString`. See:

https://github.com/latchset/kryoptic/blob/541dcbc1472e351fda333a0e06653b71129689f7/src/ossl/montgomery.rs#L75-L77

While this is consistent with the [expected format](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203418) for regular Short Weierstrass public keys, [Montgomery public keys](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203422) should be represented following RFC 7748 (I think this is the exact same as how it's formatted for Edwards curves). This also goes for the expected format of the public data used in ECDH on Montgomery curves.

For my usecase, this matters for key (de)serialization. Currently, we have to tweak our (de)serialization logic specifically for testing in a Kryoptic environment.

simo5:
#### Description
Fixes #234 

#### Checklist

- [ ] Documentation was updated

#### Reviewer's checklist:

### DIFF
--- a/src/ossl/montgomery.rs
+++ b/src/ossl/montgomery.rs
@@ -72,10 +72,10 @@ impl ECMontgomeryOperation {
         }
         let params = OsslParam::from_ptr(params)?;
         /* Public Key */
-        let point_encoded = asn1::write_single(
-            &params.get_octet_string(name_as_char(OSSL_PKEY_PARAM_PUB_KEY))?,
-        )?;
-        pubkey.set_attr(Attribute::from_bytes(CKA_EC_POINT, point_encoded))?;
+        let point = params
+            .get_octet_string(name_as_char(OSSL_PKEY_PARAM_PUB_KEY))?
+            .to_vec();
+        pubkey.set_attr(Attribute::from_bytes(CKA_EC_POINT, point))?;
 
         /* Private Key */
         let value = params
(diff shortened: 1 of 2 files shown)


Jakuje (approved):
thank you! Looks good!

Jakuje:
I think this will still need some more love for the `KRYOPTIC_EC_POINT_ENCODING` tweak we have. I think that one should apply only for the ECDSA and not for EC_EDWARDS nor EC_MONTGOMERY keys, but as far as I can see the code, we do it for this attribute type regardless the underlying key type.

simo5:
I did look at the EC POINT encoding and I think it make sense to keep it applied generally.
Thanks for the review.

PR #238, opened 2025-05-07
Fix PrivateKeyInfo ASN.1 structure

### LINKED ISSUE #51: Test interoperability of asymmetric key wrapping and unwrapping
The asymmetric key (RSA, ECDSA) wrapping and unwrapping works on kryoptic, but it involves a lot of ASN.1 encoding and decoding so it would be useful to test interoperability of these operations on some other implementation to make sure we do the right thing.

### LINKED ISSUE #236: Wrap/unwrap does not work with NSS Softokn and pkcs11-tool
After testing the wrapping and unwrapping of RSA private key using `CKM_AES_CBC_PAD` mechanism with `pkcs11-tool`, Kryoptic cannot unwrap key wrapped by Softokn (fails with `CKR_WRAPPED_KEY_INVALID`) and Softokn fails to unwrap key wrapped with Kryoptic  (fails with `CKR_TEMPLATE_INCOMPLETE`). However, wrapping and unwrapping private key only on Kryoptic alone works without problems.

I haven't found the issue related to the `CKR_TEMPLATE_INCOMPLETE` error in `pkcs11-tool`. Still, the errors suggest that there potentially might be some problems also on Kryoptic side, maybe related to the ASN.1 encoding of the key object.

Reproducer script: [kryoptic-reproducer.txt](https://github.com/user-attachments/files/20055006/kryoptic-reproducer.txt)

simo5:
This structure needs to use Algorithm Identifier and not a bare oid.

#### Description

This structure does not encode/decode properly and break KeyWrap/KeyUnwrap interoperability.

Fixes #236 
Fixes #51

#### Checklist

- [x] Documentation was updated

#### Reviewer's checklist:


src/kasn1/pkcs.rs
```diff
@@ -90,3 +90,37 @@ pub const HMAC_SHA_256_ALG: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
     oid: asn1::DefinedByMarker::marker(),
     params: AlgorithmParameters::HmacWithSha256(Some(())),
 };
+
+/// RSA Algorithm Identifier, typically used for Identifying a private or
+/// public key (for example in PrivateKeyInfo
```
Jakuje, line 95:
```suggestion
/// public key (for example in PrivateKeyInfo)
```
same on the following cases

Jakuje (approved):
otherwise lgtm

src/kasn1/pkcs.rs
simo5: good catch

PR #246, opened 2025-05-14
Fix initialization of info in uninitialized token

### LINKED ISSUE #245: Uninitialized token uses unexpected values for label, manufacturerID and model
Here's what I see when printing information about an unintialized token (using 31ba28c):

```
$ release/sq-cryptoki token list
  - Slot:
    - Description: Kryoptic Slot
    - Manufacturer: Kryoptic Project
    - Hardware version: 0.0
    - Firmware version: 1.1
    - Hardware token: no
    - Removable token: no
    - Token present: yes
  - Token:
    - Label: "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
    - Manufacturer: "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
    - Model: "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
    - Serial number: "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"
    - Initialized: no
    - Clock: not present
    - Has a protected authentication path: no
```

It appears that [`Token::get_token_info`](https://github.com/latchset/kryoptic/blob/31ba28c/src/token.rs#L184) is returning "half initialized values."  In particular, the label consists of 16 NUL bytes.   93f54cc5e2a9f0f9d3ff92e439e8faae96c98986 changed the defaults for an initialized token with the commit message:

```
Fix defaults for token info to conform to spec

The spec says all token strings are "space" padded and never zero
terminated.
```

I think it would be good to improve the values returned for an uninitialized token, too.

The spec (version 3.2, draft 13, line 1167) says:

```
	label	application-defined label, assigned during token initialization. MUST be padded with the blank character (‘ ‘). MUST NOT be null-terminated.
```

The spec doesn't seem to say what this should be when the token is uninitialized, but the empty string might be better then a bunch of NULs.

```
	manufacturerID	ID of the device manufacturer. MUST be padded with the blank character (‘ ‘). MUST NOT be null-terminated.
	model	model of the device. MUST be padded with the blank character (‘ ‘). MUST NOT be null-terminated.
```

The specification doesn't seem to say that these fields are initialized when the token is uninitialized, but since that information is available, it seems like they should be set accurately.

simo5:
#### Description

Fixes #245 

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,9 +110,9 @@ macro_rules! cast_or_ret {
     }};
 }
 
-/// Thread-local instance of the Cryptographically Secure Pseudo-Random Number
-/// Generator (CSPRNG). This is used to avoid contention and locking between
-/// different threads.
+/* Thread-local instance of the Cryptographically Secure Pseudo-Random Number
+ * Generator (CSPRNG). This is used to avoid contention and locking between
+ * different threads. */
 thread_local!(static CSPRNG: RefCell<RNG> = RefCell::new(RNG::new("HMAC DRBG SHA256").unwrap()));
 
 /// Fill a buffer with random data
--- a/src/token.rs
+++ b/src/token.rs
@@ -182,7 +182,7 @@ impl Token {
 
     /// Set the token info for uninitialized tokens
     fn uninitialized_token_info(&mut self) {
-        self.fill_token_info(&storage::StorageTokenInfo::default());
+        self.fill_token_info(&StorageTokenInfo::default());
         self.info.flags &= !CKF_TOKEN_INITIALIZED
     }
 
--- a/src/token.rs
+++ b/src/token.rs
@@ -119,10 +119,10 @@ impl Token {
     pub fn new(dbtype: &str, dbargs: Option<String>) -> Result<Token> {
         let mut token: Token = Token {
             info: CK_TOKEN_INFO {
-                label: [0u8; 32],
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -110,9 +110,9 @@ macro_rules! cast_or_ret {
     }};
 }
 
-/// Thread-local instance of the Cryptographically Secure Pseudo-Random Number
-/// Generator (CSPRNG). This is used to avoid contention and locking between
-/// different threads.
+/* Thread-local instance of the Cryptographically Secure Pseudo-Random Number
+ * Generator (CSPRNG). This is used to avoid contention and locking between
+ * different threads. */
```
Jakuje, line 115: I would probably add a note here with explanation why this is not a doc string, otherwise next one going through the missing docs string in the source will try to convert it back ;)

simo5: I am not sure we'll go thorough that again, I think it is fine as is, it is easy to change even if we regress, it is only a cosmetic annoyance.
If we care really to no "regress" I would rather add a CI test that fails on these kind of warnings

simo5:
Thanks for the review!

PR #259, opened 2025-06-03
Add some basic infrastructure to perform tracing

### LINKED ISSUE #251: Add support for logging/tracing
Generally, it is very hard to debug the failures of tests, specific configurations when the only information we get from the PKCS#11 layer is generic message CKR_GENERAL_ERROR (more specific errors are a bit more helpful, but mostly map to the input/API usage and not internal problems).

In tests, this will report assertion when we return from the specific function, but it does not show where the error originated from.

Example can be the #230, where we had few issues:
 * 0c628aa8ef40d951e544137bdb89dcf4ffe087a9 fixes the issue where the openssl did not set public key params when we asked for the KEY_PAIR. This lead into the failure on line 919, returning Err() from OpenSSL, but on the PKCS#11 layer, we could see only general error
 * #226 failed all tests with traces going somewhere into the test initialization. It would be helpful to get some trace logs showing the configuration processing and slot initialization as we go to see that it was really using wrong configuration.

We also discussed that from the OpenSSL layer, it might be helpful to get the OpenSSL error stack, except for the file:line information where the error jumped out.

The important is also the performance penalty of this processing so in play is still a macro that would be built in only when the build is debug and it would not log anything in the production builds.

It should be also configurable (through env variable?) if we should log and where we should log (stderr, file ...)

simo5:
This allows some limited logging, currently limited to mostly openssl call failures.

Fixes #251 

#### Description

The intent is to provide a lightweight and minimally invasive way to add debugging information useful to speed up debugging of issues that are happening deep into the stack. 

#### Checklist

#### Reviewer's checklist:


Cargo.toml
```diff
@@ -43,23 +43,26 @@ bimap = "0.6.3"
...
 hex = "0.4.3"
 itertools = "0.14.0"
 libc = "0.2.151"
+log = { version = "0.4.27", default-features = false, features = ["std"], optional = true }
 num-bigint = "0.4.4"
 num-integer = "0.1.45"
 num-traits = "0.2.17"
 once_cell = "1.18.0"
-rusqlite = { version = "0.31.0", optional = true }
+rusqlite = { version = "0.36.0", optional = true }
```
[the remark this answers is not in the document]
simo5: yeah I saw the RPM failing, it is not strictly needed, so I will just drop the commit that changes thee deps for now

src/interface.rs
```diff
@@ -0,0 +1,157 @@
+// Copyright 20235Simo Sorce
```
Jakuje, line 1:
```suggestion
// Copyright 2025 Simo Sorce
```

src/log.rs
```diff
@@ -0,0 +1,51 @@
...
+#[cfg_attr(target_os = "linux", link_section = ".init_array")]
+pub static INITIALIZE: extern "C" fn() = kryoptic_log_init;
+
+/// Initializes a simple logger for tracing purposes based on the values of
+/// the environment variable KRYOPTIC_TRACE:
+/// - stderr -> logs to standard error
+/// - <other> -> any other value is interpreted as a file name to log into
+/// - NOT PRESENT -> No tracing is initialized
+/// If the logger initialization encounters an error, (for example the log
+/// file can't be opened) no tracing is available and no error is reported.
```
Jakuje, line 15: You should also document the `KRYOPTIC_TRACE_LEVEL` env variable.

Jakuje (approved):
just two nits, otherwise lgtm

src/interface.rs
simo5: heh ...

Jakuje:
For rpm build, just add the missing `crate(vsprintf/default)` to the list of installed packages in here:

https://github.com/latchset/kryoptic/blob/main/.github/workflows/rpm.yml#L48

src/log.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use simplelog::{Config, LevelFilter, SimpleLogger, WriteLogger};
+
+#[used]
+#[cfg_attr(target_os = "linux", link_section = ".init_array")]
+pub static INITIALIZE: extern "C" fn() = kryoptic_log_init;
+
+/// Initializes a simple logger for tracing purposes based on the values of
+/// the environment variable KRYOPTIC_TRACE:
+/// - stdout -> logs to standard output
```
Jakuje, line 11: one more thing, should we support `stderr`? I would find it more common than stdout, which should have the application output, rather than the tracing information.

Jakuje (approved):

src/log.rs
simo5: I can;t support stderr directly, but as you can see in the example above you can easily use KRYOPTIC_TRACE=/dev/stderr

I know it is not exactly teh same thing (as it may not work correctly with redirection), but simplelog emits only to stdout (which is why I corrected the text which previously claimed stderr).

PR #262, opened 2025-06-05
Drop jsondb

### LINKED ISSUE #261: Drop the jsondb backend
The jsondb backend has been quite useful during initial development, but is really not usable in production due to the way it handles storage.

For a while we already stopped using it even in tests to source some static test data, where we now access directly the underlying json functions to read those files.

The goal of this issue is to complete the transition of the json bits to the testing infrastructure only and drop the jsondb storage driver.

simo5:
#### Description

Fixes #261 

#### Checklist

#### Reviewer's checklist:


src/lib.rs
```diff
@@ -4388,27 +4397,35 @@ static FNLIST_320: CK_FUNCTION_LIST_3_2 = CK_FUNCTION_LIST_3_2 {
...
 };
 
+/// Holds pointers to v3.0 interfaces
 static INTERFACE_300: CK_INTERFACE = CK_INTERFACE {
     pInterfaceName: INTERFACE_NAME_STD_NUL.as_ptr() as *mut u8,
     pFunctionList: &FNLIST_300 as *const _ as *const ::std::os::raw::c_void,
     flags: 0,
 };
 
+/// Holds pointers to v3.2 interfaces
```
Jakuje, line 4417: nit: I think each of these is a pointer to a single interface, not multiple interfaces:

```suggestion
/// Holds pointers to v2.40 interface
static INTERFACE_240: CK_INTERFACE = CK_INTERFACE {
    pInterfaceName: INTERFACE_NAME_STD_NUL.as_ptr() as *mut u8,
    pFunctionList: &FNLIST_240 as *const _ as *const ::std::os::raw::c_void,
    flags: 0,
};

/// Holds pointers to v3.0 interface
static INTERFACE_300: CK_INTERFACE = CK_INTERFACE {
    pInterfaceName: INTERFACE_NAME_STD_NUL.as_ptr() as *mut u8,
    pFunctionList: &FNLIST_300 as *const _ as *const ::std::os::raw::c_void,
    flags: 0,
};

/// Holds pointers to v3.2 interface
```

src/storage/mod.rs
```diff
@@ -49,7 +49,7 @@ impl Default for StorageTokenInfo {
 /// Trait for discovering available storage database backend types.
 ///
 /// Each backend implementation provides a static instance of this trait
-/// (e.g., `json::DBINFO`, `sqlite::DBINFO`) which is registered in the
+/// (e.g., `sqlite::DBINFO`) which is registered in the
```
Jakuje, line 52: should we add the nssdb one here instead?

Jakuje (approved):

src/lib.rs
simo5: ack

src/storage/mod.rs
simo5: I think one example is enough

simo5:
Thanks for the quick review!

PR #270, opened 2025-06-20
Compile-time disabling SHA1 for OpenSSL backend

beldmit:
#### Description

We don't trust SHA1 as a cryptographic secure mechanism and want to be able to disable it.
Current version implies compile-time disablement, for OpenSSL only

Resolving #181 

#### Checklist

#### Reviewer's checklist:


.gitignore
```diff
@@ -13,3 +13,5 @@ cscope.*
 tags
 .generated_files
 *.tar.gz
+/src/ossl/bindings.rs
+/src/pkcs11/
```
simo5, line 17: same here

.gitignore
```diff
@@ -13,3 +13,5 @@ cscope.*
 tags
 .generated_files
 *.tar.gz
+/src/ossl/bindings.rs
```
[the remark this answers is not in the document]
beldmit: Thanks, dropped

.github/workflows/build.yml
```diff
@@ -108,6 +111,9 @@ jobs:
             FEATURES="${FEATURES},standard,pqc"
           elif [ "${{ matrix.name }}" = "log" ]; then
             FEATURES="${FEATURES},standard,log"
+          elif [ "${{ matrix.name }}" = "no_sha1" ]; then
+            #nssdb doesn't work with no_sha1
```
Jakuje, line 115: nit: Is the comment needed here? Would not it be better in the "exclude" part above, where the db is actually excluded from the matrix?

src/ossl/ecdsa.rs
```diff
@@ -193,6 +199,9 @@ impl EcdsaOperation {
             CKM_ECDSA_SHA3_384,
             CKM_ECDSA_SHA3_512,
         ] {
+            if cfg!(feature = "no_sha1") && ckm == &CKM_ECDSA_SHA1 {
+                continue;
+            }
```
Jakuje, line 204: You should also do this in the RSA, HMAC, ECDH, ... where the sha1 based mechanisms are registered. They contribute to the list of supported mechanisms that is returned to the user.

.github/workflows/build.yml
beldmit: Fixed, thanks!

src/ossl/ecdsa.rs
simo5: This comment brings up a blind spot in this PR, namely no negative tests to ensure it is working.

simo5:
@beldmit can you add a simple test that is executed only when no_sha1 is set that queries the token for all of the supported mechanisms, and if any of the SHA1 mechanisms is present it fails ?

This way we should catch regressions as support for new services is added.

beldmit:
Sure, I will process the missing checks and will try to add the negative test

simo5:
> I pushed the fix but need a guideline to write a test

Expand on src/tests/mechs.rs

Add a test that is active only when `no_sha1` is selected, and then check that the returned vector of CK_MECHANISM_TYPEs does not contain any of the CKM_*SHA1* mechanisms.

While there you can probably drop the last test which is remanant of when we had only 4 mechanisms when we started the project :)
And if you are brave, as a bonus, you could add tests that check for presence/absence of other mechanisms based on the various features enabled (including the no_sha1 one).

src/tests/kdfs.rs
```diff
@@ -790,6 +793,7 @@ fn test_hkdf() {
 }
 
 #[cfg(feature = "pbkdf2")]
+#[cfg(not(feature = "no_sha1"))]
```
simo5, line 796: This is eliminating a whole test when no_sha1 is selected, it'd be nice to add at least a vector with sha512 or so and disable only the sha1 ones.

simo5 (commented):
One nit, otherwise once we have the mechs test this looks good for merging

src/tests/kdfs.rs
beldmit: This is intentional:

https://github.com/latchset/kryoptic/blob/47bd87c101985df717d7e219798e58fb79fdb957/src/tests/kdfs.rs#L864 enforces SHA1

So to fix this issue, I will need some more test vectors with a different prf

src/ossl/ecdsa.rs
Jakuje: Could we use there something like this too?
```rust
        let list = [
            CKM_ECDSA,
            #[cfg(not(feature = "no_sha1"))]
            CKM_ECDSA_SHA1,
            CKM_ECDSA_SHA224,
            CKM_ECDSA_SHA256,
            CKM_ECDSA_SHA384,
            CKM_ECDSA_SHA512,
            CKM_ECDSA_SHA3_224,
            CKM_ECDSA_SHA3_256,
            CKM_ECDSA_SHA3_384,
            CKM_ECDSA_SHA3_512,
        ]
        for ckm in list {
```

beldmit: Thanks, noted

src/tests/kdfs.rs
beldmit: Would you mind to raise a separate issue for this? It is easy enough but it's a part from  different elephant

beldmit: I created https://github.com/latchset/kryoptic/issues/277

src/ossl/ecdh.rs
```diff
@@ -28,7 +28,13 @@ use pkcs11::*;
...
 fn kdf_type_to_digest_alg(mech: CK_EC_KDF_TYPE) -> Result<DigestAlg> {
     Ok(match mech {
-        CKD_SHA1_KDF | CKD_SHA1_KDF_SP800 => DigestAlg::Sha1,
+        CKD_SHA1_KDF => {
+            if cfg!(feature = "no_sha1") {
+                return Err(CKR_MECHANISM_PARAM_INVALID)?;
+            } else {
+                return Ok(DigestAlg::Sha1);
+            }
+        }
```
Jakuje, line 37: My point was to adjust the code similarly also in the other cases that have the same pattern, so here
```suggestion
        #[cfg(not(feature = "no_sha1"))]
        CKD_SHA1_KDF => {
            return Ok(DigestAlg::Sha1);
        }
```
and accordingly in the other files that have the same pattern.

src/ossl/rsa.rs
```diff
@@ -152,6 +158,12 @@ fn parse_sig_params(
         CKM_SHA3_512_RSA_PKCS_PSS => (SigAlg::RsaPssSha3_512, true),
         _ => return Err(CKR_MECHANISM_INVALID)?,
     };
+    if cfg!(feature = "no_sha1")
+        && (mech.mechanism == CKM_SHA1_RSA_PKCS
+            || mech.mechanism == CKM_SHA1_RSA_PKCS_PSS)
+    {
+        return Err(CKR_MECHANISM_INVALID)?;
+    }
```
Jakuje, line 166: i would prefer the simple cfg macros over the compound if-cfg conditions
```suggestion
    #[cfg(feature = "no_sha1")]
    if mech.mechanism == CKM_SHA1_RSA_PKCS
        || mech.mechanism == CKM_SHA1_RSA_PKCS_PSS
    {
        return Err(CKR_MECHANISM_INVALID)?;
    }
```
looking that its not shorter, but looks less complex to me.

Same for similar cases in the other files.

src/tests/hashes.rs
```diff
@@ -122,34 +122,35 @@ fn test_hashes_digest() {
...
-    ret = fn_digest(
-        session,
-        value.as_mut_ptr(),
-        value.len() as CK_ULONG,
-        digest.as_mut_ptr(),
-        &mut digest_len,
-    );
-    assert_eq!(ret, CKR_OK);
-    assert_eq!(hash, digest);
+    if !cfg!(feature = "no_sha1") {
```
Jakuje, line 125: We usually did something like this in the other tests too:
```suggestion
        #[cfg(not(feature = "no_sha1"))]
        {
```
To me its easier to notice with the line starting with the `#[cfg...` that there is some conditional compilation (also with the help of syntax highlighter), rather than when the `cfg` is part of the condition, which blends with other conditions flying around.

src/tests/mechs.rs
```diff
@@ -25,6 +25,22 @@ fn test_get_mechs() {
...
+    {
+        let mut sha1_found = 0;
+        for mech in &mechs {
+            sha1_found += match &*mech {
+                &CKM_SHA1_RSA_PKCS => 1,
+                &CKM_ECDSA_SHA1 => 1,
+                &CKM_SHA1_RSA_PKCS_PSS => 1,
+                &CKM_SHA_1_HMAC => 1,
+                &CKM_SHA_1_HMAC_GENERAL => 1,
+                &CKM_SHA_1 => 1,
```
Jakuje, line 38: `CKM_SHA_1_KEY_GEN` missing here from what I grepped in the code.

src/ossl/rsa.rs
simo5: I would also prefer to avoid the cfg macro wherever possible.

In this case it will always work fine because the constants used are always defined, but the different between the macro and the attribute is:
- with the macro all branches always compile and are elided at run time
- with the attribute instead what follow the attribute is not compiled at all and can contain code that won't compile when the attribute is not set

simo5: sorry not runtime, they are still elided in a compilation step, but they will fail if they contain reference to variables, functions or constants, etc... that are not available.

src/tests/mechs.rs
simo5: Note that we do not implement all of these, but you can check for them regardless, they should never be there if no_sha1 is selected anyway

src/ossl/ecdh.rs
```diff
@@ -22,7 +22,8 @@ use pkcs11::*;
 /// ossl DigestAlg
 fn kdf_type_to_digest_alg(mech: CK_EC_KDF_TYPE) -> Result<DigestAlg> {
     Ok(match mech {
-        CKD_SHA1_KDF | CKD_SHA1_KDF_SP800 => DigestAlg::Sha1,
+        #[cfg(not(feature = "no_sha1"))]
+        CKD_SHA1_KDF => DigestAlg::Sha1,
```
simo5, line 26: what happened to  CKD_SHA1_KDF_SP800 ?

Jakuje (commented):
last few nits, otherwise lgtm.

The last though is that the writing `not(feature = "no_sha1")` is a bit hard to read due to double negation. So I am wondering if we should not make it the `sha1` feature, which would be in default (or not)?

src/rsa.rs
```diff
@@ -494,6 +494,10 @@ impl RsaPKCSMechanism {
             CKM_SHA3_384_RSA_PKCS_PSS,
             CKM_SHA3_512_RSA_PKCS_PSS,
         ] {
+            #[cfg(feature = "no_sha1")]
```
simo5, line 497: move this in the array ?

beldmit: Done, let's try

Jakuje (approved):

simo5 (approved):
LGTM

PR #273, opened 2025-06-20
Support Sp800 ECDH KDF flavor

### LINKED ISSUE #271: Add support for CKD_{SHA256, ...}_KDF_SP800
I'm working on implementing ECDH in [`sequoia-cryptoki`](https://gitlab.com/sequoia-pgp/sequoia-cryptoki).  [Currently](https://gitlab.com/sequoia-pgp/sequoia-cryptoki/-/blob/2ece2cb8/src/decrypt.rs#L226), I use `CKM_ECDH1_DERIVE` with the `NULL` KDF, export the shared key, and do the AES unwrap operation manually:

```rust
                let mpi::PublicKey::ECDH {
                    // curve,
                    // q,
                    // hash,
                    // sym
                    ..
                } = self.openpgp_key.mpis() else ...

                let mpi::Ciphertext::ECDH {
                    // Ephermeral key.
                    e,
                    // Symmetrically encrypted session key.
                    // key,
                    ..
                } = ciphertext else ...

                ....

                let ephemeral_key = e.decode_point(&Curve::Cv25519)?.0;
                let ecdh1_derive_params = Ecdh1DeriveParams::new(
                    EcKdf::null(), ephemeral_key);
                let attributes = &[
                    Attribute::Class(ObjectClass::SECRET_KEY),
                    Attribute::KeyType(KeyType::GENERIC_SECRET),
                    // Don't persist the shared key on the token.
                    Attribute::Token(false),
                    // Make the shared key as extractable.
                    Attribute::Extractable(true),
                    Attribute::Sensitive(false),
                ];
                let shared_key_handle = session.derive_key(
                    &Mechanism::Ecdh1Derive(ecdh1_derive_params),
                    object,
                    attributes)?;
                t!("Derived shared key on token");
                // Extract the shared key from the token.
                let mut shared_key = None;
                for attribute in session.get_attributes(
                    shared_key_handle,
                    &[ AttributeType::Value ])?
                {
                    if let Attribute::Value(v) = attribute {
                        shared_key = Some(v);

                    } else {
                        return Err(anyhow::anyhow!(
                            "Token returned unrequested attribute: {:?}",
                            attribute));
                    }
                }
                let shared_key = if let Some(shared_key) = shared_key {
                    shared_key
                } else {
                    return Err(anyhow::anyhow!(
                        "Failed to export shared key form token"));
                };
                t!("Extracted shared key: {} bytes", shared_key.len());
                // Decrypt the session key.
                let session_key = decrypt_unwrap(
                    self.public(),
                    &shared_key.into(),
                    ciphertext,
                    plaintext_len)?;
```

This works, but @Jakuje [pointed out](https://gitlab.com/sequoia-pgp/sequoia-cryptoki/-/issues/1) that this is not portable, because some tokens won't export the shared key.  Further, AIUI, this implementation will require FIPS certification, because we implement AES unwrap ourselves.  As such, I'm trying to implement decryption for OpenPGP's ECDH using only the PKCS#11 interface. 

The OpenPGP ECDH algorithm is [described in section 11.5 of RFC 9580](https://www.rfc-editor.org/rfc/rfc9580.html#section-11.5):

> This section describes the One-Pass Diffie-Hellman method, which is a combination of the ECC Diffie-Hellman method that establishes a shared secret and the key derivation method that processes the shared secret into a derived key. Additionally, this section describes a key wrapping method that uses the derived key to protect a session key used to encrypt a message.
> The One-Pass Diffie-Hellman method C(1, 1, ECC CDH) [[SP800-56A](https://www.rfc-editor.org/rfc/rfc9580.html#SP800-56A)] MUST be implemented with the following restrictions: the ECC Cofactor Diffie-Hellman (CDH) primitive employed by this method is modified to always assume the cofactor is 1, the KDF specified in [Section 11.4](https://www.rfc-editor.org/rfc/rfc9580.html#key-derivation-function) is used, and the KDF parameters specified below are used.
> The KDF parameters are encoded as a concatenation of the following 5 variable-length and fixed-length fields, which are compatible with the definition of the OtherInfo bit string [[SP800-56A](https://www.rfc-editor.org/rfc/rfc9580.html#SP800-56A)]:
> ...
> The key wrapping method is described in [[RFC3394](https://www.rfc-editor.org/rfc/rfc9580.html#RFC3394)]. The KDF produces a symmetric key that is used as a KEK as specified in [[RFC3394](https://www.rfc-editor.org/rfc/rfc9580.html#RFC3394)]. Refer to [Section 11.5.1](https://www.rfc-editor.org/rfc/rfc9580.html#ecdh-parameters) for the details regarding the choice of the KEK algorithm, which SHOULD be one of the three AES algorithms. Key wrapping and unwrapping is performed with the default initial value of [[RFC3394](https://www.rfc-editor.org/rfc/rfc9580.html#RFC3394)].
> ...

The most relevant bit for this discussion is that the KDF uses the method described in SP800-56A.  I believe that this means that we need to pass `CKD_SHA256_KDF_SP800` to `CKM_ECDH1_DERIVE`, but Kryoptic [only seems to support the ANSI X9.63 variants, e.g., `CKD_SHA256_KDF`](https://github.com/latchset/kryoptic/blob/b7bba7dea60b271ce36e8f3c69143c6188b7ea2d/src/ossl/ecdh.rs#L40).  These two algorithms are incompatible.  The [SP800-56A variant (page 47)](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/nist.sp.800-56Ar2.pdf) hashes:

```
H(counter || Z || OtherInfo).
```

Whereas the [X9.63 variant (page 22)](https://theswissbay.ch/pdf/man_doc/Software/ANSI/Cryptography_-_Key_Agreement_and_Key_Transport_Using_Elliptic_Curve_Cryptography.pdf) hashes:

``
H(Z || counter | [SharedInfo])
```

That is, `Z` and `counter` are reversed.

I'd appreciate any comments on my approach and analysis.  If I'm correct, then I would like to request support for `CKD_*_KDF_SP800`

FWIW, the current [generic ECDH implementation in Sequoia starts here](https://gitlab.com/sequoia-pgp/sequoia/-/blob/4ba329b4/openpgp/src/crypto/ecdh.rs#L120).

### PR DESCRIPTION
**Pull request by simo5**
#### Description

Add support for handling ECDH with SP800 flavor KDF

Fixes #271

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/ossl/ecdh.rs
+++ b/src/ossl/ecdh.rs
@@ -10,7 +10,7 @@ use std::borrow::Cow;
 use crate::attribute::CkAttrs;
 use crate::error::Result;
 use crate::mechanism::{Derive, MechOperation, Mechanisms};
-use crate::misc::bytes_to_vec;
+use crate::misc::{bytes_to_vec, zeromem};
 use crate::object::{default_key_attributes, Object, ObjectFactories};
 use crate::ossl::common::{
     cstr, get_ossl_name_from_obj, osslctx, privkey_from_object, EC_NAME,
@@ -19,7 +19,7 @@ use crate::ossl::common::{
 use ossl::bindings::{
     EVP_PKEY_PUBLIC_KEY, OSSL_PKEY_PARAM_GROUP_NAME, OSSL_PKEY_PARAM_PUB_KEY,
 };
-use ossl::derive::EcdhDerive;
+use ossl::derive::{EcdhDerive, OneStepKdfDerive, X963KdfDerive};
 use ossl::digest::DigestAlg;
 use ossl::{EvpPkey, OsslParam};
 use pkcs11::*;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/ossl/ecdh.rs [hunk 1]
```diff
@@ -221,14 +224,40 @@ impl Derive for ECDHOperation {
...
+        secret.resize(outlen, 0);
+
+        if self.kdf == CKD_NULL {
+            if outlen < keylen {
+                return Err(CKR_TEMPLATE_INCONSISTENT)?;
+            }
+            /* We need to take the tail of the raw output */
+            secret.drain(..(outlen - keylen));
+        } else {
+            /* Handle KDfs in token as OpenSSL does not support all cases */
```
Jakuje, line 241:
```suggestion
            /* Handle KDFs in token as OpenSSL does not support all cases */
```

src/tests/ecdh.rs
```diff
@@ -469,7 +469,7 @@ fn test_ecc_derive_nist() {
         derive_template.len() as CK_ULONG,
         &mut s_handle,
     );
-    assert_eq!(ret, CKR_MECHANISM_PARAM_INVALID);
+    assert_eq!(ret, CKR_OK);
```
Jakuje, line 472: I would extend a checking of the result here a bit.

I think there are also outputs in the test vectors that could be checked, but on the first sight, I was not able to decipher what needs to be changed to run the NIST derivation step here
https://github.com/latchset/kryoptic/blob/main/src/tests/ecdh_vectors.rs

Jakuje (approved):
just a nit for commit message. Otherwise lgtm.
The ecdh test is just for facepalm. I changed it at 3 places, but not in the 4th one. No idea how it worked over the last 6 months with that good confidence we did not notice in CI.

ossl/src/derive.rs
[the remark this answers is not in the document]
simo5: ok,
as for CI, it wasn't really running as the _SP800 mechs were not supported, that's why it didn't fail, it didn't matter what key it found for any given run

PR #276, opened 2025-06-24
Make PKCS#11 3.2 the default interface

### LINKED ISSUE #268: Enable PKCS#11 3.2 interface by default
The PKCS#11 3.2 headers are final so we should pull the updated headers (should be no-op) and enable them by default, also removing the pkcs11_3_2 feature:

https://github.com/latchset/pkcs11-headers?tab=readme-ov-file#note-about-32-headers

This might not be as straightforward as currently the feature depends on the ossl350 feature, but it should not so the logic will have to be changed there a bit.

simo5:
#### Description

As the subject says

Fixes #268 

#### Checklist

#### Reviewer's checklist:


Cargo.toml
```diff
@@ -13,7 +13,7 @@ license = "GPL-3.0-or-later"
 [workspace.dependencies]
 kryoptic = { version = "1.2.0", path = ".", default-features = false }
 ossl = { version = "1.2.0", path = "ossl", default-features = false }
-pkcs11 = { version = "1.2.0", path = "pkcs11", default-features = false }
+pkcs11 = { version = "1.2.0", path = "pkcs11", default-features = false, features = ["pkcs11_3_2"] }
```
Jakuje, line 16: We probably do not need to add the `features = ["pkcs11_3_2"]` here when it was removed?

pkcs11/Cargo.toml
```diff
@@ -13,3 +13,4 @@ bindgen = "0.71"
 
 [features]
 pkcs11_3_2 = []
+default = ["pkcs11_3_2"]
```
Jakuje, line 16: I think we should remove it here as it really does not do anything anymore. Or is it something for "backward compatibility"?

tools/softhsm/Cargo.toml
```diff
@@ -25,9 +25,8 @@ test = false
 clap = { version = "4.5.26", default-features = false, features = ["cargo", "derive", "help", "std", "usage"] }
 cryptoki = "0.9.0"
 libc = "0.2.151"
-pkcs11 = { version = "1.2.0", path = "../../pkcs11", default-features = false }
+pkcs11 = { version = "1.2.0", path = "../../pkcs11", default-features = false, features = ["pkcs11_3_2"] }
```
Jakuje, line 28: Same here, do we need to pass the no-op features here?

Cargo.toml
simo5: it was not removed in the pkcs11_3_2 crate

src/tests/signatures.rs
```diff
@@ -419,18 +419,15 @@ fn test_hmac_signatures() {
...
-        let ret = sig_verifysig(
-            session,
-            key_handle,
-            &mut testcase.value,
-            &mut testcase.result,
-            &mut mechanism,
-        );
-        assert_eq!(ret, CKR_OK);
-    }
+    /* check SignatureVErify API too */
```
Jakuje, line 422: nit:
```suggestion
    /* check SignatureVerify API too */
```

Jakuje (approved):

Cargo.toml
Jakuje: I think its for no good use even in the pkcs11 crate now when we consider them final so.

PR #282, opened 2025-06-25
Fix is_approved() to properly check input keys

### LINKED ISSUE #279: FIPS checking imported keys for KDF is wrong
As demonstrated in #278, the checking in `is_approved` of the source key for KDF is wrong as it requires the `CKA_OBJECT_VALIDATION_FLAGS` attribute is present on the object. But in practice, this object is ephemeral and never stored so the check will always fail and consider the base key non-approved for FIPS:

https://github.com/latchset/kryoptic/blob/edc346584d7edd7df88f064d502d1fc6567bdcce/src/fips/indicators.rs#L1268-L1270

Correctly, we should call the `check_key` on the passed object, to check the parameters of the key object are approved.

simo5:
#### Description

Keys may not have the ephemeral validation object on. In that case explicilty heck the key and act accordingly.

Also move some code in helper functions for better readability.

Fixes #279

There are no specific tests because #278 will introduce enough tests to ensure this works correctly

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/fips/indicators.rs
+++ b/src/fips/indicators.rs
@@ -1222,6 +1222,28 @@ fn check_key(
     false
 }
 
+/// Returns true if the key has an ephemeral FIPS validation flag set
+fn has_fips_flag(key: &Object) -> bool {
+    if let Ok(f) = key.get_attr_as_ulong(CKA_OBJECT_VALIDATION_FLAGS) {
+        if (f & KRF_FIPS) == KRF_FIPS {
+            return true;
+        }
+    }
+    false
+}
+
+/// Adds FIPS validation flag to a key object
+fn add_fips_flag(key: &mut Object) {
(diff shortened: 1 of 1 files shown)


src/fips/indicators.rs
```diff
@@ -1222,6 +1222,30 @@ fn check_key(
...
+    false
+}
+
+/// Adds FIPS validation flag to a key object
+fn add_fips_flag(key: &mut Object) {
+    let flag = match key.get_attr_as_ulong(CKA_OBJECT_VALIDATION_FLAGS) {
+        Ok(f) => f,
+        Err(_) => 0,
+    } | KRF_FIPS;
+    /* Ignores failures, as this is epehermal stuff and the code
```
Jakuje, line 1241:
```suggestion
    /* Ignores failures, as this is ephemeral stuff and the code
```

PR #289, opened 2025-07-01
Add config option to filter mechanisms

simo5:
#### Description

This will be useful for pkcs11-provider tests where we need to try fallback situations with tokens that do not support all the mechanisms.

#### Checklist

#### Reviewer's checklist:


simo5:
I think we should set at least 1.80 given the LazyLock feature was added in that version, let me fix the commit and add it

Jakuje:
> actually setting a rust-version will end up be more pain than gain for now. As it means we have to actively update it afterwards, unless we find a real need to impose a specific version I think we should not set one.

thanks for checking! I did not know what would be pros and cons, but asked. Ok for me not to have one.

simo5:
I will merge even if the test centos10 rpm test failed as the failure is infrastructural and not due to issues with the PR

PR #292, opened 2025-07-03
Tweak test matrix to test more configuarations

simo5:
#### Description

Note that trace log has been intentionally dropped because it is not yet useful enough when parallel tests all run at the same time, as the file gets garbled. Will re-enable once it can handle logging to a file all of the test outputs in parallel

#### Checklist

- [ ] ~Test suite updated with functionality tests~
- [ ] ~Rustdoc string were added or updated~
- [ ] ~CHANGELOG and/or other documentation added or updated~

#### Reviewer's checklist:

### DIFF
--- a/tools/Cargo.toml
+++ b/tools/Cargo.toml
@@ -24,5 +24,5 @@ test = false
 [dependencies]
 clap = { version = "4.5.26", default-features = false, features = ["cargo", "derive", "help", "std", "usage"] }
 kryoptic.workspace = true
-cryptoki = "0.9.0"
+cryptoki = "0.10.0"
 libc = "0.2.151"


.github/workflows/build.yml
```diff
@@ -17,11 +17,25 @@ jobs:
...
-        exclude: # nssdb doesn't work with no_sha1
-          - name: no_sha1
+        exclude:
+          - name: no_sha1 # nssdb doesn't work with no_sha1
             db: nssdb
+          - name: minimal
+            linking: fips
+          - name: standard
+            linking: fips
+          - name: pqc # only works with openssl 3.5.0 so exclude dyanmic
```
Jakuje, line 31:
```suggestion
          - name: pqc # only works with openssl 3.5.0 so exclude dynamic
```

.github/workflows/build.yml
```diff
@@ -17,11 +17,25 @@ jobs:
...
+            linking: dynamic
+          - name: pqc # already part of normal fips tests
+            linking: fips
+          - name: no_sha1 # already tested in fips and synamic build
```
Jakuje, line 35:
```suggestion
          - name: no_sha1 # already tested in fips and dynamic build
```

.github/workflows/build.yml
```diff
@@ -90,45 +110,46 @@ jobs:
...
+            OPTS="${OPTS} --release"
           fi
 
           # create a bugus configuration file to make sure it is not used during tests
           mkdir -p ~/.config/kryoptic/
           cp testdata/test.conf ~/.config/kryoptic/token.conf
 
           cargo build -vv $OPTS --features "$FEATURES"
-          KRYOPTIC_TRACE=trace.log cargo test -vv $OPTS --features "$FEATURES,log"
+          cargo test -vv $OPTS --features "$FEATURES"
```
Jakuje, line 143: this removes testing of the `log` feature. I think we should add it to the matrix.

.github/workflows/build.yml
```diff
@@ -17,11 +17,25 @@ jobs:
...
-          - name: no_sha1
+        exclude:
+          - name: no_sha1 # nssdb doesn't work with no_sha1
             db: nssdb
+          - name: minimal
+            linking: fips
+          - name: standard
+            linking: fips
+          - name: pqc # only works with openssl 3.5.0 so exclude dynamic
+            linking: dynamic
```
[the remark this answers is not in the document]
simo5: no, it is actually good that we can still test with 3.2 for at least a little while

Jakuje (approved):

Jakuje:
The RPM build is a mirror issue. The rust-cryptoki was updated to 0.10 in Fedora:
https://koji.fedoraproject.org/koji/taskinfo?taskID=134674014

But the update did not hit all and every mirror so we are sometimes getting the old version and sometimes the new one.

PR #293, opened 2025-07-04
Some `ossl` API work, mostly polishing

teythoon:
#### Description

I ported `sequoia-openpgp` from `rust-openssl` to (a slightly patched version of) `ossl`.  This is the first set of patches from that effort.  I picked them in the hope that they are uncontroversial, and we can get them out of the way.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/ossl/src/cipher.rs
+++ b/ossl/src/cipher.rs
@@ -8,7 +8,8 @@ use std::ffi::{c_int, c_void, CStr};
 
 use crate::bindings::*;
 use crate::{
-    cstr, trace_ossl, Error, ErrorKind, OsslContext, OsslParam, OsslSecret,
+    cstr, trace_ossl, Error, ErrorKind, OsslContext, OsslParam,
+    OsslParamBuilder, OsslSecret,
 };
 
 /// Wrapper around OpenSSL's `EVP_CIPHER`, managing its lifecycle.
@@ -266,7 +267,7 @@ impl OsslCipher {
             return Err(Error::new(ErrorKind::OsslError));
         }
 
-        let mut params: OsslParam;
+        let params: OsslParam;
         let mut params_ptr = std::ptr::null() as *const OSSL_PARAM;
 
         /* For some modes there is setup that needs to be done
@@ -276,8 +277,8 @@ impl OsslCipher {
                 ctx.aead_setup(alg, &aead)?
             }
             EncAlg::AesCts(_, mode) => {
-                params = OsslParam::with_capacity(1);
-                params.add_const_c_string(
+                let mut params_builder = OsslParamBuilder::with_capacity(1);
+                params_builder.add_const_c_string(
                     cstr!(OSSL_CIPHER_PARAM_CTS_MODE),
                     match mode {
                         AesCtsMode::CtsModeCS1 => {
@@ -291,7 +292,7 @@ impl OsslCipher {
                         }
                     },
                 )?;
-                params.finalize();
+                params = params_builder.finalize();
                 params_ptr = params.as_ptr();
             }
             _ => (),
--- a/ossl/src/digest.rs
+++ b/ossl/src/digest.rs
@@ -39,6 +39,17 @@ impl EvpMd {
     }
 }
 
+impl Clone for EvpMd {
+    fn clone(&self) -> Self {
+        let ret = unsafe { EVP_MD_up_ref(self.ptr) };
+
+        if ret != 1 {
+            panic!("EVP_MD_up_ref failed");
+        }
+        EvpMd { ptr: self.ptr }
+    }
+}
+
 impl Drop for EvpMd {
     fn drop(&mut self) {
         unsafe {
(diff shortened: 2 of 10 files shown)


ossl/src/digest.rs
```diff
@@ -37,6 +37,17 @@ impl EvpMd {
     pub unsafe fn as_mut_ptr(&mut self) -> *mut EVP_MD {
         self.ptr
     }
+
+    /// Tries to clone the `EvpMd`.
+    pub fn try_clone(&self) -> Result<EvpMd, Error> {
+        let success = unsafe { EVP_MD_up_ref(self.ptr) == 1 };
```
simo5, line 43: pleae use:
```
let ret = ..
if ret != 1 {
   return Err..
}
Ok(..)
```

We manage all the errors this way and it's easier to read code when it is consistent

ossl/src/digest.rs
```diff
@@ -37,6 +37,17 @@ impl EvpMd {
     pub unsafe fn as_mut_ptr(&mut self) -> *mut EVP_MD {
         self.ptr
     }
+
+    /// Tries to clone the `EvpMd`.
+    pub fn try_clone(&self) -> Result<EvpMd, Error> {
```
simo5, line 42: Uhmm while from the rust structure side this is somewhat of a clone, it is actually a reference, so I would call it try_reference()

ossl/src/digest.rs
```diff
@@ -76,6 +87,19 @@ impl EvpMdCtx {
...
         self.ptr
     }
+
+    /// Tries to clone the context.
+    pub fn try_clone(&self) -> Result<EvpMdCtx, Error> {
+        let mut new = Self::new()?;
+        let success =
+            unsafe { EVP_MD_CTX_copy_ex(new.as_mut_ptr(), self.as_ptr()) == 1 };
+
+        if success {
```
simo5, line 97: see above on the style to use to handle errors

ossl/src/cipher.rs
```diff
@@ -285,8 +288,8 @@ impl OsslCipher {
                 ctx.aead_setup(alg, &aead)?
             }
             EncAlg::AesCts(_, mode) => {
-                params = OsslParam::with_capacity(1);
-                params.add_const_c_string(
+                let mut builder = OsslParamBuilder::with_capacity(1);
```
simo5, line 291: this should be called params_builder

ossl/src/lib.rs
```diff
@@ -213,6 +213,15 @@ impl OsslContext {
         }
     }
 
+    /// Returns the default global context.
+    ///
+    /// See `OSSL_LIB_CTX_get0_global_default` for details.
+    pub fn global_default() -> OsslContext {
```
simo5, line 219: This was intentionally not exposed, why do you need to make it public ?

simo5 (changes requested):
Minor nits and a question, but overall this looks good.
I like the osslParamBuilder idea, I wonder though if we should rename all params to params_builder, given that often you end up reassigning the variable name on finalization.

ossl/src/digest.rs
teythoon: If you look at reference-counted smart pointers in the Rust standard library, they also call this operation clone: https://doc.rust-lang.org/std/rc/struct.Rc.html#method.clone

Therefore, I think Rust programmers would expect the function to be called `try_clone`.

ossl/src/lib.rs
simo5: I am planning to augment that code to also allow you to load the default openssl configuration in it (it currently doesn't because kryoptic case is special and it explicitly shouldn't use the default config either).

teythoon: That works fine, thanks for the pointer!

ossl/src/cipher.rs
teythoon: I confused myself.  When I first read your comment, I somehow assumed you wanted every instance of `OsslParamBuilder` to be called `params_builder`.  But now I saw that the change you commented on indeed stands out for renaming `params` to `builder`.  I think this was the first call site I converted, and I decided not to rename the variable anywhere else.

I see two options now:
- rename the variable here to `params`
- rename all instances of `OsslParamBuilder` to `params_builder`

What do you prefer?

ossl/src/digest.rs
teythoon: Yes, ideally, but [`Clone::clone`](https://doc.rust-lang.org/std/clone/trait.Clone.html#tymethod.clone) is infallible.

There is precedence for `try_clone` in the standard library, e.g.: https://doc.rust-lang.org/std/fs/struct.File.html#method.try_clone

simo5: we can make this call infallible too, the examples you provided also Panic!() in case of internal failure to clone()

simo5 (changes requested):
Please drop the last patch, it is simply incorrect.
BorrowedReferences is absolutely necessary.

Issue #300, opened 2025-07-23
Leaks secrets into the heap when exporting secret key material


teythoon:
I caught ossl leaking secrets when exporting keys, `OsslParam::from_ptr` doesn't mark the memory for cleansing.

```
#0  __memcpy_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:272
#1  0x00007fab28e77c6b in memcpy (__dest=<optimized out>, __src=<optimized out>, __len=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:29
#2  param_bld_convert (bld=0x55a159f75cc0, param=0x55a159f55a70, blk=0x55a159f55b28, secure=0x0) at ../crypto/param_build.c:344
#3  OSSL_PARAM_BLD_to_param (bld=bld@entry=0x55a159f75cc0) at ../crypto/param_build.c:383
#4  0x00007fab290528d3 in ecx_export (keydata=0x55a159f6d9f0, selection=<optimized out>, param_cb=0x7fab28e57ac0 <ossl_pkey_todata_cb>, cbarg=0x7ffdaecdf798) at ../providers/implementations/keymgmt/ecx_kmgmt.c:260
#5  0x000055a13dd0b31c in ossl::pkey::EvpPkey::todata (self=0x7ffdaecdfea0, selection=135) at src/pkey.rs:663
#6  0x000055a13dd0d29a in ossl::pkey::EvpPkey::export (self=0x7ffdaecdfea0) at src/pkey.rs:891
#7  0x000055a13dc49645 in sequoia_openpgp::crypto::backend::openssl::asymmetric::{impl#0}::ed25519_derive_public (secret=0x7ffdaece0420) at openpgp/src/crypto/backend/openssl/asymmetric.rs:208
#8  0x000055a13dadb5d0 in sequoia_openpgp::packet::key::v4::Key4<sequoia_openpgp::packet::key::SecretParts, sequoia_openpgp::packet::key::PrimaryRole>::import_secret_ed25519<sequoia_openpgp::packet::key::PrimaryRole, core::option::Option<std::time::SystemTime>> (private_key=&[u8](size=32) = {...}, ctime=...)
    at openpgp/src/packet/key/v4.rs:595
#9  0x000055a13daf734b in secret_leak_detector::test_ed25519 () at openpgp/tests/secret-leak-detector/detector.rs:161
[...]

PR #302, opened 2025-07-24
Zeroize key material exported from EvpPkey

simo5:
#### Description

The `EVP_PKEY_todata` function allocates an `OSSL_PARAM` array that can contain sensitive key material. Previously, this memory was freed without being securely cleared.

This commit introduces a new constructor,
`OsslParam::from_allocated_params_ptr`, which marks the underlying data for zeroization on drop. The `Drop` implementation is updated to cleanse the data buffers before they are freed.

`EvpPkey::todata` now uses this new constructor to ensure that sensitive key data is wiped from memory after use, preventing potential data leakage.

Fixes: #300 

#### Checklist

- [x] Test suite already covers the change

#### Reviewer's checklist:


ossl/src/lib.rs
```diff
@@ -704,6 +710,17 @@ impl<'a> OsslParam<'a> {
         }))
     }
 
+    /// Creates an `OsslParam` from an OpenSSl provider pointer that contains
+    /// allocated data that can be safely zeroized before freeing the pointer
+    /// Use with care as it will overwrite any data pointed by the params
```
Jakuje, line 715:
```suggestion
    /// allocated data that can be safely zeroized before freeing the pointer.
    /// Use with care as it will overwrite any data pointed by the params.
```

Jakuje (approved):
just typos. Otherwise lgtm

ossl/src/lib.rs
```diff
@@ -704,6 +710,17 @@ impl<'a> OsslParam<'a> {
...
+    pub fn from_allocated_params_ptr(
```
Jakuje, line 716: actually considering if the function name should be more self-explanatory like `from_sensitive_params_ptr()` or something along these lines. The "allocated" does not really say the casual reader that it should zeroize the buffers.

Jakuje: Looks good!

Jakuje (approved):

simo5:
Ok so this changes how data is exported completely to use EVP_PKEY_export, so we have direct access to the inner params created by the provider's export function.

However the zeroization of memory is a bit icky. In theory everything is always copied om the allocated area, but we do not really know how big that area is, so we try to avoid to zeroizing memory areas outsize of the projected allocated area.

ossl/src/pkey.rs
```diff
@@ -547,6 +547,57 @@ fn rsa_data_to_params(
...
+    let pslice = unsafe { std::slice::from_raw_parts(params, nelem) };
+
+    let mut params_builder = &mut *(arg as *mut OsslParamBuilder);
+    let ret = params_builder.copy_params(&pslice);
+
+    /* Zeroize any allocated data wich may hold copies of secrets.
+     * This is not the most clean way to do it, because this depends
+     * on knowing how OpenSSL internally builds parameter slices, and if
+     * that ever changes we may be overwrting memory that was not a
+     * temporary copy */
```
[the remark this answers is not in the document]
simo5: yes inndeed

ossl/src/lib.rs
```diff
@@ -653,6 +653,39 @@ impl<'a> OsslParamBuilder<'a> {
         Ok(())
     }
 
+    pub fn copy_params(&mut self, params: &[OSSL_PARAM]) -> Result<(), Error> {
```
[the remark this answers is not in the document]
simo5: ah thanks, yes

Jakuje (approved):
otherwise still looks good

simo5:
Ok given this seem to fix all issues for now, I am going to merge it, and we'll see later if openssl accepts the MR I proposed to make it more robust by using it when possible.

PR #315, opened 2025-08-18
Add missing OBJECT_VALIDATION attributes to objects

Jakuje:
#### Description

The object validation flags were somehow properly checked in the session for the validation state and missing flags on objects were ignored as they were computed on the fly. But these attributes could never make it to the user as even if they would be set during key generation, they were stripped while saving the attributes to the database (and never set for imported keys).

Fixes also bunch of related issues.

Built on top of #314 to have the build working.

#### Checklist

- [~] CHANGELOG and/or other documentation added or updated

#### Reviewer's checklist:


simo5 (approved):
This will cause us to pay the price of key validation on each retrieval ... we may want/need to add some form of caching later...
But otherwise I think this is fine for now.

Issue #244, opened 2025-05-13
Add support for SLH-DSA


Jakuje:
The third selected musketeer from the FIPS family:

 https://csrc.nist.gov/pubs/fips/205/final

PR #316, opened 2025-08-21
Add support for SLH-DSA

Jakuje:
#### Description

Fixes: #244.

#### Checklist

- [~] Test suite updated with negative tests
- [X] CHANGELOG and/or other documentation added or updated

#### Reviewer's checklist:


src/kasn1/pyca/oid.rs
```diff
@@ -182,14 +182,16 @@ pub const SHA384_OID: asn1::ObjectIdentifier =
     asn1::oid!(2, 16, 840, 1, 101, 3, 4, 2, 2);
 pub const SHA512_OID: asn1::ObjectIdentifier =
     asn1::oid!(2, 16, 840, 1, 101, 3, 4, 2, 3);
+// The SHA3 ones are wrong in pyca
+// https://github.com/pyca/cryptography/issues/13331
```
simo5, line 186: This changes the pyca header in ways that will probably make it diverge from upstream.

In this case I'd rather drop these ids, and add custom ones in src/kasn1/oid.rs

simo5 (changes requested):
Overall looks good, but I think we need to change the SHA3 OIDs fix

PR #317, opened 2025-08-26
ossl: Allow passing propq to the key creation API

Jakuje:
#### Description

This change is motivated by the need to be able to use the ML-DSA and ed448 signature algorithms in FIPS mode. Currently, the fips provider in RHEL 10.1 does not have these implemented so it fails during import of the certificate (which consists of verification of binding signatures).

This extends the ossl API to be able to provide the optional `propq` to the key creation API (as a `&Cstr` -- if you wish, we could make it some special type, but this looked like the least pain to start with.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/ossl/src/pkey.rs
+++ b/ossl/src/pkey.rs
@@ -23,13 +23,18 @@ pub struct EvpPkeyCtx {
 /// Methods for creating and accessing `EvpPkeyCtx`.
 impl EvpPkeyCtx {
     /// Fetches an algorithm by name and returns a wrapper `EvpPkeyCtx`
-    pub fn new(ctx: &OsslContext, name: &CStr) -> Result<EvpPkeyCtx, Error> {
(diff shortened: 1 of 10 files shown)


Jakuje:
Note, that this will need some adjustments if #316 will go first as it is using some of this API that I am changing.

Jakuje:
> I think the documentation of the constructor should mention what `propq` is, what it means, what it is used for, what possible values are, what one should do if one doesn't know what to pass in.
> 
> I guess `propq` is taken from the OpenSSL lingo, but it is just incomprehensible to me.

Good point! I will try to add it. From the PR I think its clear it maps to the parameter of the same name of the underlying functions. There is some writing about what properties in openssl are, but for general use, people should not need to mess with that:

https://docs.openssl.org/3.5/man7/property/

PR #329, opened 2025-09-15
Pull pyca SHA3 OIDs 

Jakuje:
#### Description

This pulls also the new version of asn1 crate. Unfortunately, its not backward compatible, but the mandatory API is safe to return None, which I reused here for simplicity.

#### Checklist

~- [ ] Test suite updated with functionality tests~
~- [ ] Rustdoc string were added or updated~
~- [ ] CHANGELOG and/or other documentation added or updated~

#### Reviewer's checklist:


src/kasn1/mod.rs
```diff
@@ -164,6 +167,9 @@ impl<'a> asn1::SimpleAsn1Writable for DerEncOctetString<'a> {
     fn write_data(&self, dest: &mut asn1::WriteBuf) -> asn1::WriteResult {
         dest.push_slice(self.as_bytes())
     }
+    fn data_length(&self) -> Option<usize> {
+        None
```
simo5, line 171: shouldn't this be `Some(self.data.len())` ?

Jakuje:
changing to draft as bringing the new asn1 into fedora would need a compat package, which is not worth the hassle now. But I just wanted to keep a track of this so lets keep it in draft until we get Fedora updated.

Jakuje:
Rebased on top of recent changes.

The python-cryptography landed so we can expect the rust-asn1 to be available any time soon:

https://bugzilla.redhat.com/show_bug.cgi?id=2395478

jeremycline:
> > @Jakuje do you think this will work now?
> 
> I think not (otherwise we would see failures in all the other pipelines). The rust-asn1 is still on 0.21:
> 
> https://src.fedoraproject.org/rpms/rust-asn1

I've got a pending update for asn1 and the cryptography stack: https://bodhi.fedoraproject.org/updates/FEDORA-2026-29d1a674fd.

I've turned off the autopush feature although I'm not sure if that'll actually do anything in Rawhide. If it does, it'd be great to include a kryoptic build in that side tag for 0.22, otherwise it'll start failing to build from source in Rawhide.

PR #330, opened 2025-09-15
Handle imports where only the Seed is provided

### LINKED ISSUE #325: Importing an ML-DSA key works, but using it fails
I'm able to generate an ML-DSA on the token and then use it for signing without a problem.  However, when I import an ML-DSA key by seed (which is what I have available), and try to sign a message, I get:

```
PKCS11 error: Some horrible, unrecoverable error has occurred.  In the worst case, it is possible that the function only partially succeeded, and that the computer and/or token is in an inconsistent state
```

Unfortunately, I don't have a simple reproducer (yet).

Tracing the code, it fails in [`mldsa_object_to_pkey`](https://github.com/latchset/kryoptic/blob/86d708b/src/ossl/mldsa.rs#L114) when trying to extract the `value`.  The error is:

```
Attribute not found, CKR_GENERAL_ERROR(5) - With message: 17
```

My feeling is that if the private key is not available, then it should be generated from the seed (either at import time or signing time), but this does not appear to be happening.

simo5:
#### Description

On import of ML-DSA or ML-KEM private keys:
- If only the seed is provided
  - generate the private key value.
- If both a seed and a private key value are provided
  - verify that the seed generates the same private key value

Fixes #325 

#### Checklist

#### Reviewer's checklist:


src/tests/mldsa.rs
```diff
@@ -1090,3 +1090,94 @@ fn test_mldsa_operations() {
...
+        CKA_VALUE,
+        value2.as_mut_ptr() as CK_VOID_PTR,
+        value2.len(),
+    )]);
+    let ret = fn_get_attribute_value(
+        session,
+        privkey2,
+        value2_template.as_mut_ptr(),
+        value2_template.len() as CK_ULONG,
+    );
```
Jakuje, line 1179: could we reuse the `extract_key_value()` or `check_attributes()` to avoid a lot of boilerplate? Same for the mlkem test.

Jakuje (approved):
just a nit for the test.

Jakuje (approved):

PR #333, opened 2025-09-16
Load default openssl config when possible

### LINKED ISSUE #319: Ossl bindings should probably load the default config
The ossl bindgins create a new "naked" context, but they should probably load in the default config by default.

May make sense to provide a way to conditionally or not load the config by allowing the request of a new context.

Finally we may also provide a setter that allows to change specific config options like the default query property

simo5:
#### Description

Changes kryoptic (in non-fips builds) to try to load the default openssl configuration on the custom context.
Because the configuration file could be missing for "legitimate" reasons, we do not treat as fatal the failure to load the default configuration.

Fixes #319 

#### Checklist

- [x] Test suite already has enough coverage for this change

#### Reviewer's checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ All notable changes to this project should be documented in this file.
   - [Handle imports where only the Seed is
      provided](https://github.com/latchset/kryoptic/pull/330)
 
+* Openssl context now loads the default openssl confgiuration file by default
+  - [Load default openssl config when
+     possible](https://github.com/latchset/kryoptic/pull/333)
+
 # [1.2.0]
 ## 2025-06-09
 
--- a/ossl/ossl.h
+++ b/ossl/ossl.h
@@ -12,6 +12,7 @@
 #include "openssl/obj_mac.h"
 #include "openssl/kdf.h"
 #include "openssl/err.h"
+#include "openssl/provider.h"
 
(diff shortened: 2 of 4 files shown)


Jakuje:
the pkcs11-provider integration test failure is a recursion when we try to load the pkcs11 provider into the kryoptic's openssl context?

Jakuje:
> the pkcs11-provider integration test failure is a recursion when we try to load the pkcs11 provider into the kryoptic's openssl context?

In that case, I think doing this in the kryoptic pkcs11 module is not that great idea :) But for any other users of ossl bindings, ye, this looks good to me.

simo5:
> > the pkcs11-provider integration test failure is a recursion when we try to load the pkcs11 provider into the kryoptic's openssl context?
> 
> In that case, I think doing this in the kryoptic pkcs11 module is not that great idea :) But for any other users of ossl bindings, ye, this looks good to me.

This is why I had created a new context an not loaded the default config in the first place ... sounds like we should revert this commit and rethink the strategy.

There are two reasons we wanted to load the default config:
1. crypto-policies
2. loading the FIPS provider when kryoptic is build with dynamic openssl bindings

Perhaps we should create a custom configuration file for kryoptic that loads just crypto-policies snippets but only the default provider. That would take care of both 1 and 2 w/o the risk of pulling in other providers that may break it?

PR #343, opened 2025-09-25
Fix Simple KDFs to work with token objects

Jakuje:
#### Description

Reported in parallaxsecond/rust-cryptoki#310 that the token objects do not behave as expected. After some investigation, I figured out I messed up the key creation there and attributes were off.

This attempts to fix it, but still the ALWAYS/NEVER attributes do not seem to be set to the right values and I am unable to figure out why. Any thoughts?

#### Checklist

~- [ ] Rustdoc string were added or updated~

#### Reviewer's checklist:

### DIFF
--- a/src/native/simplekdf.rs
+++ b/src/native/simplekdf.rs
@@ -5,11 +5,11 @@
 
 use std::fmt::Debug;
 
-use crate::attribute::CkAttrs;
+use crate::attribute::{Attribute, CkAttrs};
 use crate::error::Result;
 use crate::mechanism::*;
 use crate::misc::{bytes_to_vec, cast_params};
-use crate::object::{default_key_attributes, Object, ObjectFactories};
+use crate::object::{Object, ObjectFactories};
 use crate::pkcs11::*;
 
 #[cfg(feature = "fips")]
@@ -193,6 +193,9 @@ impl Derive for SimpleKDFOperation {
         let mut tmpl = CkAttrs::from(template);
         tmpl.add_missing_ulong(CKA_CLASS, &CKO_SECRET_KEY);
         tmpl.add_missing_ulong(CKA_KEY_TYPE, &CKK_GENERIC_SECRET);
+        let factory =
+            objfactories.get_obj_factory_from_key_template(tmpl.as_slice())?;
+        let mut dkey = factory.default_object_derive(tmpl.as_slice(), key)?;
 
         let other_data_len = match self.mech {
             CKM_CONCATENATE_BASE_AND_KEY => {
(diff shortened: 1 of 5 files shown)


src/native/simplekdf.rs
```diff
@@ -201,59 +204,36 @@ impl Derive for SimpleKDFOperation {
...
+                dkey.set_attr(Attribute::from_bool(
+                    CKA_ALWAYS_SENSITIVE,
+                    always_sensitive,
+                ))?;
+                let never_extractable = base_key_info.never_extractable
+                    && another_key_info.never_extractable;
+                dkey.set_attr(Attribute::from_bool(
+                    CKA_NEVER_EXTRACTABLE,
+                    never_extractable,
+                ))?;
```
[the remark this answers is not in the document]
Jakuje: Thanks! Thats likely the issue too.

src/tests/simplekdf.rs
```diff
@@ -70,8 +91,65 @@ fn test_concatenate_kdf() {
...
+        dk_handle,
+        &[(CKA_VALUE_LEN, 8)],
+        &[(CKA_VALUE, &exp_value)],
+        &[
+            (CKA_TOKEN, token),
+            (CKA_EXTRACTABLE, true),
+            (CKA_SENSITIVE, false),
+            (CKA_NEVER_EXTRACTABLE, false),
+            //(CKA_ALWAYS_SENSITIVE, false),
+            (CKA_ALWAYS_SENSITIVE, true), // FIXME should be false
```
simo5, line 105: what's on the original key?
Is it possible we are not setting CKA_ALWAYS_SENSITIVE at all on creation?
That could result on the default (true) being returned in internal queries (it would also be a bug probably).

src/tests/simplekdf.rs
```diff
@@ -420,3 +496,121 @@ fn test_concatenate_kdf_fips() {
...
+        session,
+        dk_handle,
+        &[(CKA_VALUE_LEN, 8)],
+        &[],
+        &[
+            (CKA_EXTRACTABLE, false),
+            (CKA_SENSITIVE, true),
+            (CKA_NEVER_EXTRACTABLE, false),
+            //(CKA_ALWAYS_SENSITIVE, false),
+            (CKA_ALWAYS_SENSITIVE, true), // FIXME should be false!
```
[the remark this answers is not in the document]
simo5: We should investigate if the original keys we create in the tests have the correct values for CKA_ALWAYS_SENSITIVE and CKA_NEVER_EXTRACTABLE, if the original keys have bad values, then you may get incorrect values here too because your code only checks the ancestor keys CKA_ALWAYS_SENSITIVE/CKA_NEVER_EXTRACTABLE values regardless of their CKA_SENSITIVE/CKA_EXTRACTABLE values.

src/tests/simplekdf.rs
```diff
@@ -54,7 +71,11 @@ fn test_concatenate_kdf() {
             (CKA_KEY_TYPE, CKK_GENERIC_SECRET),
         ],
         &[],
-        &[(CKA_EXTRACTABLE, true), (CKA_SENSITIVE, false)],
+        &[
+            (CKA_EXTRACTABLE, true),
+            (CKA_SENSITIVE, false),
+            (CKA_TOKEN, token),
```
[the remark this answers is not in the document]
Jakuje: To test also the storage use case. Initially, I tested only session objects, which worked also before, but some data is checked only when it goes to storage. I think it was the CKA_VALUE_LEN missing in my case, which is failing in https://github.com/parallaxsecond/rust-cryptoki/pull/310

src/tests/simplekdf.rs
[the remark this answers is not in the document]
Jakuje: Thats what I tried to do and it seems to be working, but some defaults might be jumping into the play. I will check.

src/tests/simplekdf.rs
Jakuje: I was trying to check the attributes in the new test below, `test_concatenate_kdf_attributes()` and the attributes on the base key looked sensible. But it might be that they were derived from the `CKA_SENSITIVE` on the fly. Let me double-check they are created correctly.

simo5: Just mentioning that we discuss this offline and we have a path forward that differs from the comment here. We will ensure importing an object also stores the the attributes properly so these attributes are always available on all key objects.

Jakuje: The current tests should be working with the current version setting the default attributes according to the specs.
What is not done is setting the default `CKA_ALWAYS_SENSITIVE` and `CKA_NEVER_EXTRACTABLE` for private keys and other secret keys (AES, HMACs ...), which requires the same semantics as we do not have now separate common create method now.
I can try to move them all into the `default_object_create()`, but then I need to distinguish the object class in there (should be in the template, I think) or create separate `private_key_object_create()`. I guess I will try the first approach as it makes the change in one place, which is easily traceable to the PKCS#11 specs

Jakuje: Done in the current iteration

simo5 (approved):
LGTM

PR #351, opened 2025-10-22
Use modern API for secure key parameter export

simo5:
Refactor key parameter export to use the modern `EVP_PKEY_todata` and `OSSL_PARAM_clear_free` functions when available (OpenSSL 4.0+).

This new API provides a more secure and robust method for handling sensitive key material, as `OSSL_PARAM_clear_free` ensures memory is zeroed before deallocation.

The build script now detects this function. For backward compatibility with older OpenSSL versions, the code falls back to the previous implementation using the deprecated `EVP_PKEY_export` and a custom callback to manually copy and zeroize the data.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/ossl/build.rs
+++ b/ossl/build.rs
@@ -58,6 +58,12 @@ impl bindgen::callbacks::ParseCallbacks for OsslCallbacks {
             println!("cargo::rustc-cfg=ossl_mlkem")
         }
     }
+
+    fn func_macro(&self, name: &str, _value: &[&[u8]]) {
+        if name == "OSSL_PARAM_clear_free" {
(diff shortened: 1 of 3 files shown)


simo5:
Note this code conditionally check for the function introduced in this PR to be available https://github.com/openssl/openssl/pull/28108
When it is we use the more robust method to deal with export and cleanup rather then the hack we perform in the callback for older versions of OpenSSL.

Once OpenSSL 4.0 is released we may eventually change the conditional to check for ossl_40, but we can't just yet.

Jakuje (approved):
lgtm

PR #355, opened 2025-11-03
Fix runtime panics in date functions in pkcs11/mod.rs

antoninnepras:
#### Description

Fixing runtime panics in date functions in pkcs11/mod.rs.

Functions string_to_ck_date and vec_to_date_validate had implementation errors causing rust panics. Out of range vectors elements were accesed. Vector was allocated but size was set to 0.

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/tests/kdfs.rs
+++ b/src/tests/kdfs.rs
@@ -1,8 +1,8 @@
 // Copyright 2024 Simo Sorce
 // See LICENSE.txt file for terms
 
+use crate::pkcs11::vendor::*;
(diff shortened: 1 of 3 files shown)


Jakuje:
> Basically any calls of the functions with correct vector length would panic. I can add some some test cases. Maybe to `src/tests/pkcs11.rs`?

Thank you. I would rather have some end-to-end test with the PKCS#11 API, probably in `src/tests/attrs.rs`, but I think this is also ok.

PR #365, opened 2025-11-11
Clear errors from signature probe

simo5:
#### Description

The ossl bindings checks for support of message_update in the specific signature provider by attempting an update operation. If this probe failed because the function was unsupported, it would leave an error on the OpenSSL error stack.

Because OpenSSL's error stack is tie to a thread local variable rather than the libctx this can propagate to the application kryoptic is loaded into.

This change wraps the underlying EVP calls with `ERR_set_mark` and `ERR_pop_to_mark`. This ensures that any errors generated by the probe are cleared, preventing them from propagating and causing spurious failures in subsequent code.

Found while working on https://github.com/latchset/pkcs11-provider/pull/657

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/ossl/src/signature.rs
+++ b/ossl/src/signature.rs
@@ -755,7 +755,9 @@ impl OsslSignature {
                      * supported.
                      */
                     let ret = unsafe {
-                        match ctx.op {
+                        /* avoid errors from propagating unnecessarily */
+                        ERR_set_mark();
+                        let r = match ctx.op {
                             SigOp::Sign => EVP_PKEY_sign_message_update(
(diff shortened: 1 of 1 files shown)


ossl/src/signature.rs
```diff
@@ -766,7 +768,11 @@ impl OsslSignature {
                                 std::ptr::null(),
                                 0,
                             ),
-                        }
+                        };
+                        /* we unconditionally popo to the error mark so no
```
Jakuje:
```suggestion
                        /* we unconditionally pop to the error mark so no
```

Jakuje (approved):

PR #377, opened 2025-11-24
Return error on subsequent C_Initialize calls

### LINKED ISSUE #375: Subsequent calls to C_Initialize do not return CKR_CRYPTOKI_ALREADY_INITIALIZED
Hello and thanks a lot for this useful library 👋 We use it a lot in `rust-cryptoki`!

Refactoring code, I noticed that the following test started to fail. We are calling `C_Initialize` twice in a row and check that the second one returns `CKR_CRYPTOKI_ALREADY_INITIALIZED`. It does not return an error when testing with `kryoptic`, as in panics with `"Initializing twice should not have been allowed"`.

```rust
fn is_initialized_test() -> TestResult {
    use cryptoki::context::CInitializeArgs;

    let pkcs11 = get_pkcs11();

    // initialize the library
    pkcs11.initialize(CInitializeArgs::OsThreads)?;

    match pkcs11.initialize(CInitializeArgs::OsThreads) {
        Err(Error::Pkcs11(RvError::CryptokiAlreadyInitialized, Function::Initialize)) => (), // expected
        Err(e) => panic!("Got unexpected error when initializing: {e}"),
        Ok(()) => panic!("Initializing twice should not have been allowed"),
    }

    pkcs11.finalize()?;

    Ok(())
}
```

simo5:
#### Description

The PKCS#11 specification requires that C_Initialize return CKR_CRYPTOKI_ALREADY_INITIALIZED if it is called more than once. This change implements this behavior by checking the initialization state at the start of the function.

This also resolves an issue where parallel test execution via `cargo test` would fail, as multiple tests would attempt to initialize the same global state. The test assertions have been updated to accept either CKR_OK or CKR_CRYPTOKI_ALREADY_INITIALIZED.

Fixes #375 

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/tests/init.rs
+++ b/src/tests/init.rs
@@ -16,7 +16,7 @@ fn test_init_token() {
     let mut args = TestToken::make_init_args(Some(testtokn.make_init_string()));
     let args_ptr = &mut args as *mut CK_C_INITIALIZE_ARGS;
     let mut ret = fn_initialize(args_ptr as *mut std::ffi::c_void);
-    assert_eq!(ret, CKR_OK);
+    assert_in!(ret, [CKR_OK, CKR_CRYPTOKI_ALREADY_INITIALIZED]);
     let mut session: CK_SESSION_HANDLE = CK_UNAVAILABLE_INFORMATION;
     let mut ro_session: CK_SESSION_HANDLE = CK_UNAVAILABLE_INFORMATION;
     let label = "Init Test";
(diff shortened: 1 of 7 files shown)


src/tests/init.rs
```diff
@@ -16,7 +16,7 @@ fn test_init_token() {
     let mut args = TestToken::make_init_args(Some(testtokn.make_init_string()));
     let args_ptr = &mut args as *mut CK_C_INITIALIZE_ARGS;
     let mut ret = fn_initialize(args_ptr as *mut std::ffi::c_void);
-    assert_eq!(ret, CKR_OK);
+    assert!(ret == CKR_OK || ret == CKR_CRYPTOKI_ALREADY_INITIALIZED);
```
Jakuje: wonder if something like this would help the reporting and would be more "rust-way":
```suggestion
    match ret {
        CK_OK | CKR_CRYPTOKI_ALREADY_INITIALIZED => (), // OK
        _ => panic!("C_Initialize returned unexpected value {ret}")
    }
```

simo5:
> > > Should we have an explicit test to verify the C_Initialize actually return the `CKR_CRYPTOKI_ALREADY_INITIALIZED` when called repetitively in one function?
> > 
> > 
> > It would have to be an external test, because we always get CKR_CRYPTOKI_ALREADY_INITIALIZED except for the first test, but we do not know which one is the first because they all run in parallel.
> 
> But if we call it twice in ONE test, it should do, shouldn't?

No. All tests run in the same binary, the very first test that is run (whichever it is) will get CKR_OK, *all* other tests will only ever get CKR_CRYPTOKI_ALREADY_INITIALIZED now.

Issue #288, opened 2025-06-30
Consolidate all fips testing in a single file


simo5:
This will make maintenance of fips tests easier.
And will also allow to methodically test each mechanism as required.

PR #388, opened 2025-12-15
Ed25519ctx is not fips approved

### LINKED ISSUE #380: The FIPS provider supports ED25519, but not ED25519CTX
I already changed the rust-cryptoki testsuite (which builds kryoptic against latest openssl 3.5) to match this expectation as the failure from the kryoptic is not very user friendly: parallaxsecond/rust-cryptoki#331. 

The related upstream issue is here and the patches landed down to 3.2: openssl/openssl#27502.

Given that we did not catch it here, I think this boils down to the missing test coverage for this case (but I did not investigate it).

The failure comes already with the `C_SignInit()` API as we try to fetch the ED25519CTX (https://github.com/parallaxsecond/rust-cryptoki/pull/324#issuecomment-3627761358):
```
Error: 0 [ERROR] Openssl Error(50856204): "FIPS internal library context, Algorithm (ED25519CTX : 0), Properties (<null>)"
Error: 0 [ERROR] ossl/src/signature.rs:37: EVP_SIGNATURE_fetch() failed: []
Error: 0 [ERROR] Generic CK error, CKR_DEVICE_ERROR(48)
test sign_verify_eddsa_with_ed25519_schemes ... FAILED
```
I think this can come up as a surprise as at the FIPS indicator level, we identifying the `CKM_EDDSA` mechanism as allowed, but it does not work for this particular parameters. Using some better error code would be the least thing we could do.

simo5:
#### Description

And OpenSSL completely removed it from the FIPs module. So fail early.

Fixes #380 
#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/ossl/eddsa.rs
+++ b/src/ossl/eddsa.rs
@@ -52,7 +52,11 @@ fn parse_params(
         if params.phFlag == CK_TRUE {
             return Ok((SigAlg::Ed25519ph, ctx));
         } else {
-            return Ok((SigAlg::Ed25519ctx, ctx));
+            if cfg!(feature = "fips") {
+                return Err(CKR_MECHANISM_PARAM_INVALID)?;
+            } else {
+                return Ok((SigAlg::Ed25519ctx, ctx));
+            }
         }
     }
     if outlen == OUTLEN_ED448 {
--- a/src/tests/eddsa.rs
+++ b/src/tests/eddsa.rs
@@ -58,6 +58,10 @@ fn test_create_eddsa_objects() {
     ));
 
     let ctx = hex::decode("666f6f").expect("Failed to decode context");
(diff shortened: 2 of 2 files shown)


Jakuje:
Could we have at least a simple test or should we put it under the #288?

simo5:
> Could we have at least a simple test or should we put it under the #288?

Well we have a test failure that indicates to me we already test for this, and actually need to catch it :)

simo5:
> > > Could we have at least a simple test or should we put it under the #288?
> > 
> > 
> > Well we have a test failure that indicates to me we already test for this, and actually need to catch it :)
> 
> oh. Good point. I see now. Its the issue that I do not run the kryoptic tests in the rust-cryptoki testsuite that I did not catch it in there.

The odd thing is that it doesn't fail locally for me, I am ... intrigued.

Issue #358, opened 2025-11-04
Handle ECDSA tests once ECDSA code in OpenSSL handles generateing public keys on import private only


simo5:
> current changes look good. Remaining item is still the ECDSA, I think

Well the upstream patch will take some discussion, I would rather just merge this and file a follow-up ticket so we do not forget to enable testing for the versions of openssl that will support it.

I think it is ok to let the code try and return an error on older versions, there isn't much else we can do except trying to return early, but it is not worth the added special code as it would need to be behind some features etc...

_Originally posted by @simo5 in https://github.com/latchset/kryoptic/issues/357#issuecomment-3486941182_

PR #402, opened 2026-01-19
Set CKA_PUBLIC_KEY_INFO for EC private keys

simo5:
#### Description

Enable the extraction of the public key from an EC private key to populate the CKA_PUBLIC_KEY_INFO attribute during object creation.

This functionality was previously disabled due to lack of support in older OpenSSL versions. This change adds a new `ossl::api_level()` function to check the OpenSSL version at compile time.

For OpenSSL versions before 4.0.0, where extraction is not supported, the `CKR_KEY_UNEXTRACTABLE` error is handled gracefully, allowing key creation to succeed without the attribute. For OpenSSL 4.0.0 and newer, the attribute is correctly populated. A corresponding test case has been re-enabled and is now conditional on the OpenSSL version.

Fixes: #358 

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/ec/ecdsa.rs
+++ b/src/ec/ecdsa.rs
@@ -44,13 +44,18 @@ static ECDSA_MECHS: LazyLock<[Box<dyn Mechanism>; 2]> = LazyLock::new(|| {
     ]
 });
 
-fn ecdsa_public_key_info(obj: &mut Object) -> Result<()> {
-    // Get raw public point from EC_POINT.
-    // For CKK_EC keys, CKA_EC_POINT is a DER-encoded OCTET STRING.
-    // SubjectPublicKeyInfo's bit string needs the raw point bytes.
-    let ec_point_der = obj.get_attr_as_bytes(CKA_EC_POINT)?;
-    let ec_point_raw = asn1::parse_single::<&[u8]>(ec_point_der)
-        .map_err(|_| CKR_ATTRIBUTE_VALUE_INVALID)?;
+fn ecdsa_public_key_info(obj: &mut Object, point: Option<&[u8]>) -> Result<()> {
+    let ec_point_raw = match point {
+        Some(p) => p,
+        None => {
+            // Get raw public point from EC_POINT.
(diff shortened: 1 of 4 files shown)


src/ec/ecdsa.rs
```diff
@@ -44,13 +44,18 @@ static ECDSA_MECHS: LazyLock<[Box<dyn Mechanism>; 2]> = LazyLock::new(|| {
...
 });
 
-fn ecdsa_public_key_info(obj: &mut Object) -> Result<()> {
-    // Get raw public point from EC_POINT.
-    // For CKK_EC keys, CKA_EC_POINT is a DER-encoded OCTET STRING.
-    // SubjectPublicKeyInfo's bit string needs the raw point bytes.
-    let ec_point_der = obj.get_attr_as_bytes(CKA_EC_POINT)?;
-    let ec_point_raw = asn1::parse_single::<&[u8]>(ec_point_der)
-        .map_err(|_| CKR_ATTRIBUTE_VALUE_INVALID)?;
+fn ecdsa_public_key_info(obj: &mut Object, point: Option<&[u8]>) -> Result<()> {
```
Jakuje: should we have a documentation string fro this function?

src/ossl/common.rs
```diff
@@ -209,7 +209,20 @@ pub fn extract_public_key(privkey: &Object) -> Result<Vec<u8>> {
 
     // 2. Export key data
     let pubkey_value = match pkey.export()? {
-        PkeyData::Ecc(mut e) => e.pubkey.take(),
+        PkeyData::Ecc(mut e) => {
+            if e.pubkey.is_some() {
+                e.pubkey.take()
+            } else {
+                if api_level().0 >= 4 {
```
Jakuje: I think this makes it quite hard to read what is going on here. Would it make sense to create some structure that would be returned by this function so it can be written at least in some way like this:
```suggestion
                if api_level().major >= 4 {
```

PR #403, opened 2026-01-19
Disable SLH-DSA in FIPS builds without ossl400

### LINKED ISSUE #318: SLH-DSA power-on tests
Follow-up from #316, which implemens the SLH-DSA. In some of the in-person meetings we discussed that we will need to ensure the power-on self tests for SLH-DSA need to be run. We do not want to run them unconditionally, as they take very long time so we want to make sure they run on first algorithm use.

I think this should be handled by the OpenSSL code, but it needs to be double-checked to confirm this.

simo5:
#### Description

The SLH-DSA module is now only compiled in FIPS builds when the `ossl400` feature is also enabled.

This is necessary because SLH-DSA requires deferred self-tests for FIPS compliance, and these tests are only available with the `ossl400` feature. Non-FIPS builds are unaffected.

Fixes #318 

#### Checklist

#### Reviewer's checklist:

### DIFF
--- a/src/enabled.rs
+++ b/src/enabled.rs
@@ -45,7 +45,12 @@ mod mlkem;
 #[cfg(feature = "mldsa")]
 mod mldsa;
 
-#[cfg(feature = "slhdsa")]
+// In fips builds enable slhdsa only if ossl400 was selected, which is required
+// for deferred self tests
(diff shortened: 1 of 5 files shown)


Jakuje (approved):
just typo

PR #410, opened 2026-01-28
Handle hmac state saving

simo5:
#### Description

Add HMAC state save/restore
HMAC is generally used in TLS for the transcript hash and needs to support saving and restoring state in order to finalize at each packet while also continuing to add to the hash in the following steps until the session is established.

Because OpenSSL does not expose serialization for HMAC, also add a KAT test and make FIPS builds also use the native HMAC code.

#### Checklist

#### Reviewer's checklist:


src/session.rs
```diff
@@ -748,15 +749,21 @@ impl Session {
             let mech = mechanisms.get(op.mech)?;
             let operation = mech.digest_restore(op.mech, op.data)?;
             <dyn Digest>::set_op(&mut self.operations, operation);
```
Jakuje: I think we should verify the key is not provided here and return `CKR_KEY_NOT_NEEDED` if it is not None.

simo5: done

PR #437, opened 2026-04-01
Add tests for trust attributes and values

simo5:
#### Description

Adds a comprehensive test case to systematically verify the creation, retrieval, and deletion of trust objects. This ensures the API correctly processes and stores all valid combinations of PKCS#11 trust attributes and values.

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/src/tests/objects.rs
+++ b/src/tests/objects.rs
@@ -383,6 +383,70 @@ fn test_create_trust_object() {
         CKR_ATTRIBUTE_VALUE_INVALID
     );
 
+    // Test case 7: Setting and verifying all trust attributes and values
+    let trust_attributes = [
+        CKA_TRUST_SERVER_AUTH,
+        CKA_TRUST_CLIENT_AUTH,
+        CKA_TRUST_CODE_SIGNING,
+        CKA_TRUST_EMAIL_PROTECTION,
+        CKA_TRUST_IPSEC_IKE,
+        CKA_TRUST_TIME_STAMPING,
+        CKA_TRUST_OCSP_SIGNING,
+    ];
+
+    let trust_values = [
+        CKT_TRUST_UNKNOWN,
+        CKT_TRUSTED,
+        CKT_TRUST_ANCHOR,
+        CKT_NOT_TRUSTED,
+        CKT_TRUST_MUST_VERIFY_TRUST,
+    ];
+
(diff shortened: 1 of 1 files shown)


src/tests/objects.rs
```diff
@@ -383,6 +383,58 @@ fn test_create_trust_object() {
...
+        CKA_TRUST_IPSEC_IKE,
+        CKA_TRUST_TIME_STAMPING,
+        CKA_TRUST_OCSP_SIGNING,
+    ];
+
+    let trust_values = [
+        CKT_TRUSTED,
+        CKT_TRUST_ANCHOR,
+        CKT_NOT_TRUSTED,
+        CKT_TRUST_UNKNOWN,
```
Jakuje: Any reason not to include `CKT_TRUST_MUST_VERIFY_TRUST` flag?

src/tests/objects.rs
```diff
@@ -383,6 +383,58 @@ fn test_create_trust_object() {
...
+                attr,
+                void_ptr!(&mut fetched_val),
+                std::mem::size_of::<CK_ULONG>(),
+            )]);
+            let ret = fn_get_attribute_value(
+                session,
+                trust_obj,
+                attr_template.as_mut_ptr(),
+                attr_template.len() as CK_ULONG,
+            );
```
Jakuje: this fetches just the one ATR we created on the object. Should we try to fetch some other trust attribute to verify default values are handled correctly? Is there some default value if not set during creation?

src/tests/objects.rs
simo5: No, just an AI miss, I will add it.

src/tests/objects.rs
simo5: All of them should be set to CKT_TRUST_UNKNOWN ... I am not sure there is value in testing that all others are in that state, but it doesn't hurt, maybe I just do an initial run to check all of them are set to CKT_TRUST_UNKNOWN, and then keep this current loop.

simo5: done

src/tests/objects.rs
simo5: done

Jakuje (approved):

PR #443, opened 2026-04-14
Add Kryoptic user guide

### LINKED ISSUE #441: Document kryoptic usage from user's point of view
Right now there is not much known how to configure kryoptic.

For users of the token we need to know environment variables, format of the configuration file, a sample usage with openssl and nss libraries...

simo5:
#### Description

Adds a new user guide (`doc/User_Guide.md`) to document the configuration and usage of the Kryoptic PKCS#11 soft token.

This documentation is added to provide users with a centralized reference for interacting with the library. It explains supported environment variables, TOML configuration structures (including global and slot-specific options), and provides practical examples for initializing tokens via `pkcs11-tool` and integrating with OpenSSL.

Fixes #441

#### Checklist

- [ ] ~Test suite updated~
- [ ] ~Rustdoc string were added or updated~

#### Reviewer's checklist:

### DIFF
--- a/doc/kryoptic.conf.man.md
+++ b/doc/kryoptic.conf.man.md
@@ -1,0 +1,103 @@
+% KRYOPTIC.CONF(5)
+% 
+% 
+
+# NAME
+
+kryoptic.conf - Configuration file for the Kryoptic PKCS#11 module
+
+# SYNOPSIS
+
+**/etc/kryoptic/token.conf**
+
+**${XDG_CONFIG_HOME}/kryoptic/token.conf**
+
+**${HOME}/.config/kryoptic/token.conf**
+
+# DESCRIPTION
+
+The **kryoptic.conf** file is a TOML-formatted configuration file used by the Kryoptic PKCS#11 module. It allows configuring global settings and provisioning multiple independent slots, each representing an idealized hardware slot where a cryptographic token can be utilized.
+
+Kryoptic searches for the configuration file in the following order:
+
+1. The path specified by the **KRYOPTIC_CONF** environment variable.
+2. **${XDG_CONFIG_HOME}/kryoptic/token.conf** (if the **XDG_CONFIG_HOME** environment variable is set).
+3. **${HOME}/.config/kryoptic/token.conf** (if the **HOME** environment variable is set).
+4. A system-wide directory, typically **/usr/local/etc/kryoptic/token.conf** or **/etc/kryoptic/token.conf** depending on build-time configuration.
+
+# GLOBAL OPTIONS
+
+**[ec_point_encoding]**
+:   Allows setting a global default encoding for `CKA_EC_POINT` attributes to maintain compatibility with applications that expect specific encodings (e.g., DER encoded EC Points).
+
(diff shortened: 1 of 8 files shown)


abbra:
Thank you, this looks great. LGTM

Jakuje:
Looks good! The only though I have if we should convert this to manual page so its directly accessible for Fedora users rather than having just upstream documentation or some documentation which is hard to discover.

abbra:
I have some code in IPA that turns markdown to man pages

simo5:
I think there is a pandoc filter for md -> grof, lemme see if I can easily add something here, the main question is how to integrate this with cargo doc ? Or should we do this only in the Fedora spec file and not through cargo ?

abbra:
You can do that with mdbook renderers: https://rust-lang.github.io/mdBook/format/configuration/renderers.html#custom-backend-commands. 

I have this in Akamu:
https://codeberg.org/abbra/akamu/src/branch/main/docs/book.toml
```
...
[output.rustdoc]
command = "python3 ../../scripts/mdbook-copy-api.py"

```
The renderer script copies rustdoc output into mdbook's output. It can do whatever you want

Jakuje:
Thinking a bit furhter, the manual pages should be more specifc. I think there should be a generic one for kryoptic (libkryptic_pkcs11?), separate for kryoptic.conf and eventually manuals for the tools (but it looks like we do not install these in Fedora?)/

simo5:
> Thinking a bit furhter, the manual pages should be more specifc. I think there should be a generic one for kryoptic (libkryptic_pkcs11?), separate for kryoptic.conf and eventually manuals for the tools (but it looks like we do not install these in Fedora?)/

Yeah we do not install the tools in fedora, they are mostly for testing, except the softhsm migration one.

I can create manpages for them I guess, but should we do it in this PR?

simo5:
@Jakuje  It took me little to generate three manpages, let me know if these go in the direction we need

doc/kryoptic.conf.man.md
```diff
@@ -0,0 +1,103 @@
...
+
+**/etc/kryoptic/token.conf**
+
+**${XDG_CONFIG_HOME}/kryoptic/token.conf**
+
+**${HOME}/.config/kryoptic/token.conf**
+
+# DESCRIPTION
+
+The **kryoptic.conf** file is a TOML-formatted configuration file used by the Kryoptic PKCS#11 module. It allows configuring global settings and provisioning multiple independent hardware slots, each representing an idealized hardware slot where a cryptographic token can be utilized.
```
Jakuje: `hardware slots`? Should this say "virtual" or "software"? I find refering to the HW in this context mildly confusing.

Jakuje (approved):

doc/kryoptic.conf.man.md
simo5: maybe I should just drop the "hardware" part

simo5: fixed

Jakuje (approved):

PR #448, opened 2026-05-05
Add ckm aes key wrap pad

abbra:
Implement CKM_AES_KEY_WRAP_PAD as defined in PKCS#11 v3.0 section 2.16. This mechanism applies PKCS#7 padding (block size 8) to input data before wrapping with the standard AES Key Wrap algorithm (RFC 3394 / NIST SP 800-38F Section 6.2). It uses the same OpenSSL cipher as CKM_AES_KEY_WRAP (AES-NNN-WRAP), with padding handled in the Kryoptic layer.

This is needed for FreeIPA migration from SoftHSMv2 to Kryoptic, where existing wrapped key material uses this mechanism.

Assisted-by: Claude Code, using Opus 4.6 model

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/src/fips/indicators.rs
+++ b/src/fips/indicators.rs
@@ -283,7 +283,7 @@ struct FipsMechanism {
 /// Struct that holds FIPS properties for keys and mechanisms
 struct FipsChecks {
     keys: [FipsKeyType; 17],
-    mechs: [FipsMechanism; 93],
+    mechs: [FipsMechanism; 94],
 }
 
 /// A constant instantiation of FIPS properties with a list
@@ -628,6 +628,18 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
         },
         FipsMechanism {
             mechanism: CKM_AES_KEY_WRAP,
+            operations: CKF_ENCRYPT | CKF_DECRYPT | CKF_WRAP | CKF_UNWRAP,
+            restrictions: [restrict!(KRY_UNSPEC), restrict!()],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
+                | CKF_WRAP
+                | CKF_UNWRAP
+                | CKF_DERIVE,
+        },
+        FipsMechanism {
+            mechanism: CKM_AES_KEY_WRAP_PAD,
             operations: CKF_ENCRYPT | CKF_DECRYPT | CKF_WRAP | CKF_UNWRAP,
             restrictions: [restrict!(KRY_UNSPEC), restrict!()],
             genflags: CKF_SIGN
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -182,6 +182,7 @@ impl AesOperation {
             CKM_AES_CTR,
             CKM_AES_CTS,
             CKM_AES_KEY_WRAP,
+            CKM_AES_KEY_WRAP_PAD,
             CKM_AES_KEY_WRAP_KWP,
         ] {
             mechs.add_mechanism(*ckm, &(*AES_MECHS)[0]);
@@ -375,7 +376,7 @@ impl AesOperation {
                     taglen: 0,
                 })
             }
-            CKM_AES_KEY_WRAP => {
+            CKM_AES_KEY_WRAP | CKM_AES_KEY_WRAP_PAD => {
                 let iv = match mech.ulParameterLen {
                     0 => AesIvData::none()?,
                     8 => AesIvData::simple(bytes_to_vec(
@@ -464,7 +465,7 @@ impl AesOperation {
             CKM_AES_CFB128 => EncAlg::AesCfb128(size),
             #[cfg(not(feature = "fips"))]
             CKM_AES_OFB => EncAlg::AesOfb(size),
-            CKM_AES_KEY_WRAP => EncAlg::AesWrap(size),
+            CKM_AES_KEY_WRAP | CKM_AES_KEY_WRAP_PAD => EncAlg::AesWrap(size),
             CKM_AES_KEY_WRAP_KWP => EncAlg::AesWrapPad(size),
             _ => return Err(CKR_MECHANISM_INVALID)?,
         })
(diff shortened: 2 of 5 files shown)


src/tests/aes.rs
```diff
@@ -814,6 +815,34 @@ fn test_aes_operations() {
...
+                handle,
+                data.as_slice(),
+                &mechanism,
+            ));
+            let dec = ret_or_panic!(decrypt(
+                session,
+                handle,
+                enc.as_slice(),
+                &mechanism,
+            ));
```
Jakuje, line 840: this tests the encrypt/decrypt opearations, but not the wrap/unwrap too?

src/fips/indicators.rs
```diff
@@ -638,6 +638,18 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
                 | CKF_UNWRAP
                 | CKF_DERIVE,
         },
+        FipsMechanism {
+            mechanism: CKM_AES_KEY_WRAP_PAD,
```
Jakuje, line 642: @simo5 is this mechanism OK from FIPS point of view?

simo5:
Ok so the mechanism sounded familiar to me and the reason it was not implement is:
```
6.16.3 AES Key Wrap
[...]
* CKM_AES_KEY_WRAP_PAD
The CKM_AES_KEY_WRAP_PAD mechanism is deprecated. CKM_AES_KEY_WRAP_KWP resp. CKM_AES_KEY_WRAP_PKCS7 shall be used instead.

**Comment by simo5** [comment]
@abbra any reason why you can't use CKM_AES_KEY_WRAP_PKCS7 instead ?

File: src/fips/indicators.rs (hunk 2 above)
Comment (simo5) [reply src/fips/indicators.rs:642]: No

**Comment by simo5** [comment]
Also PKCS#11 v3.0 section 2.16 does not exist ... and all the mechanisms are defined in section 6 anyway

**Comment by simo5** [comment]
> Also PKCS#11 v3.0 section 2.16 does not exist ... and all the mechanisms are defined in section 6 anyway

Nevermind we renamed those section in 3.1, it did exist in the old 3.0 ...

**Comment by simo5** [comment]
That said I would accept a PR for CKM_AES_KEY_WRAP_PKCS7, but I am not so hot on accepting a PR for a deprecated mechanism. I would accept it after CKM_AES_KEY_WRAP_PKCS7 is implemented and only if the data stored using CKM_AES_KEY_WRAP_PAD can't be unwrapped with CKM_AES_KEY_WRAP_PKCS7.

Note that CKM_AES_KEY_WRAP_PKCS7 should eb binary compatible with CKM_AES_KEY_WRAP_PAD so we shouldn't need to implement it, except perhaps as an alias...

File: src/ossl/aes.rs [hunk 3]
```diff
@@ -1453,6 +1472,19 @@ impl Encryption for AesOperation {
                 }
             }
             CKM_AES_KEY_WRAP | CKM_AES_KEY_WRAP_KWP => (),
+            CKM_AES_KEY_WRAP_PAD => {
+                let buf_len = self.buffer.len();
+                let pad_len = AES_KWP_BLOCK - (buf_len % AES_KWP_BLOCK);
```
simo5, line 1477: we should probably rename AES_KWP_BLOCK to something like AES_KEYWRAP_BLOCK or similar if we start using it for multiple mechanisms.

src/ossl/aes.rs
```diff
@@ -1453,6 +1472,19 @@ impl Encryption for AesOperation {
...
+                self.buffer.resize(buf_len + pad_len, pad_len as u8);
```
simo5, line 1478: `pad_len as u8` is unsafe, use u8::try_feom() and handle the error

src/ossl/aes.rs
```diff
@@ -1453,6 +1472,19 @@ impl Encryption for AesOperation {
...
+                outlen = ctx.update(self.buffer.as_slice(), cipher).or_else(
+                    |_| {
+                        self.finalized = true;
+                        Err(CKR_DEVICE_ERROR)
+                    },
+                )?;
+                zeromem(self.buffer.as_mut_slice());
```
simo5, line 1485: This should not be necessary, key wrap is single-shot only and we zeromem the buffer on deallocation

src/ossl/aes.rs
```diff
@@ -1798,6 +1839,10 @@ impl Decryption for AesOperation {
                     cipher_offset = cipher_end;
                 }
             }
+            CKM_AES_KEY_WRAP_PAD => {
+                self.buffer.extend_from_slice(cipher);
```
simo5, line 1843: if the buffer length is not zero before the extend this should fail because wrapping is a one shot operation.

src/ossl/aes.rs
```diff
@@ -1798,6 +1839,10 @@ impl Decryption for AesOperation {
...
+                cipher_offset = cipher_end;
```
simo5, line 1844: this shouldn't even happen, if padding is deferred to _final then probably this mechanism should return w/o going though openssl encryption here.

src/ossl/aes.rs
```diff
@@ -1899,6 +1944,44 @@ impl Decryption for AesOperation {
...
+                        Err(CKR_DEVICE_ERROR)
+                    },
+                )?;
+                zeromem(self.buffer.as_mut_slice());
+                self.buffer.clear();
+
+                if outlen < AES_KWP_BLOCK {
+                    return Err(CKR_ENCRYPTED_DATA_INVALID)?;
+                }
+                let pad_byte = plain[outlen - 1];
```
simo5, line 1963: The code here and below is not constant time, and will easily reveal the last byte resulting from decryption.

src/ossl/aes.rs
```diff
@@ -1344,6 +1356,21 @@ impl Encryption for AesOperation {
...
+                padded[plain.len()..].fill(pad_len as u8);
+                cipher_offset = ctx
+                    .update(padded.as_slice(), cipher)
+                    .or_else(|_| {
+                        self.finalized = true;
+                        Err(CKR_DEVICE_ERROR)
+                    })?;
+                zeromem(padded.as_mut_slice());
+                plain_offset = plain_end;
+            }
```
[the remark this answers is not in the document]
simo5: this mechanism is single-shot and we need to prevent multi-part operation, so this review is completely off

src/ossl/aes.rs
```diff
@@ -1505,6 +1537,9 @@ impl Encryption for AesOperation {
                         data_len + AES_KWP_BLOCK
                     }
                 }
+                CKM_AES_KEY_WRAP_PAD => {
+                    ((data_len / AES_KWP_BLOCK) + 2) * AES_KWP_BLOCK
```
simo5, line 1541: A coment that explains why +2 would be useful here

simo5 (changes requested):
The main issue here seems that decryption depadding is not constant time.

simo5:
I believe this is addressed by #501 so closing

abbra:
Yes, I wanted to test but had no time for that. I'll open new issue if needed

PR #449, opened 2026-05-05
Fix FIPS initialization and integrity check

FrantisekKrenzelok:
- Update misc/hmacify.sh to perform a normalization pass with objcopy before calculating the HMAC. This ensures the binary layout is finalized before the checksum is generated, preventing mismatches caused by objcopy's internal optimizations/re-alignments.

#### Description
There is a issue where the binary section alignment can change by invoking `objcopy` thus resulting in invalid hmac, this change first performs the `objcopy` on the original `.so` file and calculates the hmac on this copy which is already aligned.

#### Checklist

_I don't see these applicable_ 
- [ ] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/misc/hmacify.sh
+++ b/misc/hmacify.sh
@@ -6,7 +6,9 @@ if [[ $# -eq 0 ]]; then
 fi
 
 
-openssl dgst -binary -sha256 -mac HMAC -macopt hexkey:f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813 < $1 > $1.hmac
-objcopy --update-section .rodata1=$1.hmac $1 $1.mac
+objcopy $1 $1.norm
+openssl dgst -binary -sha256 -mac HMAC -macopt hexkey:f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813 < $1.norm > $1.hmac
+objcopy --update-section .rodata1=$1.hmac $1.norm $1.mac
 cp $1 $1.orig
 mv $1.mac $1
+rm $1.norm


misc/hmacify.sh
```diff
@@ -6,7 +6,9 @@ if [[ $# -eq 0 ]]; then
...
 
-openssl dgst -binary -sha256 -mac HMAC -macopt hexkey:f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813 < $1 > $1.hmac
-objcopy --update-section .rodata1=$1.hmac $1 $1.mac
+objcopy $1 $1.mac
+openssl dgst -binary -sha256 -mac HMAC -macopt hexkey:f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813 < $1.mac > $1.hmac
+objcopy --update-section .rodata1=$1.hmac $1.mac $1.mac2
 cp $1 $1.orig
-mv $1.mac $1
+mv $1.mac2 $1
+rm $1.mac
```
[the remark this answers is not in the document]
FrantisekKrenzelok: This is intentional

misc/hmacify.sh
```diff
@@ -6,7 +6,9 @@ if [[ $# -eq 0 ]]; then
 fi
 
 
-openssl dgst -binary -sha256 -mac HMAC -macopt hexkey:f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813 < $1 > $1.hmac
-objcopy --update-section .rodata1=$1.hmac $1 $1.mac
+objcopy $1 $1.mac
```
simo5: maybe instead of .mac, this should be called .norm (for normalized)

misc/hmacify.sh
```diff
@@ -6,7 +6,9 @@ if [[ $# -eq 0 ]]; then
...
+openssl dgst -binary -sha256 -mac HMAC -macopt hexkey:f4556650ac31d35461610bac4ed81b1a181b2d8a43ea2854cbae22ca74560813 < $1.mac > $1.hmac
+objcopy --update-section .rodata1=$1.hmac $1.mac $1.mac2
```
simo5: and here is would be $1.norm $1.mac removing "mac2" which is even more confusing.

src/fips/provider.rs
```diff
@@ -430,23 +430,45 @@ unsafe extern "C" fn fips_bio_read_ex(
...
     let mut fbio: Box<FipsBio> = unsafe { Box::from_raw(bio as *mut FipsBio) };
 
-    let ret = match fbio.op {
-        Bio::FileOp(ref mut op) => op.read(&mut readvec),
-        Bio::MemOp(ref mut op) => op.read(&mut readvec),
-    };
+    let mut total_read = 0;
+    let mut ret_status = 1;
+
+    while total_read < data_len {
```
simo5: This is nice and all but the underlying read operation should always return the requested bytes or raise an error yet ErrorKind::Interrupted is not handled here.

Can you drop this part from this PR and open a separate one?

I want to merge the PR and have a commit with jus the MAc stuff as I need to backport it elsehwere, so this read stuff in the same commit is a bit annoying

simo5 (changes requested):
PLease move bio_read fixes to a separate PR

src/fips/provider.rs
FrantisekKrenzelok: I verified, you are correct, will not be creating new PR.

simo5 (approved):
LGTM

PR #455, opened 2026-05-27
BugFix: AES - fix CCM and GCM iv counters for generators

FrantisekKrenzelok:
The internal aes iv generation counter wasn't being preserved in between message session, this change-set introduces fixes for this as well as a fix for aes iv calculation where when a byte-aligned(8x) `fixed bits` were used, the first byte of the counter region was still keeping the IV data instead of being rewritten by the counter buffer. Testing for for these 2 issues were also added.

_For more see. commit messages._

#### Checklist

- [x] Test suite updated
~~- [ ] Rustdoc string were added or updated~~
~~- [ ] CHANGELOG and/or other documentation added or updated~~
~~- [ ] This is not a code change~~

#### Reviewer's checklist:

~~- [ ] Any issues marked for closing are fully addressed~~
~~- [ ] This feature/change has adequate documentation added~~
~~- [ ] A changelog entry is added if the change is significant~~
~~- [ ] Doc string are properly updated~~no

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -488,7 +488,8 @@ impl AesOperation {
         }
 
         let mut genidx = params.iv.fixedbits / 8;
-        let mask = u8::try_from(genbits % 8)?;
+        let bits = genbits % 8;
+        let mask = if bits == 0 { 0xff } else { (1u8 << bits) - 1 };
         let genbytes = (genbits + 7) / 8;
 
         match params.iv.generator {
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -129,10 +129,8 @@ struct AesParams {
     taglen: usize,
 }
 
-#[cfg(feature = "fips")]
-impl AesParams {
-    fn zeroize(&mut self) {
-        zeromem(self.iv.buf.as_mut_slice());
+impl Drop for AesParams {
+    fn drop(&mut self) {
         zeromem(self.aad.as_mut_slice());
     }
 }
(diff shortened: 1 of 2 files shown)


src/ossl/aes.rs
```diff
@@ -980,16 +981,22 @@ impl AesOperation {
...
     ) -> Result<AesOperation> {
+        #[allow(unused_mut)]
+        let mut params = Self::init_params(&CK_MECHANISM {
+            mechanism: CK_UNAVAILABLE_INFORMATION,
+            pParameter: std::ptr::null_mut(),
+            ulParameterLen: 0,
+        })?;
+
+        #[cfg(feature = "fips")]
+        params.zeroize();
```
simo5: is this actually right, just after initialization ?

simo5: won't is wipe the AAD and IV that have just been intialized ?

src/tests/aes.rs
```diff
@@ -1389,3 +1389,462 @@ fn test_aes_macs() {
...
+        session,
+        CKM_AES_KEY_GEN,
+        std::ptr::null_mut(),
+        0,
+        &[(CKA_VALUE_LEN, 16),],
+        &[],
+        &[(CKA_ENCRYPT, true), (CKA_DECRYPT, true),],
+    ));
+
+    let (iv_orig, _, _, _, _) = get_gcm_test_data();
```
simo5: can you make use of aad as well here?

simo5 (approved):
LGTM

PR #461, opened 2026-06-16
Add support fot CKM_HOTP and OTP key factory

simo5:
#### Description

I was curious to implement HOTP because it was the simplest thing that would allow me to consider the problem of stateful keys (keys that need to update state on token after an operation)

This PR implements OTP Keys factory and provide a simple implementation of an HOTP token mechanism that uses Sign/Verify APIs to generate/verify an HOTP using a CKK_HOTP key object as the HOTP token state.

Fixes #418 

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/src/hotp.rs
+++ b/src/hotp.rs
@@ -1,0 +1,667 @@
+// Copyright 2023-2026 Simo Sorce
+// See LICENSE.txt file for terms
+
+use std::fmt::Debug;
+use std::sync::LazyLock;
+
+use crate::attribute::{Attribute, CkAttrs};
+use crate::error::Result;
+use crate::mechanism::{
+    Mac, MechOperation, Mechanism, Mechanisms, Sign, Verify,
+};
+use crate::native::hmac::HMACOperation;
+use crate::object::factory::{
+    attr_element, OAFlags, ObjectFactories, ObjectFactory, ObjectFactoryData,
+};
+use crate::object::key::{
+    default_key_attributes, default_secret_key_generate, KeyFactory,
+    SecretKeyFactory,
+};
+use crate::object::otp::OTPKeyFactory;
+use crate::object::{Object, ObjectType};
+use crate::pkcs11::*;
+
+/// Object that holds HOTP Mechanisms
+pub(crate) static HOTP_MECHS: LazyLock<[Box<dyn Mechanism>; 2]> =
+    LazyLock::new(|| {
+        [
+            Box::new(HOTPKeyMechanism::new()),
+            Box::new(HOTPMechanism::new()),
+        ]
+    });
+
+/// The HOTP Key Factory facility.
+static HOTP_KEY_FACTORY: LazyLock<Box<dyn ObjectFactory>> =
+    LazyLock::new(|| Box::new(HOTPKeyFactory::new()));
+
+/// Registers all implemented HOTP Mechanisms and Factories
+pub fn register(mechs: &mut Mechanisms, ot: &mut ObjectFactories) {
+    mechs.add_mechanism(CKM_HOTP_KEY_GEN, &(*HOTP_MECHS)[0]);
+    mechs.add_mechanism(CKM_HOTP, &(*HOTP_MECHS)[1]);
+
+    ot.add_factory(
+        ObjectType::new(CKO_OTP_KEY, CKK_HOTP),
+        &(*HOTP_KEY_FACTORY),
+    );
+}
+
+/// This is a specialized factory for objects of class CKO_OTP_KEY
+/// and CKA_KEY_TYPE of value CKK_HOTP
+///
+/// [HOTP secret key objects](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693654)
+#[derive(Debug)]
+pub struct HOTPKeyFactory {
+    data: ObjectFactoryData,
+}
+
+impl HOTPKeyFactory {
+    /// Initializes a new HOTPKeyFactory object
+    pub fn new() -> HOTPKeyFactory {
+        let mut factory = HOTPKeyFactory {
+            data: ObjectFactoryData::new(CKO_OTP_KEY),
+        };
+
+        factory.add_common_otp_key_attrs();
+
+        let attributes = factory.data.get_attributes_mut();
+
+        attributes.push(attr_element!(
+            CKA_VALUE; OAFlags::Sensitive | OAFlags::RequiredOnCreate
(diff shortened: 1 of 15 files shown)


src/object/otp.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Adds the OTP key attributes defined in the spec
+    fn add_common_otp_key_attrs(&mut self) {
+        self.add_common_secret_key_attrs();
+        let attrs = self.get_data_mut().get_attributes_mut();
+        attrs.push(attr_element!(
+            CKA_OTP_FORMAT; OAFlags::Defval | OAFlags::SettableOnlyOnCreate; Attribute::from_ulong; val CK_OTP_FORMAT_DECIMAL));
+        attrs.push(attr_element!(
+            CKA_OTP_LENGTH; OAFlags::Defval | OAFlags::SettableOnlyOnCreate; Attribute::from_ulong; val 6));
+        attrs.push(attr_element!(
+            CKA_OTP_TIME_INTERVAL; OAFlags::SettableOnlyOnCreate; Attribute::from_ulong; val CK_UNAVAILABLE_INFORMATION));
```
Jakuje: The attribute `CKA_OTP_TIME_INTERVAL` is defined only for `RSA SecurID` tokens:
https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693648

Do you want to specifically support this type of objects? They would need separate `CKK_SECURID`, but you define only the `CKK_HOTP`.

src/hotp.rs
```diff
@@ -0,0 +1,699 @@
...
+        if mech.ulParameterLen != 0 && !mech.pParameter.is_null() {
+            if mech.ulParameterLen as usize
+                != std::mem::size_of::<CK_OTP_PARAMS>()
+            {
+                return Err(CKR_MECHANISM_PARAM_INVALID)?;
+            }
+            let params = unsafe { &*(mech.pParameter as *const CK_OTP_PARAMS) };
+            if params.ulCount > 0 && params.pParams.is_null() {
+                return Err(CKR_MECHANISM_PARAM_INVALID)?;
+            }
```
Jakuje: Should we use the helper as in the other mechanisms?
```
    let params = mech.get_parameters::<CK_OTP_PARAMS>()?;
```

src/hotp.rs
```diff
@@ -0,0 +1,699 @@
...
+            .get_attr_as_ulong(CKA_OTP_LENGTH)
+            .map_err(|_| CKR_GENERAL_ERROR)? as usize;
+
+        let key_val = key.get_attr_as_bytes(CKA_VALUE)?.clone();
+        let (hmac_mech, mac_len) = match key_val.len() {
+            20 => (CKM_SHA_1_HMAC, 20),
+            32 => (CKM_SHA256_HMAC, 32),
+            64 => (CKM_SHA512_HMAC, 64),
+            _ => return Err(CKR_GENERAL_ERROR)?,
+        };
```
Jakuje, line 474: the code mapping the key_val.len() to mac_len by enumeratning the lengths sounds odd. Could we de-obfuscate it a bit?
```suggestion
        let mac_len = key_val.len();
        let hmac_mech, = match mac_len {
            20 => CKM_SHA_1_HMAC,
            32 => CKM_SHA256_HMAC,
            64 => CKM_SHA512_HMAC,
            _ => return Err(CKR_GENERAL_ERROR)?,
        };
```

src/hotp.rs
```diff
@@ -0,0 +1,699 @@
...
+
+        factory.data.finalize();
+
+        factory
+    }
+
+    pub fn validate_object(&self, obj: &mut Object) -> Result<()> {
+        obj.ensure_ulong(CKA_CLASS, CKO_OTP_KEY)
+            .map_err(|_| CKR_TEMPLATE_INCONSISTENT)?;
+        obj.ensure_ulong(CKA_KEY_TYPE, CKK_HOTP)
```
Jakuje, line 141: I am surprised to see the (time based) TOTP is not defined in the PKCS#11. Though I agree getting that one working in HW with precise clock is non-trivial.

src/object/otp.rs
simo5: Unsure, the SecurID standard is weird, they had different vesions using different algorithms.
I thought for a while to use it for TOTP, given there is no CKK_TOTP key, but not 100% sure.

src/hotp.rs
simo5: probably, yes

simo5: adding also some more unsafe handlers in the usual places to make this less handcrafty

src/hotp.rs
simo5: what about using constants instead of numbers to make it clear instead ?

src/hotp.rs
simo5: the time can be passed in input as well in theory ... I think this is mostly and abandoned part of the spec, but I can revive it if needed. For now I am content with just HOTP.

src/object/otp.rs
Jakuje: So lets drop the attributes we do not have a use for?

simo5: This is the general OTP factory, so I think we should list all attributes ?
Ah but I see technically that attribute is available only for CKK_SECURID ... uhmm, ok I'll drop.

simo5: done

Jakuje: Yeah, that was my point that its not defined for generic OTP Key (unlike the rest).

src/hotp.rs
```diff
@@ -0,0 +1,674 @@
...
+            if req != CK_OTP_PARAM_IGNORED {
+                return Err(CKR_ATTRIBUTE_VALUE_INVALID)?;
+            }
+        }
+
+        if let Ok(interval) = obj.get_attr_as_ulong(CKA_OTP_TIME_INTERVAL) {
+            if interval != CK_UNAVAILABLE_INFORMATION {
+                return Err(CKR_ATTRIBUTE_VALUE_INVALID)?;
+            }
+        }
```
Jakuje: You should also drop it from here, I htink.

simo5: true

simo5: done

Jakuje (approved):

simo5:
The op_state test fails because of a series of things with openssl 4 hitting rawhide.
I have a PR open here: https://github.com/openssl-projects/pkcs11-provider/pull/747 but it will take a while to merge due to other consequences of the above.
So merging this now as it passes all other tests.

PR #465, opened 2026-06-22
Add HOTP feature to standard build

simo5:
#### Description

Adds the hotp feature to the standard feature list in Cargo.toml, enabling CKM_HOTP and OTP token support by default. The changelog is also updated to document these newly added mechanisms.

#### Checklist

- [ ] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/src/tests/hotp.rs
+++ b/src/tests/hotp.rs
@@ -115,6 +115,7 @@ fn test_hotp() {
     testtokn.finalize();
 }
 
+#[cfg(not(feature = "no_sha1"))]
 #[test]
 #[parallel]
 fn test_hotp_vectors() {
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -52,7 +52,7 @@ jobs:
                 'crate(clap)' 'crate(clap/cargo)' 'crate(clap/derive)' 'crate(clap/help)' \
                 'crate(clap/std)' 'crate(clap/usage)' 'crate(constant_time_eq/default)' \
                 'crate(data-encoding/default)' 'crate(hex/default)' \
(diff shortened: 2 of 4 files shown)


simo5:
@Jakuje sigh, had forgotten to add the feature in a way that it would be tested in CI.
I added it to the standard build as it costs nothing to do so.

Jakuje:
:facepalm: The HOTP test vector is using SHA1 which makes it failing now?

simo5:
> 🤦 The HOTP test vector is using SHA1 which makes it failing now?

Ugh...

Jakuje:
whats the dependency issue with the itertools? Otherwise it looks ok

simo5:
> whats the dependency issue with the itertools? Otherwise it looks ok

I think something just landed on rawhide and not all of the mirrors are in sync yet ?
Going back and forth to figure it out ...

simo5:
@Jakuje I think it is good now, PTAL

src/tests/hotp.rs
```diff
@@ -115,6 +115,7 @@ fn test_hotp() {
     testtokn.finalize();
 }
 
+#[cfg(not(feature = "no_sha1"))]
```
Jakuje, line 118: just a nit that the negative feature (no_sha1) written in the cfg like this (not(no_sha1)) makes the double negation hard to comprehend. But thats for another PR to decide if we should just disable sha1 by default and have `sha1` feature.

Jakuje (approved):

src/tests/hotp.rs
simo5: it is the same elsewhere, once we decide to make no sha1 the default we'll invert the feature and people we'll had to --feature sha1 to make it work

simo5:
thanks for the review!

PR #466, opened 2026-06-24
Build against openssl-sys low-level bindings

Jakuje:
#### Description

Linking applications against different bindings to the same library (openssl) is undefined behavior in rust ecosystem.

This is removing the low-level bindings and bindgen dependency, delefating this to the openssl-sys crate providing all the low-level functions.

Unfortunately, it does not expose all the constants and defines, that need to be defined manually.

The direct static linking to libfips is preserved.

#### Checklist

- [ ] Test suite updated

#### Reviewer's checklist:


ossl/src/non-openssl-sys-bindings.rs
```diff
@@ -0,0 +1,159 @@
+// handwritten OpenSSL constants
```
simo5, line 1: I do not think we should have handwritten bindings. but simply generate new bindings where we filter and expose only what's missing in openssl-sys

simo5: Then we should ask upstream if they are ok exposing these.

simo5 (changes requested):

ossl/src/non-openssl-sys-bindings.rs
Jakuje: Technically, this is just the subset of generated constants we are using I copied from the bindgen output we have. I think the openssl-sys does not expose any defines so I thought about this as a short-term solution before we will either convince them or use something else.

simo5: I still think we should autogenerate it via bindgen, rather than manually handling it. You can change bindgen to filter everything and only leave thee defines.

Jakuje: Thats probably sane idea as we will still need the bindgen in the openssl-sys dependency. I will put it together.

ngg: Hi! Thx for implementing this, looks great. openssl-sys uses a different mechanism to find openssl than ossl did before (OPENSSL_DIR envvar vs pkg-config), pls make sure to match how you find openssl if you decide to use bindgen on top of openssl-sys. Also, my use-case would require using openssl-sys and its "vendored" feature, because my dependency graph contains both sequoia -> ossl path and other crates using openssl crate, and I'll need vendoring support for some platforms. Currently with this PR it seemed working that I define both ossl's "dynamic" feature and openssl-sys's "vendored" feature, pls keep it a supported use case. In that case openssl-sys pulls in the openssl-src crate and uses header files from that, so bindgen should use that too.

Jakuje: Thank you for the input! This makes me thinking that hardcoded list of defines should be sufficient now. We can probably work on the getting the constants generated later if needed as from what I see it would require some unification of how the openssl directory is discovered.

simo5: When you say "vendored" I assume you mean statically linked?

ngg: I'd like to suggest renaming the "dynamic" feature to "openssl-sys" or something similar, it would be more clear because it doesn't necessarily implies it will be actually dynamically linked. You could keep "dynamic" as an alias for compatibility if needed.

ngg: Yeah the feature of the openssl-sys crate is called "vendored", it means it will download openssl-src, build it and link that statically

simo5: well I have been thinking for a while to completely drop dynamic, because statically linking openssl 3.0+ is .. kinda broken anyway, although it works for limited cases when you do not need to use any additional provider ...

So perhaps we just drop dynamic, but does "vendored" always mean static linking of openssl in openssl-sys ?

simo5: also we need to explicitly provide you a way to express you want the "vendored" feature of openssl-sys so we'll have to exposed something equivalent. We could pull that feature in when "dynamic" is not set if "vendored" does indeed mean static linking

simo5: the problem though is that I never want to build against openssl-src `version = "300.2.0"` (whatever it is) when building statically and also do not care for enabling legacy providers , so perhaps we'll need a separate option there for static build that does not use openssl-sys

ngg: In my experience, static linking openssl 3.x works perfectly if you want to use its built-in providers only, but I have no experience with either its fips mode or extra providers.
If you use its "vendored" feature, it uses the openssl-src crate's builder, which builds openssl with its `no-shared` option, so it can only be linked statically: https://github.com/alexcrichton/openssl-src-rs/blob/fca4b1282a9c0edc2403f89b94821f382f739c59/src/lib.rs#L199

simo5: if you build statically fips is out of the window, but we do not care about that case, vendoring should be last resort anyway.

simo5 (commented):
github is too stupid to realize on its own when a comment is just a comment and whwn it is part of a review ... liberating all my "pending" comments

ossl/src/lib.rs
```diff
@@ -16,11 +17,18 @@ pub mod bindings {
     #![allow(unnecessary_transmutes)]
     include!(concat!(env!("OUT_DIR"), "/ossl_bindings.rs"));
 }
+#[cfg(feature = "dynamic")]
+pub mod bindings {
```
ngg, line 21: It would be great to re-export everything publicly from openssl-sys in the `bindings` mod, then other functions can be used as before.

ngg: it would also simply the other files, no need to import openssl_sys directly, only the bindings mod

ossl/src/non-openssl-sys-bindings.rs
ngg: I don't think it's possible to pull in a dep's feature if a feature is NOT set, because that would violate features being additive.
Even if this crate does not depend on "vendored", I can already use it by adding an explicit openssl-sys dependency with the vendored feature next to when I pull in sequoia or ossl.

ossl/src/lib.rs
ngg: I tried to compile sequoia-openpgp with this version of ossl, it currently fails because it would like to access a few things directly via `ossl::bindings::`

ngg:
While I was checking the bindings, I found a small issue with the current bindings generator, it's not including function names matching `OpenSSL_*` like `OpenSSL_version()` (just the full lowercase and full uppercase prefixes are allowlisted).

ossl/src/lib.rs
simo5: yes I was also going to make that suggestion

ngg: > the problem though is that I never want to build against openssl-src version = "300.2.0" (whatever it is) when building statically and also do not care for enabling legacy providers , so perhaps we'll need a separate option there for static build that does not use openssl-sys

I agree, you should use your current solution for your static building use case, especially if you want to have custom patches. My use case is different, openssl-src is good for me, that's basically tracking upstream openssl (currently version `300.6.1+3.6.3` corresponding to upstream openssl 3.6.3)

simo5: Upstream released 4.0 months ago ...

ngg: Yes, I know and I don't know why they don't use it yet. But 3.6.3 is also the latest security update from 2 weeks ago on the 3.6.x series, so it's updated at least. I don't need 4.x features, so I'm okay with having a 3.6.x if it has all security fixes. I guess they'll update to 4.x sooner or later

ossl/src/pkey.rs
```diff
@@ -828,7 +830,7 @@ fn params_to_slhdsa_data(
                     // asked for key pair here so if it not available,
                     // retry exporting just public key part
                     // https://github.com/openssl/openssl/issues/27542
-                    let p2 = pkey.export_params(EVP_PKEY_PUBLIC_KEY)?;
+                    let p2 = pkey.export_params(EVP_PKEY_PUBLIC_KEY as c_int)?;
```
simo5: why this change ?

ossl/src/lib.rs
simo5: doesn't work for us, we need to be able to serialize digest contexts and that requires 4.0

ngg: I've created a PR for openssl-src crate bumping to 4.0.1: https://github.com/alexcrichton/openssl-src-rs/pull/286

ossl/src/pkey.rs
Jakuje: Our bindings generate the `EVP_PKEY_PUBLIC_KEY` constant (and similar) as `u32`, while openssl-sys has has it handwritten as `c_int` (which is more correct to the underlying C define not mentioning any unsigned modifier):
```
#define OSSL_KEYMGMT_SELECT_PUBLIC_KEY 0x02
...
#define EVP_PKEY_PUBLIC_KEY \
    (EVP_PKEY_KEY_PARAMETERS | OSSL_KEYMGMT_SELECT_PUBLIC_KEY)
```
Other option might be changing our definition to us `i32`/`c_int`.

ossl/src/lib.rs
Jakuje: > I tried to compile sequoia-openpgp with this version of ossl, it currently fails because it would like to access a few things directly via ossl::bindings::

Good point. That was the next thing I wanted to try.

> It would be great to re-export everything publicly from openssl-sys in the bindings mod, then other functions can be used as before.

Let me try that.

Jakuje:
I tried to build sequoia-openpgp against this, but it pulls older version of openssl-sys (through `native-tls`, which is missing some of the features that we need) and causes the build failure (as it looks like we can use only one of the versions in the build process).

ngg:
> I tried to build sequoia-openpgp against this, but it pulls older version of openssl-sys (through `native-tls`, which is missing some of the features that we need) and causes the build failure (as it looks like we can use only one of the versions in the build process).

What commit of sequoia-openpgp did you try? You should use the `main` branch, it works for me, I just get a few missing `ossl::bindings::*` errors. I modified `openpgp/Cargo.toml` to use a local path for ossl, then simply ran `cargo test -p sequoia-openpgp --no-default-features --features crypto-openssl`

Jakuje:
I had some older branch, but I was able to provide replacements defines. Now I should have all the remaining defines for sequoia to build (from my branch as well as from main). The only failures I see are the leak_checks

ossl/src/lib.rs
Jakuje: Thanks for suggestion! This was done and works fine.

ngg:
> I had some older branch, but I was able to provide replacements defines. Now I should have all the remaining defines for sequoia to build (from my branch as well as from main). The only failures I see are the leak_checks

Great. The leak_checks failures are pre-existing, I have those too.

Jakuje:
> openssl-sys uses a different mechanism to find openssl than ossl did before (OPENSSL_DIR envvar vs pkg-config), pls make sure to match how you find openssl if you decide to use bindgen on top of openssl-sys.

My reading is that openssl-sys is also using pkg-config to find OpenSSL (with some handmade overrides in this file):

https://github.com/rust-openssl/rust-openssl/blob/master/openssl-sys/build/find_normal.rs#L210

I would like to avoid reimplementing this search inside of the ossl crate, which is why I did use the hardcoded list of constants we use rather than pulling bindgen again. I would love to get them into openssl-sys, but I think this is not a critical part.

We have a different semantics to build the OpenSSL for fips build, but I tried not to touch the fips build paths.

> Also, my use-case would require using openssl-sys and its "vendored" feature, because my dependency graph contains both sequoia -> ossl path and other crates using openssl crate, and I'll need vendoring support for some platforms. Currently with this PR it seemed working that I define both ossl's "dynamic" feature and openssl-sys's "vendored" feature, pls keep it a supported use case. In that case openssl-sys pulls in the openssl-src crate and uses header files from that, so bindgen should use that too.

This sounds to me like the least problematic solution. If we would like to pass the vendored feature to openssl-sys from ossl, I guess you would need also changes in sequoia to pass it from there, which does not sound to me like sustainable.

I hope I followed up on all questions and concerns. Please, let me know if I missed something or if there are some concerns. Changing to ready for review.

ossl/src/lib.rs
```diff
@@ -16,6 +17,15 @@ pub mod bindings {
...
+    #![allow(non_upper_case_globals)]
+    include!("missing_ossl_bindings.rs");
```
simo5: can we rename this to non-openssl-sys-bindings.rs ?

ossl/src/lib.rs
```diff
@@ -43,10 +53,27 @@ pub mod fips;
 /// This returns a triplet of u8 values representing Major, Minor, and
 /// Patch version.
 pub fn api_level() -> (u8, u8, u8) {
-    let patch: u8 = (OPENSSL_API_LEVEL & 0xff) as u8;
-    let minor: u8 = ((OPENSSL_API_LEVEL & 0xff00) >> 8) as u8;
-    let major: u8 = ((OPENSSL_API_LEVEL & 0xff0000) >> 16) as u8;
-    (major, minor, patch)
+    #[cfg(not(feature = "dynamic"))]
```
simo5: we should have an openssl-sys feature I think rather then overloading "dynamic"

ossl/src/pkey.rs
```diff
@@ -756,7 +756,8 @@ fn params_to_mldsa_data(
                     // asked for key pair here so if it not available,
                     // retry exporting just public key part
                     // https://github.com/openssl/openssl/issues/27542
-                    let p2 = pkey.export_params(EVP_PKEY_PUBLIC_KEY)?;
+                    let p2 =
+                        pkey.export_params(EVP_PKEY_PUBLIC_KEY as c_int)?;
```
simo5: why this change?

ossl/src/pkey.rs
```diff
@@ -1081,14 +1083,14 @@ impl EvpPkey {
     ///
     /// The `selection` argument specifies which components to export
     /// (e.g., public, private, parameters).
-    fn export_params(&self, selection: u32) -> Result<OsslParam<'_>, Error> {
+    fn export_params(&self, selection: c_int) -> Result<OsslParam<'_>, Error> {
```
simo5: This is an API change which would force us to release a new semver major version, so I would rather avoid it

ossl/build.rs
```diff
@@ -12,68 +13,63 @@ const OPENSSL_3_2_0: i64 = 0x30200000;
...
-        }
-        if name == "OSSL_PKEY_PARAM_ML_DSA_SEED" {
-            println!("cargo::rustc-cfg=ossl_mldsa")
-        }
-        if name == "OSSL_PKEY_PARAM_ML_KEM_SEED" {
-            println!("cargo::rustc-cfg=ossl_mlkem")
-        }
-    }
-
+    // FIXME for openssl-sys
```
simo5: what's the issue here?

simo5 (changes requested):
I want to retain the ability to make builds w/o openssl-sys, as we want to be able to pull in a custom configured openssl static build, rather than have to deal with the openssl-src package (which is not on 4.0 yet either).

So I think we should have a openssl-sys feature, and potentially move the actual bindings generation to a ossl-sys subpackage in order to allow to unconflicts the links = "openssl" package key in the toml

ossl/build.rs
Jakuje: I did not reimplement detection of this API that is new in OpenSSL 4.0 for openssl-sys backend. So even if they add it it will not be detected as of now.

ossl/src/pkey.rs
Jakuje: Yeah, this is the mixing of the types that we export under `bindings`. For example `EVP_PKEY_PUBLIC_KEY`. If they come from openssl-sys its `c_int`, if they come form our bindgen, its `u32`. This standardizes on the openssl-sys to avoid need to change it in sequoia, but I would be happy to come up with different solution.

ossl/src/pkey.rs
Jakuje: see the previous comment -- different type from openssl-sys and from our bindgen.

ossl/src/lib.rs
Jakuje: Good point. I will make an openssl-sys feature then.

ossl/build.rs
```diff
@@ -12,68 +13,63 @@ const OPENSSL_3_2_0: i64 = 0x30200000;
...
-        if name == "OSSL_PKEY_PARAM_SLH_DSA_SEED" {
-            println!("cargo::rustc-cfg=ossl_slhdsa")
-        }
-        if name == "OSSL_PKEY_PARAM_ML_DSA_SEED" {
-            println!("cargo::rustc-cfg=ossl_mldsa")
-        }
-        if name == "OSSL_PKEY_PARAM_ML_KEM_SEED" {
-            println!("cargo::rustc-cfg=ossl_mlkem")
-        }
-    }
```
Jakuje, line 68: I did not find a reason why these were done in this way rather than by the version, but changed them to check based on the version, hopefully without any functional damage.

ossl/src/pkey.rs
Jakuje: This is not a `pub` function so I think its not an API anyway.

ossl/src/pkey.rs
Jakuje: 1. Removed this by using custom bitfield

ossl/src/lib.rs
Jakuje: Added

ossl/src/lib.rs
Jakuje: Done.

ossl/src/lib.rs
```diff
@@ -43,10 +53,27 @@ pub mod fips;
 /// This returns a triplet of u8 values representing Major, Minor, and
 /// Patch version.
 pub fn api_level() -> (u8, u8, u8) {
```
ngg, line 55: Can we make this function return the runtime api level instead of build-time? It can be different if it's dynamically linked, runtime would make more sense. There are functions like `OPENSSL_version_major()` for this.

Also a function returning the full version string would be great, you can get that by calling `OpenSSL_version(OPENSSL_FULL_VERSION_STRING)`

Jakuje: I guess given that this is already in API, we do not want to change it. We might want a runtime version or full version string too, but it can be a separate change in separate API.

ossl/build.rs
simo5: you should be able to (conditionally) pull the openssl-sys package in the build.rs and check if the version is 4_0 and retain the check ?

ossl/build.rs
simo5: The SEED parameter was added only after the initial introduction of the functions in 3.2(?) I am ok gating this to 3.5 now.

ossl/build.rs
Jakuje: In that case I would just prefer to rewrite the check to `ossl_v400` in the code

ossl/src/lib.rs
simo5: agree, that's the api_level ossl was build against

if you build against the v3.0.7 api level no matter if you find at runtime that you are using 3.5, some things will simply not work, so both numbers are kind of relevant.

ossl/build.rs
simo5: yes that is ok by me

ossl/Cargo.toml
```diff
@@ -12,14 +11,17 @@ license = "Apache-2.0"
 test = true
 
 [build-dependencies]
+# this is technically optional, but we either use it directly or through openssl-sys
```
simo5: well this is the bindgen for building the bindings, the openssl-sys one is just for exposing them so this comment is not accurate, I qould drop it.

ossl/src/pkey.rs
```diff
@@ -11,6 +11,17 @@ use crate::{
     cstr, trace_ossl, Error, ErrorKind, OsslContext, OsslParam,
     OsslParamBuilder, OsslSecret,
 };
+use bitflags::bitflags;
+
+bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct PkeyClass: u32 {
+        const NONE        = 0;
+        const PRIVATE_KEY = 133;
```
simo5: what are these magic numbers?
I think these should be defined in terms of binding variables?

ossl/Cargo.toml
Jakuje: With the bindgen feature we use, they also run bindgen:

https://github.com/rust-openssl/rust-openssl/blob/master/openssl-sys/build/run_bindgen.rs

I think without that they use just the handwritten ones that are missing what we need.

ossl/src/pkey.rs
Jakuje: Done the intermediate calculation (hopefully correctly).

ossl/build.rs
Jakuje: Done

ossl/src/pkey.rs
```diff
@@ -986,7 +1013,7 @@ fn dsa_data_to_params(
 ///
 /// Critically this function will zeroize allocated parameters after makig a
 /// full copy of them, before returning control to OpenSSL.
-#[cfg(not(param_clear_free))]
+#[cfg(not(ossl_v400))]
```
simo5, line 989: sounds like this conditional was always wrong and should just be removed

Jakuje: good catch!

Jakuje: nah. Its more complicated than that. But it sounds like the detection never worked as you are calling the `EVP_PKEY_todata()` with 4 arguments with the openssl 4 when I change this. I am trying to wrap my head around why we should use the `EVP_PKEY_todata()` ...

Jakuje: This came from #351 which has a note about the reason not use the ossl_v400 as it was not yet released at that time.

Jakuje: Seems like this is still broken after recent changes. It failed for me at least once with OpenSSL 4.0 in the integration tests with pkcs11-provider:
```
 openssl pkeyutl -sign
              -in ${TMPPDIR}/sha256.bin
Could not find or decode private key from /__w/kryoptic/kryoptic/pkcs11-provider/builddir/tests/kryoptic/file.ec.key.pem
00E3D9B2077F0000:error:1E08010C:DECODER routines:OSSL_DECODER_from_bio:unsupported:crypto/encode_decode/decoder_lib.c:104:No supported data to decode.  Input type: DER, Input structure: PrivateKeyInfo
pkeyutl: Error loading key
```
But it did not fail during rerun.

Jakuje:
Seems like it will be more complex to spin up the ossl-sys as it looks like cargo does not support mutually exclusive dependencies on feature level. So the ways out from here are likely:

 * rename the `links` on ossl-sys to something like ossl to make cargo happy (this wont prevent anyone from linking to two different openssl's)
 * Just not provide the `links` (with the very same outcome as its done now).

(unless I miss something)

That said, I would probably prefer to stick with this version instead of creating yet another crate just for the bindings (even though I have the work mostly done) as it really does not justify the complexity added. But dropping the branch here anyway:

https://github.com/Jakuje/kryoptic/tree/ossl-sys

ossl/src/pkey.rs
simo5: I guess we should check how file.ec.key.pem is generated and see if there are cases where specific key values cause incompatibilities?
We had the recent issue with P521 keys where we were too strict with key buffer length, I wonder if something similar is happening but less frequently for other key types?

Jakuje: Yes, it could be. I will move this to the pkcs11-provider issue for follow-up investigation

Jakuje: oh, it was kryoptic issue. So I will move it to separate kryoptic issue :)

simo5:
> Seems like it will be more complex to spin up the ossl-sys as it looks like cargo does not support mutually exclusive dependencies on feature level. So the ways out from here are likely:
> 
>     * rename the `links` on ossl-sys to something like ossl to make cargo happy (this wont prevent anyone from linking to two different openssl's)
> 
>     * Just not provide the `links` (with the very same outcome as its done now).
> 
> 
> (unless I miss something)
> 
> That said, I would probably prefer to stick with this version instead of creating yet another crate just for the bindings (even though I have the work mostly done) as it really does not justify the complexity added. But dropping the branch here anyway:
> 
> https://github.com/Jakuje/kryoptic/tree/ossl-sys

we just make ossl-sys optional at the ossl crate level (just like openssl-sys is) and pull it in when dynamic, static or fips are selected, but not when openssl-sys is selected.

ossl/build.rs
```diff
@@ -322,19 +308,42 @@ fn set_pretty_panic() {
...
+            let minor = (version >> 20) & 0xFF;
+            let patch = (version >> 4) & 0xFF;
+            let version_string = format!("{}.{}.{}", major, minor, patch);
+            println!(
+                "cargo:rustc-env=OPENSSL_FULL_VERSION_STR={}",
+                version_string
+            );
+        } else {
+            panic!(
+                "No OpenSSL version detected. The openssl-sys did not provide\
```
simo5: "The openssl-sys" -> just "openssl-sys"

ossl/src/lib.rs
```diff
@@ -8,13 +8,23 @@
 /// Part of this module is automatically generated by bindgen from the OpenSSL
 /// Headers and includes a selection of functions and other items needed to
 /// access the libcrypto/libfips functions needed.
+#[cfg(not(feature = "openssl-sys"))]
```
simo5: This should now be `#[cfg(feature = "ossl-sys")]`

ossl/src/lib.rs
```diff
@@ -43,10 +53,27 @@ pub mod fips;
 /// This returns a triplet of u8 values representing Major, Minor, and
 /// Patch version.
 pub fn api_level() -> (u8, u8, u8) {
-    let patch: u8 = (OPENSSL_API_LEVEL & 0xff) as u8;
-    let minor: u8 = ((OPENSSL_API_LEVEL & 0xff00) >> 8) as u8;
-    let major: u8 = ((OPENSSL_API_LEVEL & 0xff0000) >> 16) as u8;
-    (major, minor, patch)
+    #[cfg(not(feature = "openssl-sys"))]
```
simo5: This whole part can be compressed to make optional only how we get to `version`, the rest is identical.

ossl/src/lib.rs
Jakuje: Right!

ossl-sys/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+bindgen = "0.72"
+pkg-config = "0.3"
+
+[dependencies]
+cfg-if = "1.0.0"
+libc = "0.2.151"
+
+[features]
+dynamic = [] # Builds against system libcrypto.so
+fips = [] # Builds against sources and libfips.a instead of libcrypto
```
simo5, line 28: we probably need to add a `static = []` here and a `default = ["dynamic"]` ?

ossl/Cargo.toml
```diff
@@ -16,24 +15,27 @@ bindgen = "0.72"
...
+
 ossl320 = [] # Requires at a minimum OpenSSL 3.2.0
 ossl350 = [] # Requires at a minimum OpenSSL 3.5.0
 ossl400 = [] # Requires at minimum OpenSSL 4.0
-dynamic = [] # Builds against system libcrypto.so
-fips = ["ossl350"] # Builds against sources and libfips.a instead of libcrypto
+
+openssl-sys = ["dep:openssl-sys"] # Build against openssl-sys bindings
+dynamic = ["ossl-sys/dynamic"] # Builds against system libcrypto.so
+fips = ["ossl350", "ossl-sys/fips"] # Builds against sources and libfips.a instead of libcrypto
```
simo5: Should the comments here be moved to the ossl-sys Cargo.toml file ?

simo5 (changes requested):
We need to add a check like we have in kryoptic's build.rs to ossl/build.rs:

```
#[cfg(all(feature = "openssl-sys", feature = "ossl-sys"))]
compile_error!("openssl-sys and ossl-sys are incompatible features as they both cause linking to OpenSSL");
```

Then change most places where we have `[#cfg(not(feature = "openssl-sys"))]` to `[#cfg(feaute = "ossl-sys")]`

ossl/Cargo.toml
Jakuje: Yeah, that would make sense.

ossl-sys/Cargo.toml
Jakuje: Ack for the default dynamic.

Not sure about the static. We do not have "static" feature so far so I am not sure what would be the best way to do that. I think so far the static build was the build without "dynamic" feature and with --no-default features, which is kind of odd. If we want separate static feature like this, we will likely need some more handling of it flying around.

ossl/src/lib.rs
Jakuje: Its  not. The API_LEVEL is `0xMMNNPP`, while OPENSSL_VERSION is `0xMNN00PP0` to be more confusing

Jakuje: But yes, I can use now `OPENSSL_VERSION` for both cases.

Jakuje:
> We need to add a check like we have in kryoptic's build.rs to ossl/build.rs:
> 
> ```
> #[cfg(all(feature = "openssl-sys", feature = "ossl-sys"))]
> compile_error!("openssl-sys and ossl-sys are incompatible features as they both cause linking to OpenSSL");
> ```

This is already in ossl/build.rs

> Then change most places where we have `[#cfg(not(feature = "openssl-sys"))]` to `[#cfg(feaute = "ossl-sys")]`

Will do.

The CI now fails in the cases where we implicitly assume `static`. I will make it explicit feature and add some more checks to make sure there are not both static and dynamic.

ossl-sys/Cargo.toml
simo5: sorry I didn't write this right I think

the feault should be that we get a build dynamically linking to to openssl
but we do not need a dynamic feature for now ?

Jakuje: I think thats what I did in the recent changes. If not, please clarify

ossl/Cargo.toml
```diff
@@ -16,24 +15,30 @@ bindgen = "0.72"
...
 ossl320 = [] # Requires at a minimum OpenSSL 3.2.0
 ossl350 = [] # Requires at a minimum OpenSSL 3.5.0
 ossl400 = [] # Requires at minimum OpenSSL 4.0
-dynamic = [] # Builds against system libcrypto.so
-fips = ["ossl350"] # Builds against sources and libfips.a instead of libcrypto
+
+openssl-sys = ["dep:openssl-sys"] # Build against openssl-sys bindings
+ossl-sys = ["dep:ossl-sys"] # Build against ossl-sys bindings
+
+dynamic = ["ossl-sys", "dep:ossl-sys"]
```
simo5: `"dep:ossl-sys"` is redundant here, already implied by `ossl-sys`

ossl/Cargo.toml
```diff
@@ -16,24 +15,30 @@ bindgen = "0.72"
...
+static = ["ossl-sys", "ossl-sys/static"]
+fips = ["ossl-sys", "ossl350", "ossl-sys/fips"]
```
simo5: can you bump this to "ossl400" ?

simo5 (changes requested):
Minor nits only, otherwise LGTM

Can you add a CHANGELOG entry too ?

ossl/Cargo.toml
```diff
@@ -16,24 +15,30 @@ bindgen = "0.72"
...
 vsprintf = { version = "2.0.0", optional = true }
+openssl-sys = { version = "0.9", features = ["bindgen"], optional = true }
+ossl-sys = { version = "1.5.1", optional = true, path = "../ossl-sys" }
 
 [dev-dependencies]
 hex = "0.4.3"
 serial_test = "3.1.1"
 
 [features]
+default = ["openssl-sys"]
```
ngg: I'm not sure whether you want to stay compatible with latest release, but if a crate depends on ossl with its `dynamic` feature like how sequoia does, it won't compile now because `dynamic` is not compatible with `default`.
I don't really think it's possible to stay compatible with the current design, the old "default" is closest to the new ossl-sys/static, but the old "default + dynamic" would correspond to ossl-sys without static.
The cleanest solution would probably be releasing `ossl v2.0.0` and not even trying to stay compatible and removing the `dynamic` feature all together.

simo5: Kryoptic still builds with ossl-sys dynamically linked, so dynamic is still relevant at the ossl level, otherwise the only way to dynamically link would be against openssl-sys, and I want an escape hatch to continue developemnt w/o having to hardcode local patched copies of openssl-sys when they do not expose stuff I need.

If you think we should not change the default, that is fine, we can retain the default as it was, and have openssl-sys be only an opt-in option

simo5: In fact, thinking more about it that is what we should probably do, @jakuje can we drop "openssl-sys" from default and go back to "dynamic" for kryptic-lib and to "ossl-sys" for ossl ?

ngg: `EvpPkey::fromdata()` is not compatible either with the default feature set, the new default uses openssl-sys which has these constants as `i32`, and `u32: From<i32>` is not satisfied, so if it's used with an argument like `ossl::bindings::EVP_PKEY_KEYPAIR`, then it no longer compiles.

ngg: I didn't mean to remove the possibility to use dynamic linking with ossl-sys, I just meant to remove the actual "feature" called `dynamic`, because it doesn't do anything now, same as the `ossl-sys` feature

ngg: If you don't want to release a major version of `ossl` and keep it compatible, then the default feature set should continue to mean the statically linked `ossl-sys` version and the `dynamic` feature should mean the dynamically linked `ossl-sys` version as before.
But this doesn't seem to be compatible with how ossl-sys now has a `static` feature instead of a `dynamic` feature. I like the `static` feature better, it's more symmetrical to `openssl-sys` this way which also defaults to dynamic linking, I just don't know if it's possible to have that and also keep compatibility.

simo5: we can easily change the features and keep ossl-sys as is (ossl/dynamic -> oss-sys, ossl -> ossl-sys/static)

but the incompatibility with openssl-sys EVP_PKEY_KEYPAIR is a problem

one way out is to *not* export all of openssl-sys bindings automatically, but instead manually export allow the compatible ones and override the incompatible ones with the old versions ... 

this is getting ugly though :(

simo5: I think we should merge this PR with just the minor tweak I requested though, and next week we can decide if we should release a 2.0 version of ossl and make more serious adjustments to the API, there are a few things that may be nice to change after all.

ngg: If the default of `ossl` stays using `ossl-sys`, then the pkey and similar incompatibilities are not a problem, because it will only cause errors when explicitly switching to `openssl-sys`.

I think if you remove `ossl-sys/static` feature but add `ossl-sys/dynamic` instead, you can have ossl's `default` to use `ossl-sys` (meaning static), ossl's `dynamic` forward to `ossl-sys/dynamic`, and also having neither sys features set defaulting to use ossl-sys, then it should stay compatible with 1.x versions for all the already existing feature combinations (including no-default-feature).

Then later if you choose to release 2.x, you can clean up the features a bit if you want to.

simo5: yeah this is definitely an option, in order to make a quick release, and then we work on improving the API for a 2.0, the goal would be that applications should not need to use bindings at all but only ossl specific wrappers in most cases

Jakuje: > In fact, thinking more about it that is what we should probably do, @Jakuje can we drop "openssl-sys" from default and go back to "dynamic" for kryptic-lib and to "ossl-sys" for ossl ?

Will do. Makes sense.

> EvpPkey::fromdata() is not compatible either with the default feature set, the new default uses openssl-sys which has these constants as i32, and u32: From<i32> is not satisfied, so if it's used with an argument like ossl::bindings::EVP_PKEY_KEYPAIR, then it no longer compiles.

Right. This mismatch is pain. I have a simple patch for sequoia to use the new PKeyClass (rather than depending on the constants from bindings::). Reverting the default to `dynamic` from ossl-sys will keep the existing code working, I think.

ngg:
I've created a [test branch](https://gitlab.com/sequoia-pgp/sequoia/-/commits/ngg1/ossl-test) with 2 relevant commits:
- the [first commit](https://gitlab.com/sequoia-pgp/sequoia/-/commit/272fcbd00ea522956b13e39eb587a798683f376c) just updates ossl revision to the current https://github.com/latchset/kryoptic/commit/f4a885d16dcc9873c256171fe6f7bac5b686d7b6 commit without any other changes, showing this new version is compatible with the released one for sequoia
- the [second commit](https://gitlab.com/sequoia-pgp/sequoia/-/commit/1dc5ba8b782d96bd8c54507d9d97c6f6cc8726e7) switches to use `openssl-sys`

Both looks ok.

Jakuje:
Thank you for double-checking. I had similar changes locally. Lets wait for (hopefully) final review from Simo to get this merged!

ossl-sys/README.md
```diff
@@ -0,0 +1,11 @@
+# ossl-sys
+
+A crate providing low-level Rust bindings to the OpenSSL library.
+
+`ossl` was spun off from the `kryoptic` project to provide a focused,
```
simo5: we need an ossl-sys blurb here

ossl/src/pkey.rs
```diff
@@ -1049,10 +1082,14 @@ impl EvpPkey {
...
     /// (e.g., modulus/exponent for RSA, curve/point for EC).
+    ///
+    /// The `pkey_type` argument specifies which components we import
+    /// (e.g., PkeyClass::PRIVATE_KEY, PkeyClass::PUBLIC_KEY,
+    /// PkeyClass::KEYPAIR, ...).
     pub fn fromdata(
         ctx: &OsslContext,
         pkey_name: &CStr,
-        pkey_type: u32,
+        pkey_type: impl Into<u32>,
```
simo5: doesn't this break with openssl-sys anyway ?

simo5 (approved):
just a nit and a question, but overall LGTM

ossl/src/pkey.rs
Jakuje: Yes, just switching our user to openssl-sys would break. But this is switching the feature which is kind of expected to require some changes. We might consider documenting them somewhere though

ossl-sys/README.md
Jakuje: Will do!

ossl/src/pkey.rs
simo5: well given we *do* have features, why not provide a signature function that accepts i32 when compiled against openssl-sys ?
That would reduce friction until we change the API to accept an enum or other ossl specific wrapped input ?

simo5: I know this is a bit of an undoing, but if we remove the PkeyClass thing and defer it to anther PR, then we can simply accept u32 with ossl-sys and i32 with openssl-sys.

The idea is to defer PkeyClass till a different PR where we carefully evaluate the whole fromdata API and the pkey class idea, without mixing a feature change with an API change.

Jakuje: Updated.

simo5 (approved):
LGTM

nwalfield:
Thanks everyone for working on this!

PR #470, opened 2026-06-29
Allow shorter EC private key lengths

### LINKED ISSUE #468: Verify EC keys can be properly imported even if they miss leading zeroes
Potentially related to #430, which affected p521 EC keys, but it might sometimes hit other key types.

Seems like this is still broken after recent changes. It failed for me at least once with OpenSSL 4.0 in the integration tests with pkcs11-provider:
```
 openssl pkeyutl -sign
              -in ${TMPPDIR}/sha256.bin
Could not find or decode private key from /__w/kryoptic/kryoptic/pkcs11-provider/builddir/tests/kryoptic/file.ec.key.pem
00E3D9B2077F0000:error:1E08010C:DECODER routines:OSSL_DECODER_from_bio:unsupported:crypto/encode_decode/decoder_lib.c:104:No supported data to decode.  Input type: DER, Input structure: PrivateKeyInfo
pkeyutl: Error loading key
```
But it did not fail during rerun.

_Originally posted by @Jakuje in https://github.com/latchset/kryoptic/pull/466#discussion_r3477302844_

simo5:
#### Description

Update CKA_VALUE validation for ECDSA, EDDSA, and Montgomery private keys to accept lengths up to the expected curve size, rather than requiring an exact match. This prevents the rejection of valid keys that have minimal byte representations (where leading zeros are dropped).

Additionally, a new test script is included to verify the bulk generation and importation of EC keys into the token.

Fixes #468

NOTE: this is needed because upon re-reading the spec I realized that my initial interpretation that minimal integers was not allowed (ie you needed leading zeros to be explicitly provide) was wrong.

#### Checklist

- [ ] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/misc/ec_gen_test.sh
+++ b/misc/ec_gen_test.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+TOKDIR=/tmp/ec_gen_test
+SLOTID=42
+PINVALUE="S3cr37-4-m3-bu7-n07-4-7h33"
+
+rm -fr "${TOKDIR}"
+mkdir -p "${TOKDIR}"
+
+# Kryoptic configuration
+cat << EOF > "${TOKDIR}/kryoptic.conf"
+[[slots]]
+slot = ${SLOTID}
+dbtype = "sqlite"
+dbargs = "${TOKDIR}/kryoptic.sql"
+#mechanisms
+EOF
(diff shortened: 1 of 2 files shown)


Jakuje (approved):

Jakuje:
Thanks for investigation!. It might be worth adding some test input with shorter keys like this, but it should not block the release!

PR #476, opened 2026-07-06
Prevent test deadlocks in synchronization

### LINKED ISSUE #473: Another occasional hang of very parallel tests
While building Fedora package for 1.5.2, I got another hang, which looks similar to the previous deadlock we had with previous release (#459). The RPM builds run many parallel tests (`cargo test -j112`), which was one of the causes in the previous case too. The first hanging test (if I see right) is `tests::aes::test_aes_iv_generators`):
```
test tests::aes::test_aes_iv_generators has been running for over 60 seconds
...
```
https://koji.fedoraproject.org/koji/taskinfo?taskID=147224529

I did not investigate the issue deep enough, but it looks like something to fix eventually. Builds on other platforms than x84_64 and on Fedora 45 went through just ok.

simo5:
#### Description

Replaced the blocking `write()` lock acquisition on the SYNC primitive with a polling loop using `try_write()` and a short sleep. A queued write lock could block incoming `read()` attempts, leading to deadlocks during test finalization. Using `try_write()` ensures that pending reads can complete safely.

Fixes #473 

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -5,6 +5,8 @@ use std::ffi::CString;
 use std::fs::{create_dir_all, remove_dir_all, OpenOptions};
 use std::io::Write;
 use std::sync::Once;
+use std::thread;
+use std::time::Duration;
 
 #[cfg(feature = "log")]
 use crate::log;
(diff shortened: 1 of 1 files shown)


src/tests/mod.rs
```diff
@@ -198,8 +200,22 @@ impl TestToken<'_> {
...
         } else {
             self.sync = None;
+
             /* wait for all others to complete */
-            drop(SYNC.write().unwrap());
+            loop {
+                /* This needs to use try_write() because a queued write lock
+                 * may actually block read() attempts, which would cause
+                 * tests to deadlock, sleeps between attempts. */
+                thread::sleep(Duration::from_millis(5));
```
Jakuje, line 209: not a big deal, but should we wait after the try_write fails, rather than before to not delay each of these of calls? Not a big issue due to being one-off per test finalization though.

Jakuje (approved):
the good news is that its just in tests.

src/tests/mod.rs
simo5: makes no difference, this test will have to wait no matter what in 99.9% of the cases, so waiting 5 milliseconds before or after doesn't make a material difference, and it was cleaner to code it this way :)

PR #442, opened 2026-04-14
Add docs.rs CI workflow

simo5:
#### Description

Add a GitHub Actions workflow to automatically build project documentation using cargo docs-rs on pushes and pull requests.

This ensures the documentation builds correctly in an environment similar to docs.rs and uses the nightly toolchain with warnings denied to catch documentation issues early in CI.

#### Checklist

- [ ] Test suite updated
- [ ] Rustdoc string were added or updated
- [ ] CHANGELOG and/or other documentation added or updated
- [x] This is not a code change

#### Reviewer's checklist:

- [ ] Any issues marked for closing are fully addressed
- [ ] There is a test suite reasonably covering new functionality or modifications
- [ ] This feature/change has adequate documentation added
- [ ] A changelog entry is added if the change is significant
- [ ] Code conform to coding style that today cannot yet be enforced via the check style test
- [ ] Commits have short titles and sensible text
- [ ] Doc string are properly updated

PR #479, opened 2026-07-08
Fix building docs

Jakuje:
#### Description

This was accidentally removed in 0a96cb50e5079fe343aaa0b4d161a477a08f0ee7

#### Checklist
- [X] This is not a code change

#### Reviewer's checklist:

### DIFF
--- a/ossl/Cargo.toml
+++ b/ossl/Cargo.toml
@@ -39,3 +39,7 @@ dynamic = ["ossl-sys/dynamic"]
 fips = ["ossl400", "ossl-sys/fips"]
 log = ["dep:log", "dep:vsprintf"] # Error tracing using log crate
 rfc9580 = [] # Enables features required for OpenPGP implementations
+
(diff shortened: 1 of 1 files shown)


simo5:
Actually using openssl-sys may allow us to enable docs building for crates.io as well, so far it did not work because I did not have a way to guarantee openssl was available, but they have a specific exception for openssl-sys IIRC.
Do you want to tackle enabling that as well ?
See #442 for context

Jakuje:
I agree that would be good to have a generic solution, but this is just a bandaid and fixup of what I messed (and missed) in 0a96cb50e5079fe343aaa0b4d161a477a08f0ee7 where I removed this block as no longer needed by changing default to openssl-sys, but did not add it back when we reverted.

This is currently preventing the docs of latest release to build: https://docs.rs/crate/ossl/latest

simo5 (approved):
LGTM

PR #480, opened 2026-07-12
implementation of CKM_RSA_AES_KEY_WRAP mechanism

keldonin:
#### Description

This PR adds support for `CKM_RSA_AES_KEY_WRAP` wrapping mechanism, that combines in one go RSA wrapping of a temporary AES key, then AES wrapping of any extractable key.

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:


keldonin:
@simo5 I didn't realize there was a similar PR in the queue, although it seems to be hanging for a while now. I'm fine either way, provided that support for this mechanism is eventually merged into the code base.

Let me know of your pref,

Regards

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+}
+
+impl Mechanism for RsaAesKeyWrapMechanism {
+    fn info(&self) -> &CK_MECHANISM_INFO {
+        &self.info
+    }
+
+    /// Implements the RSA AES key wrap operation (Wrap)
+    ///
+    /// [RSA AES KEY WRAP](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203524)
```
Jakuje, line 960: I think the 3.2 is already out so I would reference the latest issue here:

https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc234407237

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+        }
+
+        /* the wrapping key must be an RSA public key */
+        if wrapping_key.get_attr_as_ulong(CKA_CLASS)? != CKO_PUBLIC_KEY
+            || wrapping_key.get_attr_as_ulong(CKA_KEY_TYPE)? != CKK_RSA
+        {
+            return Err(CKR_WRAPPING_KEY_TYPE_INCONSISTENT)?;
+        }
+
+        let params = mech.get_parameters::<CK_RSA_AES_KEY_WRAP_PARAMS>()?;
```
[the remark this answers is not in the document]
Jakuje: I think this is a fix for the `get_parameters()` function, rather than this place.

Jakuje:
The #448 did not get update for months so we would better get one active merged than none.

src/fips/indicators.rs
```diff
@@ -413,6 +413,21 @@ const FIPS_CHECKS: FipsChecks = FipsChecks {
...
+                restrict!(CKK_RSA, range!(2048, 16384)),
+                restrict!(),
+            ],
+            genflags: CKF_SIGN
+                | CKF_VERIFY
+                | CKF_ENCRYPT
+                | CKF_DECRYPT
+                | CKF_WRAP
+                | CKF_UNWRAP
+                | CKF_DERIVE,
```
Jakuje, line 429: We have the `CKF_DERIVE` on other wrapping mechanisms, but not on the output AES key itself, see the AES key definition above in FipsKeyType.
If I see right, the only AES Derive operation that exists in specs are `CKM_AES_CBC_ENCRYPT_DATA` and `CKM_AES_ECB_ENCRYPT_DATA`, which are not FIPS methods so I think the genflags should not contain the DERIVE. Otherwise the derve should be also on the AES key to be consistent. This is more to @simo5 to clarify FIPS considerations here.

src/tests/rsa.rs
```diff
@@ -798,3 +798,206 @@ fn test_rsa_public_key_info() {
...
+        &[],
+        &[(CKA_WRAP, true), (CKA_ENCRYPT, true)],
+        &[(CKA_CLASS, CKO_PRIVATE_KEY), (CKA_KEY_TYPE, CKK_RSA)],
+        &[],
+        &[
+            (CKA_PRIVATE, true),
+            (CKA_SENSITIVE, true),
+            (CKA_UNWRAP, true),
+            (CKA_DECRYPT, true),
+            (CKA_EXTRACTABLE, true),
```
Jakuje, line 825: does the wrapping key need to be extractable for this test?

src/tests/rsa.rs
```diff
@@ -798,3 +798,206 @@ fn test_rsa_public_key_info() {
...
+            (CKA_MODULUS, modulus.as_slice()),
+            (CKA_PUBLIC_EXPONENT, public_exponent.as_slice()),
+            (CKA_PRIVATE_EXPONENT, private_exponent.as_slice()),
+            (CKA_PRIME_1, prime_1.as_slice()),
+            (CKA_PRIME_2, prime_2.as_slice()),
+            (CKA_EXPONENT_1, exponent_1.as_slice()),
+            (CKA_EXPONENT_2, exponent_2.as_slice()),
+            (CKA_COEFFICIENT, coefficient.as_slice()),
+        ],
+        &[(CKA_UNWRAP, true), (CKA_DECRYPT, true)],
```
Jakuje, line 960: Is the explicit DECRYPT flag on the private key needed for the unwrapping operation? I guess this is due to the reuse of the decrypt operation, which does the flag check. I wonder if we should not bypass that in this case (and do the checks only externally to low-level encryption/decryption functions).

I know the usage could be limited with the ALLOWED_MECHANISMS, but I would consider limiting the usage also on the attributes level to either wrapping or encryption without the need for the other one.

src/misc.rs
```diff
@@ -39,6 +39,75 @@ macro_rules! void_ptr {
...
+mod weak_key_wrap_tests {
+    use super::is_weak_key_wrap;
+
+    #[test]
+    fn aes128_is_always_tolerated() {
+        /* AES-128 is the baseline and never flagged, even with a 112-bit
+         * (RSA-2048) or weaker wrapping key */
+        assert!(!is_weak_key_wrap(112, 128));
+        assert!(!is_weak_key_wrap(128, 128));
+        assert!(!is_weak_key_wrap(80, 128));
```
Jakuje, line 89: nit: could these be sorted by size?

src/rsa.rs
```diff
@@ -57,6 +60,18 @@ static RSA_MECHS: LazyLock<[Box<dyn Mechanism>; 4]> = LazyLock::new(|| {
...
+static RSA_AES_KW_MECH: LazyLock<Box<dyn Mechanism>> = LazyLock::new(|| {
+    Box::new(RsaAesKeyWrapMechanism {
+        info: CK_MECHANISM_INFO {
+            ulMinKeySize: CK_ULONG::try_from(MIN_RSA_SIZE_BITS).unwrap(),
+            ulMaxKeySize: CK_ULONG::try_from(MAX_RSA_SIZE_BITS).unwrap(),
+            flags: CKF_WRAP | CKF_UNWRAP,
+        },
+    })
+});
+
```
Jakuje, line 74: nit: why not include this mechanism into the `RSA_MECHS` table? I do not have a strong opinion either way, but we have all the others. Not sure if there is some performance difference in having one or more LazyLock, boxes.

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+        params: &CK_RSA_AES_KEY_WRAP_PARAMS,
+    ) -> Result<CK_MECHANISM> {
+        if params.pOAEPParams.is_null() {
+            return Err(CKR_MECHANISM_PARAM_INVALID)?;
+        }
+        Ok(CK_MECHANISM {
+            mechanism: CKM_RSA_PKCS_OAEP,
+            pParameter: params.pOAEPParams as CK_VOID_PTR,
+            ulParameterLen: CK_ULONG::try_from(std::mem::size_of::<
+                CK_RSA_PKCS_OAEP_PARAMS,
```
Jakuje, line 904: could we use the convenience macro `sizeof!` here from `misc.rs`?

```suggestion
            ulParameterLen: sizeof!(CK_RSA_PKCS_OAEP_PARAMS),
```

src/tests/rsa.rs
simo5: when CKA_UNWRAP is used CKA_DECRYPT should not be needed, ie a key lacking it should still work, so the code need to be adjusted if the test fails w/o CKA_DECRYPT.

src/rsa.rs
simo5: yes, please reuse the existing tables where possible

src/fips/indicators.rs
simo5: IIRC these flags for an unwrap operation basically determine wat flags are set on the unwrapped key. So they should be broad, or we could be denying operations w/o a good reason?
Note this is consistent with the defaults we have on AES Key (un)wrap.

src/misc.rs
```diff
@@ -39,6 +39,75 @@ macro_rules! void_ptr {
...
 
+/// Returns `true` if wrapping an AES key of `aes_key_bits` bits with a
+/// wrapping key that provides `wrap_strength_bits` of security strength should
+/// be flagged as weak.
+///
+/// AES-128 is always considered acceptable as a baseline (per NIST guidance a
+/// 112/128-bit wrapping key is adequate to protect an AES-128 key), so a
+/// wrapping is only considered weak when a stronger AES key (>= 192 bits) is
+/// wrapped by a wrapping key of lower security strength.
+pub fn is_weak_key_wrap(
```
simo5, line 50: Can we change the name of this function to `weak_wrapping_key` ?
I have a hard time wrapping (pun, lol) my head around the current name.

src/misc.rs
```diff
@@ -39,6 +39,75 @@ macro_rules! void_ptr {
...
+
+/// Emits a warning (when the `log` feature is enabled) if a key-wrapping
+/// operation protects an AES key with a wrapping key that provides less
+/// security strength than the AES key itself.
+///
+/// See [is_weak_key_wrap] for the exact condition.
+///
+/// `wrap_strength_bits` is the NIST SP800-57 security strength of the wrapping
+/// key, and `aes_key_bits` is the size in bits of the AES key being wrapped.
+pub fn warn_weak_key_wrap(wrap_strength_bits: usize, aes_key_bits: usize) {
```
simo5, line 65: should we emit a warning or just refuse the operation ?

src/misc.rs
```diff
@@ -39,6 +39,75 @@ macro_rules! void_ptr {
...
+             reduced to {} bits",
+            wrap_strength_bits,
+            aes_key_bits,
+            wrap_strength_bits
+        );
+    }
+}
+
+#[cfg(test)]
+mod weak_key_wrap_tests {
```
simo5, line 80: -> weak_wrapping_keys_tests

src/misc.rs
```diff
@@ -39,6 +39,75 @@ macro_rules! void_ptr {
...
+    fn aes128_is_always_tolerated() {
+        /* AES-128 is the baseline and never flagged, even with a 112-bit
+         * (RSA-2048) or weaker wrapping key */
+        assert!(!is_weak_key_wrap(112, 128));
+        assert!(!is_weak_key_wrap(128, 128));
+        assert!(!is_weak_key_wrap(80, 128));
+    }
+
+    #[test]
+    fn stronger_aes_than_wrapper_is_weak() {
```
simo5, line 93: -> wrapping_key_is_weaker_than_aes_key

src/misc.rs
```diff
@@ -39,6 +39,75 @@ macro_rules! void_ptr {
...
+        /* RSA-2048 (112 bits) wrapping AES-192/256 */
+        assert!(is_weak_key_wrap(112, 192));
+        assert!(is_weak_key_wrap(112, 256));
+        /* RSA-3072/4096 (128 bits) wrapping AES-192/256 */
+        assert!(is_weak_key_wrap(128, 192));
+        assert!(is_weak_key_wrap(128, 256));
+    }
+
+    #[test]
+    fn matching_or_stronger_wrapper_is_not_weak() {
```
simo5, line 103: -> wrapping_key_is_not_weak

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
             .import_from_wrapped(keydata, template)
     }
 }
+
+/// Returns the NIST SP800-57 security strength (in bits) provided by an RSA
+/// key of the given modulus size in bits.
+fn rsa_security_bits(modulus_bits: usize) -> usize {
```
simo5, line 803: I think this should implemented by exposing security bits from the ossl layer and use EVP_PKEY_get_security_bits() in there, not hardcoded in a table in here.

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+#[derive(Debug)]
+struct RsaAesKeyWrapMechanism {
+    info: CK_MECHANISM_INFO,
+}
+
+impl RsaAesKeyWrapMechanism {
+    /// Validates the requested temporary AES key size and returns its length
+    /// in bytes.
+    fn aes_key_bytes(aes_key_bits: CK_ULONG) -> Result<usize> {
+        match aes_key_bits {
```
simo5, line 884: I think this should just be:
```
    if aes_key_bits % 8 != 0 {
        return Err(CKR_MECHANISM_PARAM_INVALID)?
    }
    aes::check_key_len(aes_key_bits / 8).map_err(|_| CKR_MECHANISM_PARAM_INVALID)?
```

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+        CK_MECHANISM {
+            mechanism: CKM_AES_KEY_WRAP_KWP,
+            pParameter: std::ptr::null_mut(),
+            ulParameterLen: 0,
+        }
+    }
+
+    /// Constructs an in-memory, ephemeral AES key object from raw bytes to be
+    /// used solely for the internal AES-KWP operation.
+    fn ephemeral_aes_key(value: &[u8]) -> Result<Object> {
```
simo5, line 921: this should use AesKeyFactory::create()

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+    }
+
+    /// Generates an in-memory, ephemeral AES key object of `aes_len` bytes.
+    ///
+    /// This routes through [default_secret_key_generate], the same
+    /// CSPRNG-backed helper used to generate all other secret keys in the
+    /// token, so the temporary key benefits from the exact same generation
+    /// path. The object is kept in memory only, never stored, and is used
+    /// solely for the internal AES-KWP operation.
+    fn generate_ephemeral_aes_key(aes_len: usize) -> Result<Object> {
```
simo5, line 940: This should use either `AesMechanism::generate_key()`
or `AES_KEY_FACTORY.as_key_factory()?.key_generate(template)?` if the former does something we do not want.

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+    ///
+    /// [RSA AES KEY WRAP](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203524)
+    /// (Version 3.1)
+    fn unwrap_key(
+        &self,
+        mech: &CK_MECHANISM,
+        wrapping_key: &Object,
+        data: &[u8],
+        template: &[CK_ATTRIBUTE],
+        key_template: &Box<dyn ObjectFactory>,
```
simo5, line 1074: key_templae -> key_factory

Also I think this should be used instad of directly importing the AES module from ossl as importing directly tightly couples the implementation of AES to that of RSA, the proper abstractions hould be used instead so that RSA and AES implementations remain decoupled and different cryptography backends for AES and RSA can be mixed.

src/rsa.rs
simo5: Actually this function can simply be:
```
    if aes_key_bits % 8 != 0 {
        return Err(CKR_MECHANISM_PARAM_INVALID)?
    }
    aes_key_bits / 8
```

There is no need to do validation here, because when using the proper ObjectFactory::generate/import function the AES code will check the lenght is valid anyway.

simo5 (changes requested):
I think we need to streamline some stuff, but over all this is going in the right direction.

src/rsa.rs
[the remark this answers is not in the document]
keldonin: OK, leaving it as it is.

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+        }
+
+        /* the unwrapping key must be an RSA private key */
+        if wrapping_key.get_attr_as_ulong(CKA_CLASS)? != CKO_PRIVATE_KEY
+            || wrapping_key.get_attr_as_ulong(CKA_KEY_TYPE)? != CKK_RSA
+        {
+            return Err(CKR_UNWRAPPING_KEY_TYPE_INCONSISTENT)?;
+        }
+
+        let params = mech.get_parameters::<CK_RSA_AES_KEY_WRAP_PARAMS>()?;
```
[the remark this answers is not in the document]
keldonin: same as above. Left as it is, since it is suggested this should be fixed instead within `get_parameters()`.

src/rsa.rs
```diff
@@ -781,3 +797,327 @@ impl Mechanism for RsaPKCSMechanism {
...
+            usize::try_from(params.ulAESKeyBits)?,
+        );
+
+        /* export the target key material to be wrapped */
+        let mut keydata =
+            key_template.as_key_factory()?.export_for_wrapping(key)?;
+
+        /* AES-KWP output length (C2), per NIST SP800-38F: the input padded to
+         * an 8-byte boundary plus an 8-byte integrity check block */
+        let c2_len = ((keydata.len() + AES_BLOCK_SIZE - 1) / 8) * 8;
```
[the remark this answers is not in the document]
keldonin: AES is by design using 16 bytes blocks. Changing the block size means a different algorithm.
I have adjusted the code to use `(AES_BLOCK_SIZE>>1)` instead (in a further commit).

src/rsa.rs
keldonin: good catch, adjusting in a further commit.

src/fips/indicators.rs
keldonin: Given that other FIPS mechanisms allow it `CKM_AES_KEY_WRAP` and `CKM_AES_KEY_KWP` I would recommend to allow it.
Also, AFAICS, FIPS 800-108F makes no restriction about unwrapped key material. You may not support yet mechanisms that derive a key based on another key, but if so in the future, it might become a regression.

src/tests/rsa.rs
keldonin: Not at all. Same for decrypt and extractable. These are removed in a further commit.

src/tests/rsa.rs
keldonin: Indeed `CKA_DECRYPT` not needed here. Removed in a next commit.

src/misc.rs
keldonin: done

src/rsa.rs
keldonin: I'm fine with that. It's just that this is a combined mechanism, not just an RSA one.
Moved to the end of the `RSA_MECHS` table.

src/misc.rs
keldonin: I'm no native speaker, so I may not have a strong feeling about the original name.

I changed it to: `wrap_downgrades_security_strength`, and the warn has been changed to `warn_downgraded_security_for_wrapped_key` which is rather explicit IMO (even though a bit long). 
If you have counter proposals, I'm fine as well.

src/misc.rs
keldonin: That's a great question. I used to be very strict on this; I had to change my views over time. In reality, AFAIK the NIST standards do not forbid the pattern; simply, you know what you are doing. A warning is educative enough IMO.

src/misc.rs
keldonin: adjusted accordingly.

src/misc.rs
simo5: Well I proposed "weak_wrapping_key" in my comment, didn't I :-)
But I am ok with how your names work too, although I might suggest using "lowers" rather than "downgrades_security" to make the name a bit shorter.

IE `wrapping_lowers_strength` / `warn_on_lower_strength_wrapping_key`

src/misc.rs
keldonin: made adjustments too.

PR #483, opened 2026-07-30
Change license from GPLv3 to Apache 2.0

### LINKED ISSUE #474: Can the license be changed?
I've had quite a few folks ask me about this project and whether or not the license could be moved to something like Apache License 2.0 so it could be included in a range of commercial projects.

Is this something that is possible?

simo5:
#### Description

Replace the GNU General Public License v3.0 (GPLv3) with the Apache License 2.0 in the LICENSE.txt file. This changes the project to a more permissive open-source license.
This allows greater flexibility for Open Source projects that can't easily link/use GPL 3.0 code.

Note that most of the code used by Kryoptic is already ASL 2.0 licensed so this is not as big a change as it may look.

I have reviewed the list of contributors and privately contacted the authors with relevant amount of code not already covered with ASL 2.0 license (all code under the ossl/ directory is already ASL 2.0)

I will leave this PR open for a week an will merge it after 7 days if there is no objection from someone that have a valid copyright claim in the code. I am assuming silence-consent if no comment are made. 

Fixes #474

#### Checklist


Jakuje:
Please, change the licence also in the spec file, in README.md and in the Cargo.toml

The files in the ossl and ossl-sys crate still have the following header `// See LICENSE.txt file for terms`, while the file in `ossl` is called `LICENSE-2.0.txt` (which is a bit orthogonal issue). I think the file should be just renamed. 

Given that we publish the separate crates, I think each of them should have separate license file in the end.

Otherwise lgtm

simo5:
Great points!

> Please, change the licence also in the spec file, in README.md and in the Cargo.toml

ack

> The files in the ossl and ossl-sys crate still have the following header `// See LICENSE.txt file for terms`, while the file in `ossl` is called `LICENSE-2.0.txt` (which is a bit orthogonal issue). I think the file should be just renamed.
> 
> Given that we publish the separate crates, I think each of them should have separate license file in the end.


Yeah I will change the file name and place it at the root of each crate.

> Otherwise lgtm

Thanks for the feedback.

packaging/kryoptic.spec
```diff
@@ -22,14 +22,13 @@ Version:        1.5.2
...
 
-SourceLicense:  GPL-3.0-or-later
+SourceLicense:  Apache-2.0
 # LICENSE.dependencies contains a full license breakdown
 License: %{shrink:
  Apache-2.0 AND
  (Apache-2.0 OR BSL-1.0) AND
  (Apache-2.0 OR MIT) AND
  BSD-3-Clause AND
- GPL-3.0-or-later AND
```
Jakuje, line 32: Confirmed this matches the rpmbuild output:
```
### BEGIN LICENSE SUMMARY ###

# Apache-2.0
# Apache-2.0 OR MIT
# BSD-3-Clause
# ISC
# MIT
# MIT OR Apache-2.0
# MIT-0 OR Apache-2.0
# Unlicense OR MIT
# Zlib
###  END LICENSE SUMMARY  ###
```
in

https://github.com/latchset/kryoptic/actions/runs/30636532290/job/91175427839?pr=483

but it looks like we have one additional ` (Apache-2.0 OR BSL-1.0)` license identifier which should go away from the list

README.md
```diff
@@ -96,22 +96,12 @@ This command accepts the same feature set as the build command
...
-and will be open to making a change if we think that such a change would in fact
-be in the best interest of the project. Note that requests of this kind
-need to come with a well-reasoned rationale that shows benefits both for
-the requesting party and the upstream project.
+The license is currently set as the Apache Software License 2.0 the same license
+OpenSSL uses.
 
+In versions prior than and including 1.5.2 the license for the token code was 
+the GPLv3.0+ as released by the FSF, we changed it to make the project more
+easily reusable cross a wider set of Open Source communities.
```
Jakuje:
```suggestion
easily reusable across a wider set of Open Source communities.
```
?

Jakuje (commented):
some of the non-top level Cargo.toml still have explicit `license = "Apache-2.0"` instead of `license.workspace = true`, but that should not be an issue. But using the former adds a duplication, while the other keeps it single source of truth.

simo5:
Just rebased on main, no other changes

Jakuje:
The rustls/ossl directory now has 2 LICENSE files:
https://github.com/latchset/kryoptic/tree/main/rustls/ossl
not sure how it happened. I think we should remove the `LICENSE-2.0.txt` from there.

Same for ossl and ossl-sys after this change being merged -- there is already LICENSE-2.0.txt which should be replaced by the LICENSE.txt.

The comment last time was not addressed or commented on if it is something that bothers us:

> some of the non-top level Cargo.toml still have explicit license = "Apache-2.0" instead of license.workspace = true, but that should not be an issue. But using the former adds a duplication, while the other keeps it single source of truth.

I am also not sure if we should not keep one LICENSE.txt in the root of the repository (this is removed now), because the top-level crate kryroptic still needs one.

simo5:
@Jakuje hopefully fixed all the concerns

Jakuje (approved):

Jakuje:
Actually this was not resolved:

> The rustls/ossl directory now has 2 LICENSE files:
> https://github.com/latchset/kryoptic/tree/main/rustls/ossl
> not sure how it happened. I think we should remove the LICENSE-2.0.txt from there.

simo5:
A week has passed and no objections, so merging.

PR #485, opened 2026-08-25
Introduce new API to run in permissive FIPS mode

Jakuje:
#### Description

This resolves the following:
 * The applications using ossl bindings with `ossl::OsslContext::new_lib_ctx()` were not inheriting the FIPS mode of the global context and were not loading FIPS provider where they should
 * If application loaded a FIPS provider using `OsslContext::load_fips_provider()`, the FIPS mode was enforced and it could not work for the algorithms that are not yet certified (but considered crucial = PQC)

This is done by few changes:
~ * Changing the API `new_lib_ctx()` to load default configuration file, which inherently brings the fips provider in FIPS mode~
~ * Changing kryoptic to explicitly not do this with `new_lib_ctx_raw()` otherwise we can very easily deadlock if we load a kryoptic into pkcs11-provider, that would be enabled in the openssl configuration~
 * Introducing a new API to use fips provider in permissive mode. This is done by loading default provider and setting a propq to the context, which is being passed down to any algorithm fetching and further context creation.

#### Checklist

- [~] Test suite updated
- [~] CHANGELOG and/or other documentation added or updated

#### Reviewer's checklist:

### DIFF
--- a/ossl/src/lib.rs
+++ b/ossl/src/lib.rs
@@ -235,18 +235,28 @@ pub(crate) use trace_ossl;
 pub struct OsslContext {
     context: *mut OSSL_LIB_CTX,
     providers: Vec<*mut OSSL_PROVIDER>,
+    propq: Option<&'static CStr>,
     #[cfg(feature = "fips")]
     is_fips: bool,
 }
 
 static LEGACY_PROVIDER_NAME: &CStr = c"legacy";
+static DEFAULT_PROVIDER_NAME: &CStr = c"default";
 static FIPS_PROVIDER_NAME: &CStr = c"fips";
 
+static FIPS_PROVIDER_PERMISIVE_PROPQ: &CStr = c"?fips=yes";
+
 impl OsslContext {
+    /// Creates a new empty OpenSSL library context.
+    ///
+    /// For most of the applications, its recommented to load the default
+    /// configuration file after this call to make sure the application
+    /// behaves as expected using `load_default_configuration()`.
     pub fn new_lib_ctx() -> OsslContext {
         OsslContext {
             context: unsafe { OSSL_LIB_CTX_new() },
             providers: Vec::new(),
+            propq: None,
             #[cfg(feature = "fips")]
             is_fips: false,
         }
@@ -257,6 +267,7 @@ impl OsslContext {
         OsslContext {
             context: ctx,
             providers: Vec::new(),
+            propq: None,
             #[cfg(feature = "fips")]
             is_fips: false,
         }
@@ -267,6 +278,7 @@ impl OsslContext {
         OsslContext {
             context: ctx,
             providers: vec![prov as *mut OSSL_PROVIDER],
+            propq: None,
             is_fips: true,
         }
     }
@@ -306,21 +318,15 @@ impl OsslContext {
         self.load_configuration_file(None)
     }
 
-    pub fn load_legacy_provider(&mut self) -> Result<(), Error> {
+    fn load_provider(&mut self, name: &CStr) -> Result<(), Error> {
         if cfg!(feature = "fips") {
             return Err(Error::new(ErrorKind::OsslError));
         }
-
-        if unsafe {
-            OSSL_PROVIDER_available(self.ptr(), LEGACY_PROVIDER_NAME.as_ptr())
-        } == 1
-        {
+        if unsafe { OSSL_PROVIDER_available(self.ptr(), name.as_ptr()) } == 1 {
             return Ok(());
         }
 
-        let provider = unsafe {
-            OSSL_PROVIDER_load(self.ptr(), LEGACY_PROVIDER_NAME.as_ptr())
-        };
+        let provider = unsafe { OSSL_PROVIDER_load(self.ptr(), name.as_ptr()) };
         if provider.is_null() {
             Err(Error::new(ErrorKind::OsslError))
         } else {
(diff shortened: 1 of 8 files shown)


ossl/src/lib.rs
```diff
@@ -243,20 +246,33 @@ static LEGACY_PROVIDER_NAME: &CStr = c"legacy";
...
             #[cfg(feature = "fips")]
             is_fips: false,
         }
     }
 
+    /// Creates a new OpenSSL library context for general use.
+    ///
+    /// This involves common setup of the context including loading
+    /// of default configuration file.
+    pub fn new_lib_ctx() -> OsslContext {
```
simo5: I do not think we can change behavior this way, unfortunately new_lib_ctx() needs to remain the way it was.

But we can add a new function `new_lib_ctx_with_conf()` ? Or simply recommend that libraries explicitly call `ctx.load_default_configuration();` after obtaining a new context.

ossl/src/lib.rs
```diff
@@ -364,6 +387,13 @@ impl OsslContext {
     pub fn ptr(&self) -> *mut OSSL_LIB_CTX {
         self.context
     }
+
+    pub fn get_propq(&self) -> *const i8 {
```
simo5: this should return a *const c_char

simo5: I would also rename it to propq_ptr()

simo5 (changes requested):
The general approach look sok to me, except the change in semantics.

ossl/src/lib.rs
Jakuje: In that case I would rather go with the second approach as it will need a change of applications anyway. And add a documentation note.

ossl/src/lib.rs
Jakuje: Thanks! Resolved.

ossl/src/lib.rs
Jakuje: Changed and added documentation

simo5: I am not sure I understand the change you made.

I still think you should load the configuration and not force load the fips provider in your case, so what we have here, does not really allow that?

Unless your plan is that now the sequence should be:
```
let mut ctx = new_lib_ctx();
ctx.load_default_configuration();
if ctx.fips_is_enabled() {
    ctx.load_fips_provider_permissive();
}
```

Is this how you envision using it?

simo5: I am asking becasue you unconditionally try to load the fips provider now in load_fips_provider_permissive() ... so if you want permissive mode but only if fips is already enabled, then the above is the only option ...

simo5: I think a better API would be to have just set_permissive_fips() which does not try to load the fips provider, but only checks if it is available, and sets the propq only in that case.

Then the code above can be simplified this way:
```
let mut ctx = new_lib_ctx();
ctx.load_default_configuration();
ctx.set_permissive_fips();
```

This will work regardless of what's in the default config ?

ossl/src/lib.rs
```diff
@@ -348,6 +359,15 @@ impl OsslContext {
         }
     }
 
+    pub fn load_fips_provider_permissive(&mut self) -> Result<(), Error> {
+        /// Load FIPS provider into existing context, if available. Prefer
+        /// the FIPS provider for all the operations, but if the algorithm
+        /// is not available, fall back to default provider.
+        self.load_fips_provider()?;
+        self.propq = Some(CString::new("?fips=yes").unwrap());
```
simo5: Why a CString now?
Pretty sure we can define a static CStr and avoid unnecessary allocations here

ossl/src/lib.rs
Jakuje: Thank you for the clarification. This sounds probably better. I guess I misunderstood during our discussion. Will adjust.

ossl/src/lib.rs
Jakuje: Resolved.

ossl/src/non-openssl-sys-bindings.rs
```diff
@@ -22,6 +22,7 @@ pub const OSSL_PKEY_PARAM_MGF1_DIGEST: &[u8; 12] = b"mgf1-digest\0";
 pub const OSSL_ASYM_CIPHER_PARAM_OAEP_DIGEST: &[u8; 7] = b"digest\0";
 
 // Digest
+pub const OSSL_DIGEST_NAME_RIPEMD160: &[u8; 10] = b"RIPEMD160\0";
```
simo5: uh, where is this coming from ?

simo5 (approved):
Except for style issue and stray change to bindings LGTM

ossl/src/non-openssl-sys-bindings.rs
Jakuje: I can move it to separate PR, but without this I had issues to build ossl with openssl-sys bindings and non-default features (rfc9580). We do not cover this features combination anywhere in CI.

Jakuje:
> Except for style issue and stray change to bindings LGTM

I am still testing it and it still does not seem to work correctly. The ML-DSA-87-Ed447 signature verification still fails with this change + sequoia change from https://gitlab.com/sequoia-pgp/sequoia/-/merge_requests/1884

```
Header V6 ML-DSA-87+Ed448/SHA512 Signature, key ID 05707a62: BAD
```

ossl/src/non-openssl-sys-bindings.rs
simo5: what about adding a single build target with the feature?

Jakuje: Yeah, we should do that. But I was thinking if we should do it as a separate target building whole kryoptic with these features (they are not even exposed to the main crate) or have separate pipeline for the ossl.

simo5: A separate build job just for ossl with its own matrix

ossl/src/lib.rs
```diff
@@ -329,23 +337,42 @@ impl OsslContext {
...
+        /// In strict FIPS mode, the default property query rqeuires
+        /// all operations to go through the FIPS provider. But there
+        /// are some operations we might want to keep working even
+        /// though they are not available in the FIPS provider.
+        /// Due to the way how OpenSSL works, to make this working we
+        /// need to explicitly load the default provider and then
+        /// configure the property query to prefer fips provider, but
+        /// fall back to default one if the algorithm is not available.
+        ///
+        /// This is no-op when the FIPS provider is not loaded.
```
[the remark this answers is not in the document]
Jakuje: Too much python coding recently ...

ossl/src/lib.rs
```diff
@@ -329,23 +337,42 @@ impl OsslContext {
...
-        let provider = unsafe {
-            OSSL_PROVIDER_load(self.ptr(), FIPS_PROVIDER_NAME.as_ptr())
-        };
-        if provider.is_null() {
-            Err(Error::new(ErrorKind::OsslError))
-        } else {
-            self.providers.push(provider);
-            Ok(())
+            self.load_default_provider()?;
+            self.propq = Some(FIPS_PROVIDER_PERMISIVE_PROPQ);
```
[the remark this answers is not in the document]
Jakuje: This is slightly orthogonal to the issue we solve here, but I think it makes sense to solve it too. I think the `load_fips_provider()` API was only accidentally exposed with the `fips` feature, but I do not think it was ever meant to be working. Moved the check into the `load_provider()`.

Jakuje:
This is tested to work correctly with all algorithms except for ML-DSA. Reporeted as openssl/openssl#32613.

Reproducer:
 * set the system into FIPS mode
 * clone this branch and apply the following patch
 * run the tests: `cargo test --features=pqc tests::mldsa::test_mldsa_operations`
 * Observe generic failure:

```
thread 'tests::mldsa::test_mldsa_operations' (22584) panicked at src/tests/mldsa.rs:369:23:
Generic CK error, CKR_DEVICE_ERROR(48)
```

Patch:
```diff
diff --git a/src/ossl/common.rs b/src/ossl/common.rs
index ffe2b76..b3d190c 100644
--- a/src/ossl/common.rs
+++ b/src/ossl/common.rs
@@ -53,7 +53,12 @@ pub fn osslctx() -> &'static OsslContext {
 
     #[cfg(not(feature = "fips"))]
     {
-        OSSL_CONTEXT.get_or_init(|| ossl::OsslContext::new_lib_ctx())
+        OSSL_CONTEXT.get_or_init(|| {
+		let mut ctx = ossl::OsslContext::new_lib_ctx();
+		let _ = ctx.load_default_configuration().unwrap();
+		let _ = ctx.set_permissive_fips().unwrap();
+		ctx
+	})
     }
 }
 
```
(note, that this is running the kryoptic tests in the fips context, which is something that we do not expect to be working, but it sounds easier to reproduce than writing a new import test in the ossl).

simo5:
Sounds like the above does not reproduces on all openssl versions, so it may be a problem confined to specific (maybe even downstream) versions.

simo5:
Turns out this problem happens only with FIPS provider based on 3.0.7 because that FIPS provider had a partial SHAKE implementation only ...
Given that is a very special issue, I am going to merge this PR as is, and we'll probably have some clever way to deal with the specific issue in a separate PR later

PR #487, opened 2026-08-27
Introduce separate CI for ossl crate

Jakuje:
#### Description

Some of the ossl crate features combinations were not tested so this is adding the bits.

Limited to dynamic linking so static linking is tested enough for the fips module.

#### Checklist

- [ ] Test suite updated

#### Reviewer's checklist:

### DIFF
--- a/.github/workflows/ossl.yml
+++ b/.github/workflows/ossl.yml
@@ -1,0 +1,85 @@
+name: Build ossl
+
+on:
+  pull_request:
+    branches: ["main"]
+    paths:
+      - 'ossl/**'
+  push:
+    branches: ["main"]
+
+
+jobs:
+  build:
+    name: Build ossl
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [minimal, all]
+        linking: [ossl-sys/dynamic, openssl-sys]
+
+    container: quay.io/fedora/fedora:latest
+    steps:
+      - name: Get Date for DNF cache entry
+        id: get-date
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ossl.yml
```diff
@@ -0,0 +1,98 @@
...
+jobs:
+  build:
+    name: Build ossl
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [i686, minimal, all, rfc9580]
+        linking: [ossl-sys/dynamic, openssl-sys]
+        build: [debug, release]
```
simo5: I do not think we need both debug and release builds, we can save compute resources by doing just one or the other ... I wonder if the matrix could be randomized so that statistically all options are checked just not all at each run ...

simo5 (approved):
LGTM

.github/workflows/ossl.yml
simo5: actually given we already test most of ossl in the regular builds, would it make sense to test only rfc5980 here ?

Jakuje: Agreed. Will drop.
In the meantime, openssl build started failing. Great!

simo5: Honestly I'd go further and drop i686 and collapse the rfc feature, ending up with only minimal and all

Jakuje: At this moment its 8 jobs each finishing under 2 minutes. But I can reduce it further. Likely the all and none ones are the most important

Jakuje: Reduced to 4

PR #488, opened 2026-08-27
Support SHAKE digests and ML-DSA mu generator

### LINKED ISSUE #484: ossl: Support providing μ when signing with ML-DSA
OpenSSL has the [OSSL_SIGNATURE_PARAM_MU](https://docs.openssl.org/3.5/man7/EVP_SIGNATURE-ML-DSA/#ml-dsa-signature-parameters) signing parameter, which is very convenient when implementing some sort of signing service where μ can be calculated client-side and the things you're signing are many gigabytes.

Currently this is possible in ossl, but not quite as convenient as it could be since there's no definition for the parameter so you need to do some variation of:

```rust
let mut parameters = ossl::OsslParamBuilder::with_capacity(1);
parameters.add_owned_int(c"mu", 1)?;
```

It'd be nice if that was a parameter for `ossl::signature::mldsa_params` and it'd also be nice to provide a helper to calculate mu. It's not particularly difficult for the caller to handle, but I think a user would expect to find it if there's a toggle to provide it.

simo5:
#### Description

Add support for SHAKE-128 and SHAKE-256 digest algorithms, including configuring the XOF output length via set_xoflen.

Introduce MuGenerator and helper functions to compute and pass the exact mu value required for ML-DSA signatures according to FIPS 204, enabling one-shot ML-DSA signing and verification with pre-computed mu values. Tests are added to verify ML-DSA-44 test vectors.

Assisted-by: Gemini:Gemini 3.7 Flash

Fixes #484 

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:


ossl/src/tests/mldsa.rs
```diff
@@ -0,0 +1,353 @@
+// Copyright 2025 Simo Sorce
```
Jakuje:
```suggestion
// Copyright 2026 Simo Sorce
```

ossl/src/tests/mldsa.rs
```diff
@@ -0,0 +1,353 @@
...
+};
+use crate::tests::test_ossl_context;
+
+#[test]
+fn test_mldsa_with_mu() {
+    let libctx = test_ossl_context();
+
+    if !EvpPkey::available(libctx, EvpPkeyType::Mldsa44) {
+        return;
+    }
```
Jakuje: why this guard? What would be the case where this would not be available when we include this only if 
```
#[cfg(all(ossl_v350, not(feature = "fips")))]
```

ossl/src/tests/mldsa.rs
```diff
@@ -0,0 +1,353 @@
...
+    // Compute MU correctly
+    let mut mu = vec![0u8; 64];
+    let mut mu_gen = MuGenerator::new(libctx, &key, Some(&ctx))
+        .expect("Failed to create mu generator");
+    mu_gen.update(&msg[0..50]).unwrap();
+    mu_gen.finalize(Some(&msg[50..]), &mut mu).unwrap();
+
+    // Verify using MU
+    let verify_params =
+        mldsa_with_mu(Some(&ctx), false).expect("Failed to build params");
```
Jakuje, line 324: what about the mu signatures? They do not work and they do not need to be tested?

ossl/src/tests/mldsa.rs
simo5: I'll drop it, pedantic AI added it :)

ossl/src/tests/mldsa.rs
simo5: the OpenSSL interfaces are symmetrical here, if verification works signature will work too, but I can add a test signature too ...

simo5:
Fixed nits and added a sign with mu test that gets verified with the regular verification function to test cross compatibility

Jakuje (approved):

PR #489, opened 2026-09-01
Deal with slightly broken Shake on older FIPS modules

simo5:
#### Description

This patchset tries to handle dealing with slightly broken shake with 3.0.7 based fips modules.

Depends on https://github.com/openssl/openssl/pull/32636

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:


simo5:
Seem like loading the default configuration in static builds fails...

simo5:
In the static build we fail to load coniguration, probably because there is no file where openssl statically built code expects it, then again it makes no sense to try to load providers in a static build, so I simply marked all these new tests as to run only in dynamic builds.

Jakuje (approved):

PR #493, opened 2026-09-08
Fix CKM_AES_CCM one-shot C_EncryptMessage/C_DecryptMessage length probe

### LINKED ISSUE #491: C_EncryptMessage(pData=NULL) for CKM_AES_CCM reports pulCiphertextLen=0 instead of the required length
# C_EncryptMessage(pData=NULL) for CKM_AES_CCM reports pulCiphertextLen=0 instead of the required length

## Summary

Per PKCS#11 v3.0 §5.9.4 (`C_EncryptMessage`) and the general Cryptoki "get output length" idiom
(§5.2), calling `C_EncryptMessage` with `pCiphertext = NULL_PTR` must return `CKR_OK` and set
`*pulCiphertextLen` to the exact number of bytes the real call will produce, without performing
the encryption. For `CKM_AES_CCM`, Kryoptic instead returns `CKR_OK` with `*pulCiphertextLen`
left at `0` — the caller has no way to size its output buffer, and a caller that (correctly, per
spec) allocates a buffer of the reported size and calls again ends up passing a 0-length
buffer, which most FFI bindings (including .NET's P/Invoke marshalling) pass through as a null
pointer. The second call then also returns `CKR_OK`, silently producing empty ciphertext instead
of the real one.

## Versions

- Kryoptic: `v1.5.2` (tag `112ace3`), and confirmed still present on unreleased `main` as of commit
  `7200332` (2026-09-04).
- Built with: `cargo build --release -p kryoptic --features pqc`.
- OpenSSL: 3.5.7.
- OS: Linux x86_64.

## Reproduction

Captured directly at the FFI boundary from the KerckhoffsLabs.Security.Cryptography.Pkcs11 .NET
wrapper (which follows the standard two-call idiom: call once with a null output pointer to learn
the length, allocate a buffer of that length, call again). Instrumented the second argument
(`ctLen`, straight from the native out-parameter) immediately after the first call:

```
[DIAG] probe returned rv=CKR_OK ctLen=0 (plaintext was 18 bytes)
```

Independently, `KRYOPTIC_TRACE=stdout KRYOPTIC_TRACE_LEVEL=trace` shows both calls (the caller's
own null-then-real sequence) landing on Kryoptic's side with a null ciphertext pointer — the
second call is null too, because the 0-length buffer the caller allocated (per Kryoptic's own
reported length) pins to a null pointer:

```
C_EncryptMessage: s_handle=2 parameter=0x... parameter_len=56 associated_data=0x0
  associated_data_len=0 plaintext=0x... plaintext_len=18 ciphertext=0x0
  pul_ciphertext_len=0x...
C_EncryptMessage: ret=0
C_EncryptMessage: s_handle=2 parameter=0x... parameter_len=56 associated_data=0x0
  associated_data_len=0 plaintext=0x... plaintext_len=18 ciphertext=0x0
  pul_ciphertext_len=0x...
C_EncryptMessage: ret=0
```

The `parameter`/`parameter_len=56` is a `CK_CCM_PARAMS` (nonce, AAD pointer/len, tag bits/data,
data length) built for a 12-byte nonce, no AAD, 128-bit tag, and `ulDataLen = 18` matching the
plaintext. For CCM, ciphertext length equals plaintext length (the tag is carried separately in
the mechanism params on this call convention), so the expected reported length is 18, not 0.

A minimal from-scratch repro isn't possible through the `cryptoki` crate (the crate used by
Kryoptic's own `cdylib/tests/*.rs`): as of `cryptoki 0.12.0`, `cryptoki::mechanism::aead` only
implements `GcmParams`/`GcmMessageParams` — there is no `CcmParams` — so a caller wanting to
exercise `C_EncryptMessage` for CCM through that crate would need to hand-build a `CK_CCM_PARAMS`
via raw FFI (`cryptoki-sys`) directly against the loaded module. The evidence above was instead
captured from the actual failing caller (a mature, spec-following PKCS#11 client library that
successfully drives this exact idiom against SoftHSM2, opencryptoki, and NSS softoken).

## Suggested fix

In the `C_EncryptMessage` handler for `CKM_AES_CCM` (`kryoptic_lib::fns::encryption`), when
`pCiphertext` is `NULL_PTR`, compute and report the correct output length (equal to
`ulDataLen` from the `CK_CCM_PARAMS`, since CCM's tag is returned via the params structure on this
call convention, not appended to the ciphertext) instead of leaving `*pulCiphertextLen` at its
initial/default value. Worth checking whether the same length-probe path has the same gap for
`CKM_AES_GCM` and `CKM_CHACHA20_POLY1305` message-mode calls, since they likely share the
same dispatch code.

## Impact

Every `C_EncryptMessage` caller that follows the standard two-call length-discovery idiom (rather
than guessing/over-allocating a buffer up front) gets silent, non-erroring empty ciphertext for
`CKM_AES_CCM` instead of a working encryption or a clear error.

AlexLaroche:
## Summary

Fixes #491.

A one-step `C_EncryptMessage`/`C_DecryptMessage` call with a NULL output buffer (the standard
PKCS#11 "probe the required length" idiom) reported `0` for `CKM_AES_CCM` instead of the real
length, because `msg_encryption_len`/`msg_decryption_len` read `self.params.datalen` — which is
only populated once `msg_encrypt_begin`/`msg_decrypt_begin` has parsed the per-message
`CK_CCM_MESSAGE_PARAMS`, i.e. never, on the probe call of the one-shot entry point (only the
real, second call reaches that code path). A caller that (correctly) allocates a buffer sized to
the reported `0` and calls again ends up encrypting/decrypting into an empty buffer.

`MsgEncryption::msg_encryption_len`'s own doc comment already specifies that `final=true` means
"the next operation is either msg_encrypt or msg_encrypt_final" — i.e. the one-shot call is
supposed to be treated as final. `encrypt_message()`/`decrypt_message()` were instead hardcoding
`false` for it.

- Pass `final=true` (not `false`) from the one-shot `C_EncryptMessage`/`C_DecryptMessage` handlers.
- Make CCM's `msg_encryption_len`/`msg_decryption_len` report `self.buffer.len() + data_len` when
  `final` is true (matching what the real call will actually write, whether or not
  `msg_*_begin` has already run) and `0` otherwise (intermediate chunks are only buffered).
- Add a regression test exercising the exact reported scenario: probe, then the real one-shot
  encrypt call, then the same for decrypt, checking the round-trip.

## Test plan

- [x] New regression test `test_aes_ccm_message_one_shot_length_probe` confirmed to fail
      (reports length 0) against the unfixed code, and to pass with the fix.
- [x] `cargo test --release --features pqc -p kryoptic-lib --lib` — all 122 pre-existing unit
      tests plus the new one pass.
- [x] `cargo test --release --features pqc,integration_tests -p kryoptic` — all cdylib
      integration tests (`rc_ecdh`, `rc_eddsa_compat`, `rc_login`, `rc_re_initialize`,
      `rust_cryptoki`) still pass.
- [x] `make check-format` clean.

### DIFF
--- a/src/fns/encryption.rs
+++ b/src/fns/encryption.rs
@@ -668,8 +668,14 @@ fn encrypt_message(
     }
 
     if ciphertext.is_null() {
+        // A one-step C_EncryptMessage call is, by the msg_encryption_len
+        // contract (see MsgEncryption::msg_encryption_len), equivalent to
+        // a multi-part finalization -- pass final=true so mechanisms whose
+        // output length depends on it (e.g. CKM_AES_CCM, which buffers and
+        // writes nothing until the whole message is available) report the
+        // length the subsequent real call will actually produce.
         let retlen =
-            CK_ULONG::try_from(operation.msg_encryption_len(plen, false)?)
+            CK_ULONG::try_from(operation.msg_encryption_len(plen, true)?)
                 .map_err(|_| CKR_GENERAL_ERROR)?;
         unsafe {
             *pul_ciphertext_len = retlen;
@@ -1062,8 +1068,11 @@ fn decrypt_message(
     }
 
     if plaintext.is_null() {
+        // See the identical comment in encrypt_message: a one-step
+        // C_DecryptMessage call is equivalent to a multi-part
+        // finalization for the msg_decryption_len contract.
         let retlen =
-            CK_ULONG::try_from(operation.msg_decryption_len(clen, false)?)
+            CK_ULONG::try_from(operation.msg_decryption_len(clen, true)?)
                 .map_err(|_| CKR_GENERAL_ERROR)?;
         unsafe {
             *pul_plaintext_len = retlen;
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -2189,13 +2189,31 @@ impl MsgEncryption for AesOperation {
     fn msg_encryption_len(
         &mut self,
         data_len: usize,
-        _fin: bool,
+        fin: bool,
     ) -> Result<usize> {
         if self.finalized {
             return Err(CKR_OPERATION_NOT_INITIALIZED)?;
         }
         match self.mech {
-            CKM_AES_CCM => Ok(self.params.datalen),
+            CKM_AES_CCM => {
+                if fin {
+                    /* CCM is one-shot in OpenSSL: nothing is written to the
+                     * cipher buffer until the whole message (whatever was
+                     * already buffered by prior msg_encrypt_next calls, plus
+                     * this final chunk) has been accumulated. Using
+                     * self.params.datalen here would be wrong (and, for the
+                     * one-step msg_encrypt entry point, simply unset) before
+                     * the per-message CK_CCM_MESSAGE_PARAMS have been parsed
+                     * by msg_encrypt_begin/msg_encrypt_new -- buffer.len() +
+                     * data_len is always the correct total regardless of
+                     * whether that has happened yet. */
+                    Ok(self.buffer.len() + data_len)
+                } else {
+                    /* Intermediate (non-final) chunks are only buffered;
+                     * msg_encrypt_next never writes any output for CCM. */
+                    Ok(0)
+                }
+            }
             CKM_AES_GCM => Ok(data_len),
             _ => Err(self.op_err(CKR_GENERAL_ERROR)),
         }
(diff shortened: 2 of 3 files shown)


simo5:
Please squash commit

AlexLaroche:
Squashed into a single commit (883b780).

simo5 (approved):

PR #495, opened 2026-09-08
Report the real required length on CKR_BUFFER_TOO_SMALL

### LINKED ISSUE #494: CKR_BUFFER_TOO_SMALL never reports the actual required buffer length
# CKR_BUFFER_TOO_SMALL never reports the actual required buffer length

## Summary

PKCS#11 v3.2 §5.2 requires that when a function returns `CKR_BUFFER_TOO_SMALL` because
an output buffer is too small, it must set the output-length parameter to the actual
number of bytes required, so the caller can allocate a correctly-sized buffer and retry.
Kryoptic's `CK_RV` conversion layer discards this value everywhere: `Error::buf_too_small(n)`
carries the required size in its `reqsize` field, but nothing between there and the `extern "C"`
boundary ever reads `reqsize()` and writes it back into the caller's `*pulXxxLen`. The pointer
is simply left holding whatever the caller originally passed in.

A caller that follows the standard "try a reasonably-sized buffer, and on
`CKR_BUFFER_TOO_SMALL` retry with the size the token just told you" idiom gets stuck: the
"size the token just told you" is wrong (it's an echo of what it rejected), so the retry fails
identically, forever, with no way to discover the real size short of guessing ever-larger
buffers or falling back to the separate null-buffer probe idiom instead.

## Root cause (from the code)

`src/fns/encryption.rs`'s `fn_encrypt` (and every other `extern "C" fn_*` wrapper across the
crate — `fn_decrypt`, `fn_sign`, `fn_encrypt_message`, etc. all follow the same pattern):

```rust
let rv = match encrypt(s_handle, pdata, data_len, encrypted_data, pul_encrypted_data_len) {
    Ok(()) => CKR_OK,
    Err(e) => e.rv(),
};
```

`e.rv()` returns only the `CK_RV` code; `e.reqsize()` (the field `Error::buf_too_small`
specifically exists to carry — see `src/error.rs`) is never consulted, so `*pul_encrypted_data_len`
is never updated on this path. `rg -n "reqsize" src/fns/` returns zero matches anywhere in the
crate's C-ABI dispatch layer.

This isn't RSA-specific or encrypt-specific — `Error::buf_too_small(...)` is constructed at 11
call sites across both mechanism families, on both directions:

```
src/ossl/rsa.rs:531   (RSA decrypt)
src/ossl/rsa.rs:591   (RSA encrypt)
src/ossl/aes.rs:712, 1259, 1388, 1429, 1651, 2079, 2129, 2287, 2339
                      (AES encrypt/decrypt, plain and message-mode)
```

Every one of them is reachable from an `extern "C"` entry point that discards `reqsize()`
identically, so the same symptom (reported length == whatever was passed in, not the real
required size) should reproduce for AES and for decrypt as well as encrypt — I've only
independently confirmed the RSA `C_Encrypt` case below, but the fix belongs in the shared
conversion path, not in RSA specifically.

## Why existing tests miss this

Every existing RSA/AES encrypt test in `src/tests/` either pre-allocates an already-large-enough
buffer up front, or — like the shared `tests::util::encrypt` helper — probes the length with a
`NULL` output buffer first (a *different*, working code path: `encryption_len()` returns the
correct size via `Ok(...)`, no `reqsize()` involved). Neither exercises a deliberately-too-small
non-`NULL` buffer, so the `CKR_BUFFER_TOO_SMALL` retry path — and its broken length report — has
no coverage anywhere in the suite.

## Reproduction

Added as `src/tests/rsa.rs::test_rsa_encrypt_buffer_too_small_reports_required_len` (run with
`cargo test --release --features pqc -p kryoptic-lib --lib test_rsa_encrypt_buffer_too_small_reports_required_len -- --nocapture`):

```rust
#[test]
#[parallel]
fn test_rsa_encrypt_buffer_too_small_reports_required_len() {
    let mut testtokn = TestToken::initialized(
        "test_rsa_encrypt_buffer_too_small_reports_required_len",
        None,
    );
    let session = testtokn.get_session(true);
    testtokn.login();

    let (hpub, _hpri) = ret_or_panic!(generate_key_pair(
        session,
        CKM_RSA_PKCS_KEY_PAIR_GEN,
        &[(CKA_MODULUS_BITS, 2048)],
        &[],
        &[(CKA_TOKEN, false), (CKA_ENCRYPT, true)],
        &[(CKA_CLASS, CKO_PRIVATE_KEY), (CKA_KEY_TYPE, CKK_RSA)],
        &[],
        &[(CKA_TOKEN, false), (CKA_SENSITIVE, true), (CKA_DECRYPT, true)],
    ));

    let mut mechanism: CK_MECHANISM = CK_MECHANISM {
        mechanism: CKM_RSA_PKCS,
        pParameter: std::ptr::null_mut(),
        ulParameterLen: 0,
    };
    let ret = fn_encrypt_init(session, &mut mechanism, hpub);
    assert_eq!(ret, CKR_OK);

    // Deliberately too small: a 2048-bit key's RSA_PKCS ciphertext is exactly 256 bytes.
    let data = "plaintext";
    let mut enc: [u8; 8] = [0; 8];
    let mut enc_len: CK_ULONG = enc.len() as CK_ULONG;
    let ret = fn_encrypt(
        session,
        data.as_ptr() as *mut u8,
        data.len() as CK_ULONG,
        enc.as_mut_ptr(),
        &mut enc_len,
    );
    assert_eq!(ret, CKR_BUFFER_TOO_SMALL);
    assert_eq!(enc_len, 256); // FAILS: enc_len comes back as 8, the buffer we passed in
    testtokn.finalize();
}
```

Actual result:

```
thread 'tests::rsa::test_rsa_encrypt_buffer_too_small_reports_required_len' panicked:
assertion `left == right` failed
  left: 8
 right: 256
```

I also traced this independently through a real-world PKCS#11 client (the
KerckhoffsLabs.Security.Cryptography.Pkcs11 .NET wrapper), which follows exactly this
"pre-size, then retry on CKR_BUFFER_TOO_SMALL with the reported length" idiom for `C_Encrypt`
(to avoid a full-operation null-buffer probe on every call). Its first call passes a 33-byte
buffer (17-byte OAEP plaintext + 16 bytes of slack) and gets `CKR_BUFFER_TOO_SMALL` back with
the reported length also 33; retrying with a 33-byte buffer fails identically, forever — it
never learns that a 2048-bit key actually needs 256 bytes.

## Suggested fix

In each `extern "C" fn_*` wrapper's `Result<()>` → `CK_RV` conversion (or, better, in one shared
helper all of them go through), when `e.kind() == ErrorKind::BufferTooSmall`, write `e.reqsize()`
back into the relevant output-length pointer before returning `CKR_BUFFER_TOO_SMALL`. Given how
many `extern "C" fn_*` functions duplicate the same `match ... { Ok(()) => CKR_OK, Err(e) => e.rv() }`
shape, a small macro or shared helper that both wrappers can call — taking the output-length
pointer and the `Result` — would fix every call site (encrypt, decrypt, message-mode variants,
and any future one) at once rather than needing an audit-and-patch per function.

## Impact

Any caller using the standard PKCS#11 "try a buffer, retry on CKR_BUFFER_TOO_SMALL with the
reported size" idiom for RSA or AES encrypt/decrypt gets stuck in a loop it cannot break out of
except by falling back to the separate null-buffer-probe idiom (which does work, since it's a
different, `Ok`-returning code path). This is a real behavioral trap for any client library that
picked the (spec-sanctioned, and often more efficient — it avoids running some tokens' AEAD
operations twice) size-then-retry approach instead of always probing with `NULL` first.

AlexLaroche:
## Summary

Fixes #494.

PKCS#11 v3.2 §5.2 requires a function returning `CKR_BUFFER_TOO_SMALL` to write the actual
required size into the output-length parameter, so a caller can retry with a correctly-sized
buffer. A caller following that standard "try a buffer, retry on `CKR_BUFFER_TOO_SMALL` with the
reported size" idiom instead got stuck retrying with the same wrong size forever, across 18 call
sites, via two related but distinct instances of the same defect:

- `Error::buf_too_small(n)` carries the required size in its `reqsize` field, but every
  `extern "C" fn_*` wrapper's `Err(e) => e.rv()` conversion discarded it. Affected: `C_Encrypt`/
  `C_Decrypt`, their `Update`/`Final` streaming variants, `C_Encrypt(Decrypt)Message(Next)`
  message-mode variants, and `C_WrapKey` (11 sites spanning RSA and AES) — plus
  `RsaPKCSOperation::decrypt`, which used a bare `Err(CKR_BUFFER_TOO_SMALL)?` (via `Error::ck_rv`,
  which always zeroes `reqsize`) instead of `Error::buf_too_small`, so it needed its own fix even
  after the shared helper existed.
- A grep for that same bare `Err(CKR_BUFFER_TOO_SMALL)?` idiom outside the encrypt/decrypt/wrap
  paths found 7 more instances of the identical defect — a real, already-computed required size
  sits right there in scope and is silently dropped: `C_Sign`/`C_SignFinal`, `C_GetSlotList`/
  `C_GetMechanismList`, `C_GetInterfaceList`, `C_EncapsulateKey`/ML-KEM, and HOTP's
  `Sign::sign_final` (unreachable via the C ABI in practice, fixed anyway for consistency). Two
  sites with the same surface pattern — `C_Digest`/`C_DigestFinal` — were already correct and
  needed no change.

Left alone: `ossl/mlkem.rs`'s own `ErrorKind::BufferSize` handling inside `encapsulate()` can't be
fixed the same way — the underlying `ossl::Error` type carries no size field at all, so there's
nothing to propagate without extending the `ossl` crate itself. Its only caller
(`fns/keymgmt.rs::encapsulate_key`, fixed here) already checks the size before calling in, so this
looks unreachable today.

## Fix

- A shared `report_reqsize()` helper (`fns/mod.rs`, following the same crate-wide-visible pattern
  as the existing `fail_if_cka_token_true`) that writes `reqsize()` back into the output-length
  pointer when the error is `BufferTooSmall`, applied at the 11 sites sharing that shape.
- The remaining sites (RSA decrypt, `C_Sign`/enumeration/HOTP) each get the size written back
  directly at their own call site, matching what `C_Digest` already did correctly.

`C_Sign` and `C_GetSlotList`/`C_GetMechanismList` in particular had no test coverage anywhere
despite being among the most commonly used PKCS#11 functions: the standard caller pattern is
either to probe with a `NULL` buffer first (a different, already-correct branch) or supply an
already-large-enough buffer, so the too-small-non-`NULL`-buffer path went completely unexercised.

## Test plan

- [x] 13 new regression tests, one per distinct call-site shape (not one per algorithm — the bug
      and fix are in the mechanism-agnostic wrapper layer): RSA encrypt/decrypt one-shot, AES
      streaming `Update`/`Final`, AES-GCM message-mode one-shot/`Next`, `C_WrapKey`,
      `C_Sign`/`C_SignFinal`, `C_GetSlotList`/`C_GetMechanismList`/`C_GetInterfaceList`, and
      `C_EncapsulateKey` (ML-KEM). Each confirms the exact real-world required size (e.g. 256
      bytes for a 2048-bit RSA operation, 768 for ML-KEM-512's ciphertext).
- [x] All 136 unit tests (`cargo test --features pqc -p kryoptic-lib --lib`) pass.
- [x] All cdylib integration tests (`cargo test --features pqc,integration_tests -p kryoptic`)
      pass.
- [x] Cross-checked against a real-world caller (the KerckhoffsLabs.Security.Cryptography.Pkcs11
      .NET wrapper, which uses exactly the size-then-retry idiom for encrypt/decrypt): all 6 RSA
      OAEP/PKCS1 round-trip failures this caused are fixed, with no regressions.

### DIFF
--- a/src/tests/rsa.rs
+++ b/src/tests/rsa.rs
@@ -798,3 +798,185 @@ fn test_rsa_public_key_info() {
 
     testtokn.finalize();
 }
+
+/// Regression test: C_Encrypt with a too-small output buffer must return
+/// CKR_BUFFER_TOO_SMALL *and* report the real required length in
+/// pulEncryptedDataLen (PKCS#11 v3.2 5.2), so a caller can retry with a
+/// correctly-sized buffer. Every existing RSA encrypt test either
+/// pre-allocates an already-large-enough buffer or probes with a NULL
+/// buffer first (see tests::util::encrypt) -- neither exercises this path,
+/// which is why this went unnoticed: on a too-small non-NULL buffer,
+/// encrypt() in ossl/rsa.rs returns Err(Error::buf_too_small(outlen)), but
+/// the CK_RV conversion in fns/encryption.rs (`Err(e) => e.rv()`) discarded
+/// e.reqsize() and never wrote it back into *pul_encrypted_data_len --
+/// which was left holding whatever the caller originally passed in.
+#[test]
+#[parallel]
+fn test_rsa_encrypt_buffer_too_small_reports_required_len() {
+    let mut testtokn = TestToken::initialized(
+        "test_rsa_encrypt_buffer_too_small_reports_required_len",
+        None,
+    );
+    let session = testtokn.get_session(true);
+    testtokn.login();
+
+    let (hpub, _hpri) = ret_or_panic!(generate_key_pair(
+        session,
+        CKM_RSA_PKCS_KEY_PAIR_GEN,
+        &[(CKA_MODULUS_BITS, 2048)],
+        &[],
+        &[(CKA_TOKEN, false), (CKA_ENCRYPT, true),],
+        &[(CKA_CLASS, CKO_PRIVATE_KEY), (CKA_KEY_TYPE, CKK_RSA),],
+        &[],
+        &[
+            (CKA_TOKEN, false),
+            (CKA_SENSITIVE, true),
+            (CKA_DECRYPT, true),
+        ],
+    ));
+
+    /* OAEP, not classic RSA_PKCS v1.5 encryption: raw PKCS#1v1.5 encryption
+     * is denied in FIPS-mode builds, and this test must pass there too. */
+    let params = CK_RSA_PKCS_OAEP_PARAMS {
+        hashAlg: CKM_SHA256,
+        mgf: CKG_MGF1_SHA256,
+        source: CKZ_DATA_SPECIFIED,
+        pSourceData: std::ptr::null_mut(),
+        ulSourceDataLen: 0,
+    };
+    let mut mechanism: CK_MECHANISM = CK_MECHANISM {
+        mechanism: CKM_RSA_PKCS_OAEP,
+        pParameter: &params as *const _ as CK_VOID_PTR,
+        ulParameterLen: sizeof!(CK_RSA_PKCS_OAEP_PARAMS),
+    };
+    let ret = fn_encrypt_init(session, &mut mechanism, hpub);
+    assert_eq!(ret, CKR_OK);
+
+    /* Deliberately too small: a 2048-bit key's RSA_PKCS_OAEP ciphertext is
+     * exactly 256 bytes. */
+    let data = "plaintext";
+    let mut enc: [u8; 8] = [0; 8];
+    let mut enc_len: CK_ULONG = enc.len() as CK_ULONG;
+    let ret = fn_encrypt(
+        session,
+        data.as_ptr() as *mut u8,
+        data.len() as CK_ULONG,
+        enc.as_mut_ptr(),
+        &mut enc_len,
+    );
+    assert_eq!(ret, CKR_BUFFER_TOO_SMALL);
+    assert_eq!(
+        enc_len, 256,
+        "CKR_BUFFER_TOO_SMALL must report the real required length (256 \
+         for a 2048-bit RSA_PKCS_OAEP ciphertext), not leave \
+         *pulEncryptedDataLen at whatever the caller originally passed in (8)"
+    );
+
+    /* Retry with a correctly-sized buffer, exactly as a real caller would;
+     * confirm the operation is still active and produces real output. */
+    let mut enc2 = vec![0u8; enc_len as usize];
+    let mut enc2_len = enc2.len() as CK_ULONG;
+    let ret = fn_encrypt(
+        session,
+        data.as_ptr() as *mut u8,
+        data.len() as CK_ULONG,
+        enc2.as_mut_ptr(),
+        &mut enc2_len,
+    );
+    assert_eq!(ret, CKR_OK);
+    assert_eq!(enc2_len, 256);
+
+    testtokn.finalize();
+}
+
+/// Same as test_rsa_encrypt_buffer_too_small_reports_required_len, but for the mirror
+/// C_Decrypt call site (a distinct fn_encrypt/fn_decrypt pair in fns/encryption.rs, so a
+/// distinct place the fix could have been missed).
+#[test]
+#[parallel]
+fn test_rsa_decrypt_buffer_too_small_reports_required_len() {
+    let mut testtokn = TestToken::initialized(
+        "test_rsa_decrypt_buffer_too_small_reports_required_len",
+        None,
+    );
+    let session = testtokn.get_session(true);
+    testtokn.login();
+
+    let (hpub, hpri) = ret_or_panic!(generate_key_pair(
+        session,
+        CKM_RSA_PKCS_KEY_PAIR_GEN,
+        &[(CKA_MODULUS_BITS, 2048)],
+        &[],
+        &[(CKA_TOKEN, false), (CKA_ENCRYPT, true),],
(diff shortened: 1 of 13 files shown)


src/tests/rsa.rs
```diff
@@ -798,3 +798,167 @@ fn test_rsa_public_key_info() {
...
+        &[],
+        &[
+            (CKA_TOKEN, false),
+            (CKA_SENSITIVE, true),
+            (CKA_DECRYPT, true),
+        ],
+    ));
+
+    let mechanism: CK_MECHANISM = CK_MECHANISM {
+        mechanism: CKM_RSA_PKCS,
```
simo5: Please use OAEP as classic PKCS#1.5 encryption is denied in FIPS mode in our builds

simo5 (changes requested):
Overall good, except the issue with FIPS builds.

src/tests/rsa.rs
AlexLaroche: Good catch, switched both new tests (encrypt and decrypt) to CKM_RSA_PKCS_OAEP. Verified locally the fix passes; pushed as 9f24f3b.

simo5 (approved):

simo5:
Great finding and contribution, thanks!

PR #497, opened 2026-09-09
Fix AES-CCM shift-overflow that rejects data over 1 byte at nonceLen=7

### LINKED ISSUE #496: AES-CCM message mode: 7-byte nonce rejects any message longer than 1 byte (shift overflow in L capacity check)
## Summary

`C_EncryptMessage`/`C_DecryptMessage` (and the streaming `*MessageNext` calls) with `CKM_AES_CCM` reject any message longer than 1 byte when the nonce is exactly 7 bytes long, returning `CKR_MECHANISM_PARAM_INVALID`. A 7-byte nonce is a valid, spec-legal CCM configuration and should be the *least* restrictive nonce length, not the most.

## Where

`src/ossl/aes.rs`, `init_msg_params()`:

```rust
if params.ulNonceLen < 7 || params.ulNonceLen > 13 {
    return Err(CKR_MECHANISM_PARAM_INVALID)?;
}
let l = 15 - params.ulNonceLen;
if params.ulDataLen == 0
    || params.ulDataLen > (1 << (8 * l))
    || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
{
    return Err(CKR_MECHANISM_PARAM_INVALID)?;
}
```

## Root cause

Per NIST SP 800-38C, CCM's nonce length `N` and length-field width `L` satisfy `L = 15 - N`, with `N` valid over `[7, 13]` (so `L` ranges over `[8, 2]`). `L` bytes encode the message length, so the max representable message length is `2^(8*L)` bytes.

`CK_ULONG` is `std::os::raw::c_ulong` — 64 bits wide on Linux x86_64. For `nonceLen = 7`, `L = 8`, so the shift amount `8 * l` is **64**. Shifting a 64-bit integer left by 64 is out of range (valid amounts are `0..=63`):

- In a debug build (`overflow-checks = true`), this panics: "attempt to shift left with overflow."
- In a release build (`overflow-checks = false`, the default `cargo build --release` profile), it doesn't panic — LLVM lowers `<<` to the native x86-64 `shl` instruction, which masks the shift count modulo the register width, so `1u64 << 64` evaluates to `1u64 << (64 & 63)` = `1u64 << 0` = **1**.

So for `nonceLen = 7` the capacity check silently collapses from "reject messages over ~2^64 bytes" to "reject messages over **1 byte**" — the exact opposite of intent, since `L=8` is meant to be the *most* permissive case.

## Reproduction

Any `CKM_AES_CCM` message-mode encrypt/decrypt with `ulNonceLen == 7` and `ulDataLen > 1` fails with `CKR_MECHANISM_PARAM_INVALID`. Concretely, in a `C_EncryptMessageInit`/`C_EncryptMessage` round trip with a 7-byte nonce, 16-byte tag, and a 40-byte plaintext:

```
C_EncryptMessage returned CKR_MECHANISM_PARAM_INVALID
```

The same call with `ulNonceLen` 12 or 13 (same plaintext) succeeds, since `8*L` stays well under 64 for those (`L=3` and `L=2` respectively) and the shift never overflows.

## Suggested fix

Use `checked_shl`, whose contract is exactly "returns `None` if the shift amount is ≥ the bit width," instead of relying on the CPU's shift-masking behavior:

```rust
let l = 15 - params.ulNonceLen;
let max_len = 1u64.checked_shl((8 * l) as u32).unwrap_or(CK_ULONG::MAX);
if params.ulDataLen == 0
    || params.ulDataLen > max_len
    || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
{
    return Err(CKR_MECHANISM_PARAM_INVALID)?;
}
```

This preserves the intended cap for `nonceLen` 8–13 (`L < 8`) and correctly treats `nonceLen = 7` (`L = 8`) as "no meaningful cap short of `CK_ULONG::MAX`."

## Environment

Found while testing the Kryoptic backend against the latest `main` (`de40e034fdaf328544ee535bd94453ea58476af5`) from a C# PKCS#11 wrapper's test suite; confirmed by inspection this is present at `v1.5.2` too (the surrounding code is unchanged across that range) — not a regression, a pre-existing bug.

AlexLaroche:
## Summary

Fixes #496.

NIST SP 800-38C's length field `L = 15 - nonceLen` ranges over `[2, 8]` for the valid nonce lengths `[13, 7]`. Both `AesOperation::init_params` (the classic `CK_CCM_PARAMS` path, `C_EncryptInit`/`C_DecryptInit`) and `init_msg_params` (the message-mode `CK_CCM_MESSAGE_PARAMS` path, `C_EncryptMessage`/`C_DecryptMessage`) computed the resulting data-length cap as a bare `1 << (8 * l)`.

`CK_ULONG`'s width varies by platform (32 bits on i686, 64 bits on x86_64/aarch64/...). For `nonceLen == 7` (`l == 8`), the shift amount `8 * l` is 64 -- out of range on a 64-bit target. On a 64-bit release build (`overflow-checks` off), this doesn't panic: LLVM lowers `<<` to the native `shl` instruction, which masks the shift count modulo the register width, so the shift silently evaluates as if shifting by 0, giving `1` instead of the intended "no meaningful cap." The check then rejected any message longer than 1 byte for exactly the nonce length meant to be the *least* restrictive one (`L=8` is supposed to allow up to ~2^64 bytes).

Found both occurrences by grepping for the same `1 << (8 * l)` pattern after root-causing the message-mode one first -- the classic `CK_CCM_PARAMS` path had the identical bug, just never exercised by an existing test with `nonceLen == 7`.

## Fix

Replace the bare shift with `checked_shl()`, called directly on `CK_ULONG`:

```rust
let max_data_len: CK_ULONG = (1 as CK_ULONG)
    .checked_shl(8 * l as u32)
    .unwrap_or(CK_ULONG::MAX);
```

`checked_shl()`'s contract is exactly "returns `None` if the shift amount is >= the type's own bit width" -- which is precisely "no representable cap, i.e. every `CK_ULONG` value fits" -- correct on any width, with only one type (`CK_ULONG`) ever in play. The only cast is `1 as CK_ULONG`, casting the literal `1`, which is always lossless regardless of target width.

(An earlier version of this fix routed the computation through `usize` before casting back to `CK_ULONG`. That worked but was unnecessary -- `CK_ULONG` already has its own `checked_shl`/`MAX`, so there's no need for a second type at all. Thanks to @simo5 for pushing on this in review.)

## Test plan

- Added `test_aes_ccm_nonce7_allows_data_longer_than_one_byte` (classic `CK_CCM_PARAMS` path) and `test_aes_ccm_message_nonce7_allows_data_longer_than_one_byte` (message-mode path), both encrypting/decrypting a plaintext well over one byte with a 7-byte nonce and a full round trip.
- Both new tests fail on `main` and pass with this fix.
- Full `cargo test --features pqc --lib -p kryoptic-lib`: 138 passed, 0 failed.
- Verified the fixed expression type-checks cleanly under `--target i686-unknown-linux-gnu` (full crate cross-compilation isn't feasible locally without a 32-bit OpenSSL sysroot, so this was checked in isolation).

### DIFF
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -213,8 +213,19 @@ impl AesOperation {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
                 }
                 let l = 15 - params.ulNonceLen;
+                /* l can be as large as 8 (nonceLen == 7), and CK_ULONG's
+                 * width varies by platform (32 bits on i686, 64 on
+                 * x86_64/aarch64/...), so a plain `1 << (8 * l)` can shift
+                 * out of range. checked_shl()'s contract is exactly "None
+                 * if the shift amount is >= the type's own bit width",
+                 * which is precisely "no representable cap, i.e. every
+                 * CK_ULONG value fits" -- correct on any width without
+                 * needing to reason about any other type. */
+                let max_data_len: CK_ULONG = (1 as CK_ULONG)
+                    .checked_shl(8 * l as u32)
+                    .unwrap_or(CK_ULONG::MAX);
                 if params.ulDataLen == 0
-                    || params.ulDataLen > (1 << (8 * l))
+                    || params.ulDataLen > max_data_len
                     || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
                 {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
@@ -762,8 +773,19 @@ impl AesOperation {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
                 }
                 let l = 15 - params.ulNonceLen;
+                /* l can be as large as 8 (nonceLen == 7), and CK_ULONG's
+                 * width varies by platform (32 bits on i686, 64 on
+                 * x86_64/aarch64/...), so a plain `1 << (8 * l)` can shift
+                 * out of range. checked_shl()'s contract is exactly "None
+                 * if the shift amount is >= the type's own bit width",
+                 * which is precisely "no representable cap, i.e. every
+                 * CK_ULONG value fits" -- correct on any width without
+                 * needing to reason about any other type. */
+                let max_data_len: CK_ULONG = (1 as CK_ULONG)
+                    .checked_shl(8 * l as u32)
+                    .unwrap_or(CK_ULONG::MAX);
                 if params.ulDataLen == 0
-                    || params.ulDataLen > (1 << (8 * l))
+                    || params.ulDataLen > max_data_len
                     || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
                 {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
--- a/src/tests/aes.rs
+++ b/src/tests/aes.rs
@@ -2209,6 +2209,156 @@ fn test_aes_ccm_message_one_shot_length_probe() {
     testtokn.finalize();
 }
 
+/// Regression test: CKM_AES_CCM with a 7-byte nonce (the maximum-length nonce, giving the
+/// SP800-38C length field L = 15 - 7 = 8) must accept messages longer than 1 byte.
+///
+/// AesOperation::init_params computed the per-nonce-length data cap as `1 << (8 * l)`. CK_ULONG
+/// is 64 bits wide on this (LP64) target, and for l == 8 that shift amount is 64 -- out of range
+/// for a u64 shift, which silently wraps to `1 << 0 == 1` in a release build (no panic, since
+/// overflow-checks are off). The cap then rejected any data longer than 1 byte for exactly the
+/// nonce length meant to be the *least* restrictive one.
+#[test]
+#[parallel]
+fn test_aes_ccm_nonce7_allows_data_longer_than_one_byte() {
+    let mut testtokn = TestToken::initialized(
+        "test_aes_ccm_nonce7_allows_data_longer_than_one_byte",
+        None,
+    );
+    let session = testtokn.get_session(true);
+    testtokn.login();
+
+    let handle = ret_or_panic!(generate_key(
+        session,
+        CKM_AES_KEY_GEN,
+        std::ptr::null_mut(),
+        0,
+        &[(CKA_VALUE_LEN, 16),],
+        &[],
+        &[(CKA_ENCRYPT, true), (CKA_DECRYPT, true),],
+    ));
+
+    let data = b"this plaintext is well over one byte long".to_vec();
+    let tag_len = 16usize;
+    let nonce = b"1234567".to_vec();
+    assert_eq!(nonce.len(), 7);
+    let aad = b"AUTH ME".to_vec();
+
+    let mut param = CK_CCM_PARAMS {
+        ulDataLen: data.len() as CK_ULONG,
+        pNonce: nonce.as_ptr() as *mut CK_BYTE,
+        ulNonceLen: nonce.len() as CK_ULONG,
+        pAAD: aad.as_ptr() as *mut CK_BYTE,
+        ulAADLen: aad.len() as CK_ULONG,
+        ulMACLen: tag_len as CK_ULONG,
+    };
+    let mut mechanism: CK_MECHANISM = CK_MECHANISM {
+        mechanism: CKM_AES_CCM,
+        pParameter: &mut param as *mut CK_CCM_PARAMS as CK_VOID_PTR,
+        ulParameterLen: sizeof!(CK_CCM_PARAMS),
+    };
+
+    let ret = fn_encrypt_init(session, &mut mechanism, handle);
+    assert_eq!(ret, CKR_OK);
+
+    let mut enc = ret_or_panic!(encrypt_update(session, &data));
+    let mut enc_final = ret_or_panic!(encrypt_final(session));
+    enc.append(&mut enc_final);
+    assert_eq!(enc.len(), data.len() + tag_len);
+
+    let dec = ret_or_panic!(decrypt(session, handle, &enc, &mechanism));
+    assert_eq!(dec, data);
+
+    testtokn.finalize();
+}
+
+/// Same as test_aes_ccm_nonce7_allows_data_longer_than_one_byte, but for the message-mode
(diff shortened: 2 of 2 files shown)


simo5 (approved):
LGTM

simo5:
@AlexLaroche if you are being assisted by AI in these PRs, can you add an Assisted-by: trailer line in the commits that specifies which AI was used?

Thanks.

src/ossl/aes.rs
```diff
@@ -213,8 +213,15 @@ impl AesOperation {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
                 }
                 let l = 15 - params.ulNonceLen;
+                /* l can be as large as 8 (nonceLen == 7), and CK_ULONG is
+                 * 64 bits wide on LP64 targets, so a plain `1 << (8 * l)`
+                 * shifts by 64 -- out of range for a u64 shift amount.
+                 * checked_shl() returns None exactly in that case (no
+                 * representable cap, i.e. every CK_ULONG value fits). */
+                let max_data_len: CK_ULONG =
```
simo5: we may have to change this to usize so it doesn't blow up on 32 bit platforms...

AlexLaroche: Good catch — switched to usize for the shift (tracks the platform word size the same way CK_ULONG does) and cast back. Verified it type-checks clean under --target i686-unknown-linux-gnu; pushed as 5e8da35.

AlexLaroche:
Done — squashed to a single commit with an `Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>` trailer.

src/ossl/aes.rs
```diff
@@ -213,8 +213,20 @@ impl AesOperation {
...
+                 * on x86_64/aarch64/...), same as `usize` -- so a plain
+                 * `1 << (8 * l)` can shift out of range for either width.
+                 * checked_shl() returns None exactly when the shift amount
+                 * is out of range (no representable cap, i.e. every
+                 * CK_ULONG value fits), and `as CK_ULONG` on usize::MAX
+                 * in that case correctly yields CK_ULONG::MAX on any
+                 * platform where the two share a width. */
+                let max_data_len: CK_ULONG =
+                    1usize.checked_shl(8 * l as u32).unwrap_or(usize::MAX)
+                        as CK_ULONG;
```
simo5: I do not know if using `as CK_ULONG` is correct, because "as" just truncates, so a value that is > u32::MAX by 1 would be cast to 0 ?

simo5: Ie would a l = 4 which gives a 8 * l = 32 end up returning the correct amount for us?

AlexLaroche: Not on any platform this project builds for, because usize and CK_ULONG are always the same width there -- 32 bits on i686, 64 bits on x86_64/aarch64. A cast between two integer types of equal width doesn't truncate anything, it's just a relabeling of the same bits, so usize::MAX as CK_ULONG is always exactly CK_ULONG::MAX. Your scenario needs the source type to be wider than the destination, which isn't the case here.

Where it would be a real bug: on Windows, usize is 64-bit but c_ulong (CK_ULONG) is only 32-bit even on 64-bit Windows (LLP64). If this ran there, `as CK_ULONG` genuinely could drop bits the way you describe. Not a live bug since this project doesn't target Windows, but I agree the correctness of that line currently rests on an assumption the compiler doesn't check rather than something enforced -- happy to switch it to CK_ULONG::try_from(...).map_err(...)? so it fails loudly instead of silently truncating if that assumption is ever broken.

AlexLaroche: Yes. l=4 means nonceLen = 15 - 4 = 11. On a 32-bit target usize is 32 bits wide, and checked_shl's rule is "return None if the shift amount is >= the number of bits in the type" -- a shift of 32 against a 32-bit type hits that exactly (32 >= 32), so it correctly returns None and falls back to CK_ULONG::MAX.

Is that the right answer, not just the non-panicking one? Yes, separately from the shift mechanics: SP 800-38C says L=4 permits messages up to 2^32 bytes, but on a 32-bit target CK_ULONG itself can never hold anything bigger than 2^32 - 1 in the first place. So the real CCM limit for L=4 already exceeds what the type can represent, meaning "no cap beyond CK_ULONG::MAX" is the mathematically correct behavior for L=4 on 32-bit, not a lucky fallback. Same reasoning holds for l=5..8 on 32-bit and l=8 on 64-bit; l=1..3 still compute a real, smaller-than-MAX cap and enforce it.

simo5: wait if CK_ULONG and usize are the same why would changing from CK_ULONG to usize fix the error the original test was raising?  something does not add up...

simo5: oh I think you were using an explicit 1u64 before ... but that is not needed I think, we should be able to just do:

```
let max_data_len: CK_ULONG = 1.checked_shl(8 * l as u32).unwrap_or(CK_ULONG::MAX);
```

AlexLaroche: Fair question, and I owe you a clearer answer — my last reply skipped a step and made it sound like CK_ULONG and usize were somehow the cause, which was misleading.

What actually happened: the original code's bare 1 << (8 * l) already had the right type on every platform (the literal 1 infers as CK_ULONG, which is 32-bit on i686 and 64-bit on x86_64) — its only problem was the shift-overflow logic bug. My first attempt to fix that logic bug introduced 1u64, which is a fixed 64-bit type on every platform, unlike CK_ULONG. That's what broke i686: u64 no longer matched the 32-bit CK_ULONG there, so it failed to compile.

The follow-up fix replaced u64 with usize, which — like CK_ULONG — adapts its width to the platform (32-bit on i686, 64-bit on x86_64). So it wasn't "switching from CK_ULONG to usize" that fixed anything; it was undoing my own u64 regression by going back to a width that tracks the platform the way CK_ULONG always did.

Sorry for the noise.

simo5: sure but then I think we should just go back to only using CK_ULONG, there is no sense in mixing in usize if that changes nothing ...

simo5 (approved):

src/ossl/aes.rs
AlexLaroche: Agreed, done -- dropped usize entirely and now use CK_ULONG throughout: (1 as CK_ULONG).checked_shl(8 * l as u32).unwrap_or(CK_ULONG::MAX). Squashed and pushed as 08ceec1.

simo5:
Thanks!

PR #499, opened 2026-09-09
Allow zero-length data in message-mode and classic AES-CCM AEAD

### LINKED ISSUE #498: Message-mode AEAD (C_EncryptMessage/C_DecryptMessage/*MessageNext) rejects zero-length payloads with CKR_ARGUMENTS_BAD
## Summary

`C_EncryptMessage`, `C_DecryptMessage`, and their streaming `*MessageNext` counterparts unconditionally reject a zero-length data chunk with `CKR_ARGUMENTS_BAD`, even when the pointer to it is non-null. This blocks a standard AEAD use case: authenticating associated data with no attached payload (e.g. CCM/GCM with empty plaintext, authenticating only headers/metadata).

## Where

`src/fns/encryption.rs::encrypt_message`:

```rust
fn encrypt_message(
    ...
    plaintext: CK_BYTE_PTR,
    plaintext_len: CK_ULONG,
    ...
) -> Result<()> {
    if parameter.is_null()
        || parameter_len == 0
        || plaintext.is_null()
        || plaintext_len == 0
        || pul_ciphertext_len.is_null()
    {
        return Err(CKR_ARGUMENTS_BAD)?;
    }
    ...
```

`plaintext_len == 0` is bundled into the same guard as `plaintext.is_null()`, so a real, non-null, zero-length buffer is rejected exactly like a genuinely malformed (null) one — before the function ever reaches its own "is this the NULL-ciphertext length-probe call?" branch further down.

The same pattern is duplicated at the other three message-mode entry points:
- `decrypt_message`: `ciphertext.is_null() || ciphertext_len == 0`
- `encrypt_message_next`: `plaintext_part.is_null() || plaintext_part_len == 0`
- `decrypt_message_next`: `ciphertext_part.is_null() || ciphertext_part_len == 0`

## Why this is wrong

NIST SP 800-38C explicitly permits a CCM payload length of 0 (encrypting/authenticating only the associated data, no payload). Nothing in the PKCS#11 v3.x message-based encryption spec forbids an empty plaintext/ciphertext chunk either. A caller doing AAD-only authentication with a real (non-null), zero-length data pointer is making a spec-legal call, but gets `CKR_ARGUMENTS_BAD` instead.

## Reproduction

Using `CKM_AES_CCM` message-mode encryption with a real (non-null) zero-length plaintext buffer and non-empty AAD:

```
C_EncryptMessage returned CKR_ARGUMENTS_BAD
```

The same call with a 1+ byte plaintext succeeds (once #496/#497's nonce-length bug is also accounted for).

## Suggested fix

Split the null-pointer check from the zero-length check at each of the four call sites — keep rejecting a genuinely null pointer, but stop rejecting a valid, non-null, zero-length buffer. E.g. for `encrypt_message`:

```rust
if parameter.is_null()
    || parameter_len == 0
    || plaintext.is_null()
    || pul_ciphertext_len.is_null()
{
    return Err(CKR_ARGUMENTS_BAD)?;
}
```

(dropping `|| plaintext_len == 0`), and the equivalent change at the other three sites.

AlexLaroche:
## Summary

Fixes #498.

`fns/encryption.rs`'s `encrypt_message`, `decrypt_message`, `encrypt_message_next`, and `decrypt_message_next` each bundled a `data_len == 0` check into the same reject-condition as a genuinely null pointer, so a real, non-null, zero-length buffer was rejected exactly like a malformed one with `CKR_ARGUMENTS_BAD`.

NIST SP 800-38C explicitly permits a CCM payload length of 0 (authenticating only the associated data, with no attached payload), and nothing in the PKCS#11 v3.x message-based encryption spec forbids an empty plaintext or ciphertext chunk either — so a caller doing AAD-only authentication with a correctly-formed, zero-length data pointer was hitting an argument-validation bug, not a real spec violation.

## Fix

Drop each `_len == 0` condition from the four message-mode entry points, keeping the null-pointer checks.

While fixing this, I found the identical-looking `ulDataLen == 0` guard in `AesOperation::init_params` (the classic, non-message-mode `CK_CCM_PARAMS` path) was guarding a second, deeper bug: `encrypt_update`'s CCM buffering condition (`plain.len() < self.params.datalen`) can never become true when `datalen == 0` (there is nothing left to wait for), so no OpenSSL `update()` call was ever made, and `encrypt_final`'s tag retrieval failed downstream with a confusing `CKR_DEVICE_ERROR` (`ossl_ccm_get_ctx_params: tag not set`) instead of succeeding.

Added a `datalen == 0` special case that calls `update()` with an empty slice — OpenSSL's CCM provider needs at least one `EVP_EncryptUpdate` call, even empty, to leave the AAD phase and compute the tag — then dropped that guard too. (`decrypt_update`'s CCM logic already handles `datalen == 0` correctly: its trigger is "do we now have the complete buffer," which holds even when that buffer is empty, and it already calls `update()` with an empty slice in that case — this asymmetry between encrypt and decrypt is what pointed at the fix.)

## Test plan

- Added `test_aes_ccm_classic_empty_data_authenticates_aad` (classic `CK_CCM_PARAMS` path via `C_Encrypt`/`C_Decrypt`) and `test_aes_ccm_message_empty_data_authenticates_aad` (message-mode path via `C_EncryptMessage`/`C_DecryptMessage`, including the NULL-ciphertext length-probe call), both encrypting/decrypting an AAD-only CCM message (zero-length plaintext/ciphertext, non-empty AAD) and confirming a full round trip.
- Both new tests fail on `main` and pass with this fix.
- Full `cargo test --features pqc --lib -p kryoptic-lib`: 140 passed, 0 failed.
- cdylib integration tests (`cargo test --features pqc --test '*' -p kryoptic`): all passing.

### DIFF
--- a/src/fns/encryption.rs
+++ b/src/fns/encryption.rs
@@ -659,7 +659,6 @@ fn encrypt_message(
     if parameter.is_null()
         || parameter_len == 0
         || plaintext.is_null()
-        || plaintext_len == 0
         || pul_ciphertext_len.is_null()
     {
         return Err(CKR_ARGUMENTS_BAD)?;
@@ -852,7 +851,6 @@ fn encrypt_message_next(
     if parameter.is_null()
         || parameter_len == 0
         || plaintext_part.is_null()
-        || plaintext_part_len == 0
         || pul_ciphertext_part_len.is_null()
     {
         return Err(CKR_ARGUMENTS_BAD)?;
@@ -1058,7 +1056,6 @@ fn decrypt_message(
     if parameter.is_null()
         || parameter_len == 0
         || ciphertext.is_null()
-        || ciphertext_len == 0
         || pul_plaintext_len.is_null()
     {
         return Err(CKR_ARGUMENTS_BAD)?;
@@ -1249,7 +1246,6 @@ fn decrypt_message_next(
     if parameter.is_null()
         || parameter_len == 0
         || ciphertext_part.is_null()
-        || ciphertext_part_len == 0
         || pul_plaintext_part_len.is_null()
     {
         return Err(CKR_ARGUMENTS_BAD)?;
--- a/src/ossl/aes.rs
+++ b/src/ossl/aes.rs
@@ -224,8 +224,9 @@ impl AesOperation {
                 let max_data_len: CK_ULONG = (1 as CK_ULONG)
                     .checked_shl(8 * l as u32)
                     .unwrap_or(CK_ULONG::MAX);
-                if params.ulDataLen == 0
-                    || params.ulDataLen > max_data_len
+                /* ulDataLen == 0 (an AAD-only message, authenticating no payload) is
+                 * spec-legal per NIST SP 800-38C and not rejected here. */
+                if params.ulDataLen > max_data_len
                     || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
                 {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
@@ -784,8 +785,9 @@ impl AesOperation {
                 let max_data_len: CK_ULONG = (1 as CK_ULONG)
                     .checked_shl(8 * l as u32)
                     .unwrap_or(CK_ULONG::MAX);
-                if params.ulDataLen == 0
-                    || params.ulDataLen > max_data_len
+                /* ulDataLen == 0 (an AAD-only message, authenticating no payload) is
+                 * spec-legal per NIST SP 800-38C and not rejected here. */
+                if params.ulDataLen > max_data_len
                     || params.ulDataLen > (CK_ULONG::MAX - params.ulMACLen)
                 {
                     return Err(CKR_MECHANISM_PARAM_INVALID)?;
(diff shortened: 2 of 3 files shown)


simo5:
Is a zero length plaintext legal for all other mechanisms ?
You basically dropped the check for all mechanisms and then changed only the CCM code to cope with it, implicitly allowing all mechanisms to accept a zero length plaintext ?

AlexLaroche:
Good catch, checked rather than assumed. AesOperation is the only type implementing MsgEncryption/MsgDecryption, so CCM and GCM are the only two mechanisms that can reach these shared checks -- no other mechanism is affected.

For GCM specifically: msg_encrypt_next/msg_decrypt_next already guard their ctx.update() calls behind `plain.len() > 0` / `cipher.len() > 0` (predates this PR), so it never needed CCM's kind of special-casing -- it was already safe with zero-length data. CCM was the one that needed the encrypt_update fix, because its buffering trigger (`plain.len() < self.params.datalen`) can never fire when datalen == 0.

Rather than just assert this, I added test_aes_gcm_message_empty_data_authenticates_aad (AAD-only GCM round-trip via C_EncryptMessage/C_DecryptMessage) alongside the CCM ones, and it passes. Pushed as 56ce475.

simo5 (approved):
LGTM

PR #500, opened 2026-09-10
Template attributes

### LINKED ISSUE #434: Implement template attributes
Implement template attributes as described in:
https://github.com/oasis-tcs/pkcs11/pull/71

simo5:
#### Description

Implements proper support for storing and using template attibutes.

Fixes #434

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:


src/fns/keymgmt.rs
```diff
@@ -841,14 +883,41 @@ fn encapsulate_key(
...
+                    merged_tmpl,
+                    rstate.get_fips_behavior(slot_id)?,
+                )?;
+                merged_tmpl
+            } else {
+                raw_tmpl
+            }
+        } else {
+            raw_tmpl
+        };
```
Jakuje, line 879: this is repeated 3 times already. Would it be worth to externalize it into some helper function/macro. If I see right, the only difference here is the attribute name.

src/object/factory.rs
```diff
@@ -889,6 +888,152 @@ impl ObjectFactories {
...
+        }
+
+        if (ck_attr.pValue as usize) % std::mem::align_of::<CK_ATTRIBUTE>() != 0
+        {
+            ck_attr.ulValueLen = CK_UNAVAILABLE_INFORMATION;
+            if *result == CKR_OK {
+                *result = CKR_TEMPLATE_INCONSISTENT;
+            }
+            return Ok(());
+        }
```
Jakuje: where does the memory alignment requirement come from?

src/object/factory.rs
```diff
@@ -889,6 +888,152 @@ impl ObjectFactories {
...
+            ck_attr.ulValueLen = 0;
+            return Ok(());
+        }
+
+        let app_attrs: &mut [CK_ATTRIBUTE] = unsafe {
+            std::slice::from_raw_parts_mut(
+                ck_attr.pValue as *mut CK_ATTRIBUTE,
+                app_count,
+            )
+        };
```
Jakuje, line 936: could we  use `chunk_exact()` instead of the unsafe from_raw_parts?

https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact

src/tests/keys.rs
```diff
@@ -1033,3 +1033,186 @@ fn test_sensitive_attributes_matrix() {
...
+        pValue: inner_wrap_tmpl.as_ptr() as *mut _,
+        ulValueLen: CK_ULONG::try_from(std::mem::size_of_val(&inner_wrap_tmpl))
+            .unwrap(),
+    }];
+    let ret = fn_set_attribute_value(
+        session,
+        wrapping_key,
+        set_wrap_tmpl.as_ptr() as *mut _,
+        1,
+    );
```
Jakuje: why the wrap template can not be set during key generation?

src/fns/keymgmt.rs
simo5: let me look into this

src/object/factory.rs
simo5: not sure, I think this is an hallucination I missed

src/object/factory.rs
simo5: I am not sure how that would work, chunks_exact does not transmute the slice,so we get CK_ATTRIBUTE sized arrays, that we'd still need to cast unsafely back to a CK_ATTRIBUTE

src/tests/keys.rs
simo5: I guess we can

simo5:
@Jakuje I resolve the issues, PTAL

src/attribute.rs
```diff
@@ -289,6 +290,106 @@ static ATTRMAP: &[Attrmap<'_>] = &[
...
+    Ok(buf)
+}
+
+/// Deserializes a canonical byte slice into a vector of Attributes.
+pub fn deserialize_template(data: &[u8]) -> Result<Vec<Attribute>> {
+    if data.is_empty() {
+        return Ok(Vec::new());
+    }
+    if data.len() < 4 {
+        return Err(CKR_DEVICE_MEMORY)?;
```
Jakuje: I wonder if this is the most appropriate error.  I would consider `CKR_DEVICE_ERROR` a better one or the general one `CKR_GENERAL_ERROR` (applies also to the other return values from this function).

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ Unreleased changes
   - [Implement
      CKM_AES_KEY_WRAP_PKCS7](https://github.com/latchset/kryoptic/pull/501)
 
+* Added support for template attributes like CKA_DERIVE_TEMPLATE, which can
+  be used to enforce specific attribute on key create through use of other keys
+  (Derivation, encapsualtion, wrapping)
```
Jakuje:
```suggestion
  (derivation, encapsualtion, decapsulation, wrapping).
```

src/tests/mlkem.rs
```diff
@@ -718,3 +718,195 @@ fn test_mlkem_public_key_info() {
...
+        pValue: &mut enc_true as *mut _ as *mut std::ffi::c_void,
+        ulValueLen: sizeof!(CK_BBOOL),
+    }];
+    let set_enc_tmpl = [CK_ATTRIBUTE {
+        type_: CKA_ENCAPSULATE_TEMPLATE,
+        pValue: inner_enc_tmpl.as_ptr() as *mut _,
+        ulValueLen: CK_ULONG::try_from(std::mem::size_of_val(&inner_enc_tmpl))
+            .unwrap(),
+    }];
+    let ret = fn_set_attribute_value(
```
Jakuje: the same comment I did yesterday about the wrapping -- here we do the same rain dance of generating key pair and then setting encapsulate/decapsulate templates for the keys using separate calls. While I think we should test both of the paths, I find providing this directly at key creation time cleaner.

The specs really do not specify if it should be modifiable or not after creation (and in fact the `CKA_ENCAPSULATE_TEMPLATE`  is never even mentioned in the specs), but I would consider it sane default to allow setting this group of attributes only at creation time, adding the appropriate `OAFlags::SettableOnlyOnCreate` flags to the appropriate factories, what do you think? Keeping this attribute freely editable makes the enforcement completely bollocks. 

Actually the same comment applies to `CKA_ALLOWED_MECHANISMS` which can be considered a similar "enforcement" mechanism, which could be simply overridden by SetAttributeValue without any restrictions.

src/attribute.rs
simo5: let's do CKR_DEVICE_ERROR

CHANGELOG.md
simo5: well decapsulation is part of encapsulation ( I did not mention unwrapping either )

src/tests/mlkem.rs
simo5: The spec does not say this can be only set at create time, and so I do not think we should enforce that (some tests also depend on being able to change it.

The enforcement is voluntary it is not meant to block the user actually owning the key

simo5:
@Jakuje I should have addressed all the points, PTAL

src/tests/mlkem.rs
Jakuje: Ok, I guess it would be for longer discussion. I think otherwise, but it should not block this PR.

Jakuje (approved):

simo5:
Thanks for the review!

PR #501, opened 2026-09-10
Implement CKM_AES_KEY_WRAP_PKCS7

simo5:
#### Description

Adds the missing CKM_AES_KEY_WRAP_PKCS7 implementation, this mechanism is identical to the deprecated CKM_AES_KEY_WRAP_PAD and can be used do unwrap keys wrapped with that mechanism.

I believe this supersedes #448 

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:


simo5:
@ab PTAL

src/ossl/aes.rs
Jakuje, line 646: the `encrypt_new` and `decrypt_new` function are almost the same, except for the `op: CKF_DECRYPT` or `op: CKF_ENCRYPT` parameter and `cipher_initialize` last parameter. I think we could avoid code duplication and create also some shared helper (likely in separate PR) as there is already too much of (mostly FIPS) boilerplate.

src/ossl/aes.rs
```diff
@@ -1491,14 +1506,40 @@ impl Encryption for AesOperation {
...
                 return Err(CKR_GENERAL_ERROR)?;
             }
         };
 
         #[cfg(feature = "fips")]
-        self.fips_approval.finalize();
+        {
+            if self.mech == CKM_AES_KEY_WRAP_PKCS7 {
+                self.fips_approval.set(false);
+            }
```
Jakuje: we do set fips approval to false at init time already. There is no way it would be in the allowed state at finalize time if I right, so I think this is technically a dead code?

src/ossl/aes.rs
simo5: could be spun off as a separate issue

src/ossl/aes.rs
simo5: probably

simo5: In fact I just realized the whole exercise of setting the fips_approval for the mechanism is moot, because the mechanism is declared not fips approved at the mechanism level, so it is never approved, regardless what we do in the implementation.

Let me strip away these useless checks.

src/ossl/aes.rs
Jakuje: Created #508.

Jakuje (approved):

PR #503, opened 2026-09-11
Fix SP800-108 DKM length: embed bit-length, not byte-length

### LINKED ISSUE #502: SP800-108 KDF: DKM length (L) encoded in bytes instead of bits, corrupting derived output
## Summary

`CKM_SP800_108_COUNTER_KDF` (and presumably `_FEEDBACK_KDF`/`_DOUBLE_PIPELINE_KDF`, which share the same code path) derives incorrect key material whenever a caller includes a `CK_SP800_108_DKM_LENGTH` segment. The derived output doesn't match NIST SP 800-108 or other conformant implementations (e.g. .NET's `System.Security.Cryptography.SP800108HmacCounterKdf`) for the exact same inputs.

## Root cause

`src/native/sp800_108.rs`, `dkm_update()`:

```rust
fn dkm_update(
    param: &Sp800DKMLengthFormat,
    klen: usize,
    slen: usize,
    op: &mut Box<dyn Mac>,
) -> Result<()> {
    let mut len = match param.method {
        CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS => slen,
        CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS => klen,
        _ => return Err(CKR_MECHANISM_PARAM_INVALID)?,
    } as u64;
    /* up to 64 bits */
    match param.bits {
        8 => { ... }
        16 => { ... }
        ...
    }
}
```

`klen`/`slen` are computed (in the caller, around line 421-434) from `CKA_VALUE_LEN`, which PKCS#11 defines in **bytes**. `dkm_update` feeds that byte count directly into the length field, only truncating/formatting it per the requested bit-width (`param.bits`) -- there is no `* 8` conversion anywhere in this function or its callers.

NIST SP 800-108 defines `L` explicitly as "the length of the output... in **bits**". Every conformant implementation (including .NET's `SP800108HmacCounterKdf`) encodes the bit-length there, not the byte-length. Since `L` is part of the fixed input hashed on *every* counter iteration, this corrupts the entire derived output whenever a `CK_SP800_108_DKM_LENGTH` segment is used -- not just an edge case.

## Reproduction

Derive a key with `CKM_SP800_108_COUNTER_KDF`, PRF data including a `CK_SP800_108_DKM_LENGTH` segment (method `CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS`, 32-bit big-endian), and compare against a NIST-vector or `SP800108HmacCounterKdf`-produced key of the same length/label/context: the two do not match. E.g. for a 32-byte (256-bit) derived key, Kryoptic embeds `32` (`0x00000020`) as `L`; a conformant implementation embeds `256` (`0x00000100`).

## Suggested fix

Convert `klen`/`slen` from bytes to bits before truncating/formatting, e.g.:

```rust
let mut len = match param.method {
    CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS => slen,
    CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS => klen,
    _ => return Err(CKR_MECHANISM_PARAM_INVALID)?,
} as u64
    * 8;
```

(keeping the existing "intentionally truncated by casting" behavior for the subsequent `% maxsize!(N)` + width-based formatting).

Both `CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS` and `CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS` go through this same function, so both are affected.

## Environment

Found while testing the Kryoptic backend against `main` (currently at `926417f84f5595277efcbe38b2849a57b8e8269c`) from a C# PKCS#11 wrapper's test suite, comparing derived output against .NET's `SP800108HmacCounterKdf` for identical inputs. The affected code is unrelated to anything changed recently in this range, so this is not a regression -- a pre-existing bug.

AlexLaroche:
## Summary

Fixes #502.

`dkm_update()` fed `klen`/`slen` (computed from `CKA_VALUE_LEN`, which PKCS#11 defines in bytes) directly into the `CK_SP800_108_DKM_LENGTH` field with no conversion. NIST SP 800-108 defines `L` (the DKM length embedded in the fixed input) as the output length in **bits**. Since `L` is part of the fixed input hashed on every counter iteration, this corrupted the entire derived output whenever a caller used a `CK_SP800_108_DKM_LENGTH` segment -- not an edge case.

## Fix

Multiply by 8 (in `u64`, after the cast, so it can't overflow `usize` on 32-bit targets) before the existing width-based truncation/formatting in `src/native/sp800_108.rs::dkm_update`. Both `CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS` and `CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS` go through this same function, so both were affected and both are fixed by the same change.

## Test plan

- Added `test_sp800_kdf_dkm_length_is_in_bits` with an independently-computed expected value: `HMAC-SHA256(KI, counter=1 (32-bit BE) || L=256 (32-bit BE, the bit-length of a 32-byte key))`. A single iteration exactly fills a 32-byte output with a 32-byte-digest PRF, so no truncation/concatenation logic is exercised -- isolating the DKM-length encoding itself.
- Confirmed the new test fails on `main` (produces the byte-length-based output) and passes with this fix.
- Full `cargo test --features pqc --lib -p kryoptic-lib`: 142 passed, 0 failed.

### DIFF
--- a/src/native/sp800_108.rs
+++ b/src/native/sp800_108.rs
@@ -190,19 +190,24 @@ impl Sp800Operation {
     /// Feed the length of the sum of the segments or of the keys
     /// to be produced
     ///
+    /// `klen`/`slen` must already be in *bits* (NIST SP800-108 defines L, the DKM
+    /// length fed into the fixed input, as the output length in bits) -- converting
+    /// from PKCS#11's byte-based CKA_VALUE_LEN is the caller's responsibility, done
+    /// once where klen/slen are computed rather than on every call here.
+    ///
     /// NOTE: In this function the len is intentionally truncated by
     /// casting, do not convert with try_from()
     fn dkm_update(
         param: &Sp800DKMLengthFormat,
-        klen: usize,
-        slen: usize,
+        klen: u64,
+        slen: u64,
         op: &mut Box<dyn Mac>,
     ) -> Result<()> {
         let mut len = match param.method {
             CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS => slen,
             CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS => klen,
             _ => return Err(CKR_MECHANISM_PARAM_INVALID)?,
-        } as u64;
+        };
         /* up to 64 bits */
         match param.bits {
             8 => {
@@ -286,8 +291,8 @@ impl Sp800Operation {
         params: &Vec<Sp800Params>,
         op: &mut Box<dyn Mac>,
         ctr: usize,
-        dkmklen: usize,
-        dkmslen: usize,
+        dkmklen: u64,
+        dkmslen: u64,
     ) -> Result<()> {
         let mut seen_dkmlen = false;
         let mut seen_iter = false;
@@ -332,8 +337,8 @@ impl Sp800Operation {
         op: &mut Box<dyn Mac>,
         iv: &[u8],
         ctr: usize,
-        dkmklen: usize,
-        dkmslen: usize,
+        dkmklen: u64,
+        dkmslen: u64,
     ) -> Result<()> {
         let mut seen_dkmlen = false;
         let mut seen_iter = false;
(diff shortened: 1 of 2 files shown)


src/native/sp800_108.rs
```diff
@@ -198,11 +198,17 @@ impl Sp800Operation {
         slen: usize,
         op: &mut Box<dyn Mac>,
     ) -> Result<()> {
+        /* klen/slen come from CKA_VALUE_LEN and the HMAC segment size, both of which
```
simo5: This is the worng place for the fix.
The dkm_update function was always supposed to get the klen and slen in bits. The bug is in the caller pass it in bytes instead.

The caller should be changed to multiply by eight the values before passing them as inputs to dkm_update

simo5 (changes requested):
right fix, wrong place, easy change

src/native/sp800_108.rs
simo5: note that if usize is not sufficient, it is ok to change the type of slen/klen to u64

simo5 (approved):
LGTM

PR #505, opened 2026-09-12
Add support for CKM_CHACHA20 and CKM_CHACHA20_POLY1305

### LINKED ISSUE #504: Support CKM_CHACHA20/CKM_CHACHA20_POLY1305 as PKCS#11 mechanisms
## Summary

Kryoptic doesn't expose ChaCha20 or ChaCha20-Poly1305 as PKCS#11 mechanisms -- `CKM_CHACHA20`, `CKM_CHACHA20_KEY_GEN`, and `CKM_CHACHA20_POLY1305` only appear in `src/pkcs11/mod.rs`'s `const_to_elem!` name-lookup table (used for trace/debug output, alongside plenty of other constants Kryoptic doesn't implement, e.g. the GOST mechanisms right next to them), with no mechanism registration, no dispatch, and no `CKK_CHACHA20` key-object factory anywhere.

## What already exists

The underlying crypto is already built and working -- it's just not connected to the PKCS#11 surface. The shared `ossl` crate (a Cargo workspace member of `kryoptic-lib` itself, not just of `rustls/ossl`) has a complete `EncAlg::ChaCha20`/`EncAlg::ChaCha20Poly1305` cipher-context implementation in `ossl/src/cipher.rs`, including the AEAD setup path (`ctx.aead_setup(...)`), with its own test coverage (`ossl/src/tests/chacha20.rs`). This was added for the `rustls/ossl` TLS provider (`e60364edb`, `a3f51f8df`), but the engine itself has no dependency on TLS and is directly usable from `kryoptic-lib`.

## What's missing on the PKCS#11 side

1. A mechanism module (e.g. `src/chacha20.rs`, parallel to `src/ossl/aes.rs`) implementing `CKM_CHACHA20`/`CKM_CHACHA20_POLY1305`/`CKM_CHACHA20_KEY_GEN` against `EncAlg::ChaCha20`/`ChaCha20Poly1305`.
2. A `CKK_CHACHA20` key-object factory (mirroring `AesKeyFactory`) -- no trace of one today.
3. Parameter marshalling for `CK_CHACHA20_PARAMS`/`CK_SALSA20_CHACHA20_POLY1305_PARAMS`.
4. A Cargo feature flag -- `Cargo.toml` has `aes`, `rsa`, `mlkem`, etc., but no `chacha20`/`chacha`.
5. A registration point in `src/enabled.rs`'s `register_all()` -- every mechanism family has exactly one `#[cfg(feature = "...")] mod ...;` plus one `<name>::register(mechs, ot);` line there; ChaCha20 has neither.

## Suggested approach

Follow the existing per-family pattern exactly (`src/ossl/aes.rs` + its `enabled.rs` wiring is the closest template), building on `EncAlg::ChaCha20`/`ChaCha20Poly1305` in the shared `ossl` crate rather than reimplementing the cipher.

AlexLaroche:
## Summary

Adds the two PKCS#11 v3.0 ChaCha20 mechanisms, which previously had no implementation at all: `CKM_CHACHA20`/`CKM_CHACHA20_KEY_GEN`/`CKM_CHACHA20_POLY1305` only appeared in `pkcs11/mod.rs`'s `const_to_elem!` name-lookup table (used for trace/debug output, alongside plenty of other constants this project doesn't implement, e.g. the GOST mechanisms right next to them there) -- no mechanism registration, no dispatch, and no `CKK_CHACHA20` key-object factory existed.

## What already existed

The underlying crypto was already built and working. The shared `ossl` crate (a workspace member of `kryoptic-lib` itself, not just of `rustls/ossl`) has a complete `EncAlg::ChaCha20`/`EncAlg::ChaCha20Poly1305` cipher-context implementation in `ossl/src/cipher.rs`, including the AEAD setup path, with its own test coverage (`ossl/src/tests/chacha20.rs`). It was added for the `rustls/ossl` TLS provider, but has no dependency on TLS and is directly usable from `kryoptic-lib`.

## What this PR adds

- A new `chacha20` Cargo feature (added to `standard`), registered in `enabled.rs`.
- `src/chacha20.rs` -- a `CKK_CHACHA20` key factory (fixed 256-bit key, unlike AES's several selectable sizes) and mechanism registration.
- `src/ossl/chacha20.rs` -- the classic `Encryption`/`Decryption` operation (adapted from `AesOperation`'s CTR/GCM paths), **and** `MsgEncryption`/`MsgDecryption` message-mode support for `CKM_CHACHA20_POLY1305` (`C_MessageEncryptInit`/`C_EncryptMessage`/`C_EncryptMessageNext`/`C_MessageEncryptFinal` and decrypt counterparts).

Message-mode support turned out not to be optional: [pkcs11-provider](https://github.com/latchset/pkcs11-provider) -- the flagship downstream consumer this project's own CI tests against -- doesn't use the classic params path for `CKM_CHACHA20_POLY1305` at all. Its `src/cipher.c` is built entirely around `CK_SALSA20_CHACHA20_POLY1305_MSG_PARAMS`, the same message-mode API `CKM_AES_GCM`/`CKM_AES_CCM` already implement here. Testing an earlier, classic-only revision of this PR against pkcs11-provider's own CI confirmed it: the AEAD test failed with `EVP_Encrypt* failed!`, and the TLS test negotiated `ECDHE-ECDSA-CHACHA20-POLY1305` and then failed decryption with `bad record mac`. `CKM_CHACHA20_POLY1305` gets its own mechanism-registry entry with `CKF_MESSAGE_ENCRYPT`/`CKF_MESSAGE_DECRYPT`/`CKF_MULTI_MESSAGE` (mirroring how GCM/CCM get a separate, message-flagged entry from AES's other modes); plain `CKM_CHACHA20` stays classic-only, since message mode has no use case for a non-AEAD stream cipher.

## FIPS

Neither mechanism is added to the `fips` feature bundle -- ChaCha20 and ChaCha20-Poly1305 are not NIST/FIPS-approved algorithms (unlike AES-GCM/CCM). This follows the existing convention for `hotp` (also excluded from `fips` for the same kind of reason): the feature simply isn't listed as a `fips` dependency, no additional `#[cfg(not(feature = "fips"))]` guard needed on top of that, since a FIPS build doesn't combine extra, non-approved features on its own.

## Scope

- Only the IETF/RFC 8439 `CK_CHACHA20_PARAMS` layout (96-bit nonce, 32-bit block counter) is supported for plain `CKM_CHACHA20` -- the spec's alternative, original (64-bit-nonce/64-bit-counter) layout is rejected with `CKR_MECHANISM_PARAM_INVALID`, since OpenSSL's `"ChaCha20"` cipher only implements the IETF layout as a single fixed-width IV.
- Wrap/unwrap, key derivation, and MAC are **not** implemented for either mechanism in this change.

## Testing

`src/tests/chacha20.rs` reuses the RFC 8439 test vectors from `ossl/src/tests/chacha20.rs`, which already vets them against the raw OpenSSL cipher this mechanism is built on:
- ChaCha20 known-answer vector, both one-shot and split across arbitrary `C_EncryptUpdate` chunk sizes.
- ChaCha20-Poly1305 known-answer key/nonce/AAD: round-trip and tamper-rejection, classic and message-mode.
- Multi-part encrypt/decrypt tests for both the classic path (tag-tail buffering in small, uneven chunks) and message mode (`C_EncryptMessageBegin`/`C_EncryptMessageNext`).
- A message-mode test reusing one initialized session across two independently-nonced messages -- the reason message mode exists.
- Key generation and parameter/key-size validation.

Full local suite: 138 `kryoptic-lib` tests pass, 0 failures.

## Known CI failure: pkcs11-provider bug, not this PR's

The `pkcs11-provider` integration job's `aead` and `tls` tests still fail on this PR, but I've root-caused both to the exact same pre-existing bug in `pkcs11-provider` itself, reproduced locally with debug tracing on both sides:

- **`aead` test**: `C_MessageEncryptInit` fails with `CKR_KEY_TYPE_INCONSISTENT`. `pkcs11-provider`'s generic raw-key-import path for ciphers (`p11prov_cipher_legacy_init` in `src/cipher.c`, comment literally says *"importing the AES key"*) hardcodes `CKK_AES` for any cipher under test, including ChaCha20 -- a leftover from when AES was the only cipher family it exercised. Our `ChaChaOperation` correctly requires `CKA_KEY_TYPE == CKK_CHACHA20` per spec, so the mismatched key is rejected.
- **`tls` test**: same root cause, one level removed. The suite runs twice: once with only the server routing crypto through `pkcs11-provider` (that pass fully succeeds, including ChaCha20 -- verified byte-correct against OpenSSL's own default-provider crypto in both directions), and once with `-propquery ?provider=pkcs11` forcing *both* sides through it. Only in that second pass does `pkcs11-provider` also route the TLS traffic-key import through the same hardcoded-`CKK_AES` path used by the `aead` test, so `C_MessageDecryptInit` fails with `CKR_KEY_TYPE_INCONSISTENT` on the very first encrypted record -- surfacing as "bad record mac" rather than a clean init error, but confirmed via debug tracing to be the identical cause. AES suites pass under the same second pass because `CKK_AES` happens to be correct for them.

Loosening our key-type check to accept `CKK_AES` for `CKM_CHACHA20_POLY1305` would paper over this rather than fix it (real type confusion), so I left it as-is. This looks like something to fix on the `pkcs11-provider` side.

Fixes: https://github.com/latchset/kryoptic/issues/504

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ ossl.workspace = true
 
 [features]
 aes = []
+chacha20 = []
 ecc = []
 ecdsa = ["ecc"]
 ecdh = ["ecc"]
@@ -106,7 +107,7 @@ ecc_all = ["ecc_min", "ec_montgomery", "eddsa"]
 hash_all = ["hash", "hmac"]
 kdf_all = ["hkdf", "pbkdf2", "sp800_108", "sshkdf", "tlskdf", "simplekdf"]
 
-standard = ["sqlitedb", "ecc_all", "ffdh", "hash_all", "kdf_all", "rsa", "hotp"]
+standard = ["sqlitedb", "ecc_all", "ffdh", "hash_all", "kdf_all", "rsa", "hotp", "chacha20"]
 
 fips = ["ossl/fips", "sqlitedb", "rusqlite/bundled", "aes", "ecc_all", "ffdh",
 "hash_all", "kdf_all", "rsa", "pqc", "dep:vsprintf"]
--- a/src/chacha20.rs
+++ b/src/chacha20.rs
@@ -1,0 +1,287 @@
+// Copyright 2026 Alexandre Laroche
+// See LICENSE.txt file for terms
+
+//! This module implements the PKCS#11 mechanisms for ChaCha20 and
+//! ChaCha20-Poly1305 as defined in
+//! [RFC 8439](https://www.rfc-editor.org/rfc/rfc8439): _ChaCha20 and
+//! Poly1305 for IETF Protocols_.
+//!
+//! Only the IETF parameter layout is supported: a 96-bit nonce (and, for
+//! `CKM_CHACHA20`, a 32-bit block counter) -- the shape OpenSSL's
+//! `EncAlg::ChaCha20`/`EncAlg::ChaCha20Poly1305` ciphers expect. PKCS#11
+//! v3.0's alternative, original (64-bit-nonce/64-bit-counter) layout for
+//! `CKM_CHACHA20` is not implemented.
+//!
+//! `CKM_CHACHA20_POLY1305` also implements message-mode
+//! (`C_MessageEncryptInit`/`C_EncryptMessage`/`C_EncryptMessageNext`/
+//! `C_MessageEncryptFinal` and their decrypt counterparts), since that is
+//! how real-world consumers (e.g. pkcs11-provider, for TLS record
+//! encryption) actually drive it.
+//!
+//! Neither mechanism is included in the `fips` feature bundle: ChaCha20
+//! and ChaCha20-Poly1305 are not NIST/FIPS-approved algorithms (unlike
+//! AES-GCM/CCM), so they are excluded the same way `hotp` already is --
+//! by simply never appearing in the `fips` feature's dependency list,
+//! rather than needing a `#[cfg(not(feature = "fips"))]` guard on top of
+//! that (a FIPS build never combines extra, non-approved features on
+//! its own).
+
+use std::fmt::Debug;
+use std::sync::LazyLock;
+
+use crate::attribute::Attribute;
+use crate::error::Result;
+use crate::mechanism::*;
+use crate::object::*;
+use crate::ossl::chacha20::*;
+use crate::pkcs11::*;
+
+/// ChaCha20 key size (256 bits), fixed by RFC 8439.
+pub const CHACHA20_KEY_SIZE: usize = 32;
+
+/// Object that holds ChaCha20 Mechanisms
+///
+/// `CKM_CHACHA20_POLY1305` gets its own entry, with the extra
+/// `CKF_MESSAGE_ENCRYPT`/`CKF_MESSAGE_DECRYPT`/`CKF_MULTI_MESSAGE` flags:
+/// message-mode (`C_MessageEncryptInit`/`C_EncryptMessage`/...) is how
+/// real-world PKCS#11 consumers (e.g. pkcs11-provider, for TLS record
+/// encryption) actually drive AEAD ciphers, reusing one initialized cipher
+/// context across many independently-nonced messages -- mirroring how
+/// `CKM_AES_GCM`/`CKM_AES_CCM` get a separate, message-flagged entry from
+/// AES's other, non-AEAD modes. Plain `CKM_CHACHA20` has no such use case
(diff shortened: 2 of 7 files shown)


AlexLaroche:
https://github.com/openssl-projects/pkcs11-provider/pull/793

simo5 (approved):
LGTM

PR #510, opened 2026-09-15
Add checked constructor for OsslContext

### LINKED ISSUE #509: Improper validation of unsafe function return for OSSL_LIB_CTX_new()
> ## Summary
> 
> `ossl`'s `OsslContext::new_lib_ctx()` stores the raw return value of `OSSL_LIB_CTX_new()` without checking it for NULL. `OSSL_LIB_CTX_new()`
> is a C function that can legitimately fail and return NULL. When it does, the NULL context flows into `load_legacy_provider()` → `OSSL_PROVIDER_available(NULL, ...)`, which crashes inside OpenSSL (`pthread_rwlock_wrlock` on a NULL object) instead of returning an
> error.
> 
> We hit this in production through `rpm-sequoia` 1.10.3 / `sequoia-openpgp` 2.4.1 (crypto-openssl backend) on openEuler: openscap calls `rpmReadConfigFiles()` from a worker thread, which reaches `sequoia_openpgp::crypto::backend()`'s `OnceLock::get_or_init()` — the first (and, in that environment, failing) call of `OSSL_LIB_CTX_new()`.
> 
> ## Affected code
> 
> `ossl` 1.5.2, `src/lib.rs`:
> 
> ```rust
> pub fn new_lib_ctx() -> OsslContext {
>     OsslContext {
>         context: unsafe { OSSL_LIB_CTX_new() },   // <-- NULL not checked
>         providers: Vec::new(),
>         #[cfg(feature = "fips")]
>         is_fips: false,
>     }
> }
> ```
> 
> The caller in `sequoia-openpgp` 2.4.1, `src/crypto/backend/openssl.rs::context()`, then swallows the failure of the *next* call as well:
> 
> ```rust
> OSSL_CONTEXT.get_or_init(|| {
>     let mut context = ossl::OsslContext::new_lib_ctx();
>     // Try to load legacy provider. Ignore the failures as the worst
>     // what could happen is that we won't support some of the legacy
>     // algorithms.
>     let _ = context.load_legacy_provider();   // <-- crashes before returning
>     context
> })
> ```
> 
> The comment's premise ("the worst that could happen is that we won't support some of the legacy algorithms") does not hold: if `OSSL_LIB_CTX_new()` returned NULL, `load_legacy_provider()` segfaults on that NULL.
> 
> ## Observed crash (gdb backtrace)
> 
> Environment: openEuler (glibc 2.38, OpenSSL 3.5.8), `oscap xccdf eval` (runs `rpmReadConfigFiles()` from worker threads), rpm-sequoia 1.10.3 with sequoia-openpgp 2.4.1, crypto-openssl backend:
> 
> ```text
> Thread 10 "common_main" received signal SIGSEGV
> #0  pthread_rwlock_wrlock () from /usr/lib64/libc.so.6          # deref NULL+0x18
> #1  CRYPTO_THREAD_write_lock () from /usr/lib64/libcrypto.so.3
> #2  ERR_get_next_error_library () from /usr/lib64/libcrypto.so.3
> #3  ?? () from /usr/lib64/libcrypto.so.3
> #4  OSSL_PROVIDER_available () from /usr/lib64/libcrypto.so.3
> #5-#11 ?? () from /usr/lib64/librpm_sequoia.so.1
> #12 _rpmInitCrypto () from /usr/lib64/librpm_sequoia.so.1
> #13 rpmReadConfigFiles () from /usr/lib64/librpm.so.9
> #14-#15 ?? () from /usr/lib64/libopenscap.so.33
> ```
> 
> `segfault at 18 ip ... in libc.so.6[89a06, base+1c6000]` — a write through a NULL pointer at offset 0x18, consistent with `CRYPTO_THREAD_write_lock` locking a NULL `CRYPTO_RWLOCK` inside OpenSSL's provider store: with a NULL libctx, `OSSL_PROVIDER_available` cannot get the store and proceeds over a NULL pointer.
> 
> ## Why this matters beyond our case
> 
> - `OSSL_LIB_CTX_new()` is documented to be able to fail (allocation failure; failure to load/parse a configuration file when the libctx is created with a config — `OSSL_LIB_CTX_load_config` / `config_diagnostics`). Any caller is then one unchecked step away from a segfault.
> - In `sequoia-openpgp`'s current design, `context()` is the process-wide entry point for the whole crypto-openssl backend. A transient failure (e.g. racing global OpenSSL init in a multithreaded embedder) turns into a hard crash of the embedding process (in rpm's case, the FFI wrapper `catch_unwind`-aborts, but here the crash is a raw SIGSEGV, which no catch_unwind can intercept).
> 
> ## Suggested fixes
> 
> In `ossl`:
> 
> a. Make `OsslContext::new_lib_ctx()` (and `from_ctx`) return a `Result<OsslContext, Error>` when `OSSL_LIB_CTX_new()` yields NULL, and/or add `assert!(!self.context.is_null())` guards in `load_legacy_provider()`/`Drop` so the failure is loud and Rust-side, not a C-level segfault.
> 
> b. Optionally also guard `OSSL_LIB_CTX_free(NULL)` (currently a no-op in OpenSSL, but the pattern should not rely on it).
> 
> In `sequoia-openpgp` (secondary):
> 
> c. Handle the `Result` from the above in `src/crypto/backend/openssl.rs::context()` — e.g. fall back to the process default context (`OSSL_LIB_CTX_get0_default()`) so that a failing dedicated libctx degrades into "backend uses the default context" instead of a crash.
> 
> ## Environment where observed
> 
> - openEuler mainline (glibc 2.38, OpenSSL 3.5.8-1.oe1), x86_64
> - rpm-sequoia 1.10.3, sequoia-openpgp 2.4.1 (crypto-openssl), ossl 1.5.2
> - Trigger: `oscap xccdf eval` (openscap 1.4.4) calling `rpmReadConfigFiles()` from worker threads
> - Control experiment: the same environment runs fine with a rpm-sequoia build whose `_rpmInitCrypto` does not call into the backend — the crash appears exactly when `OSSL_LIB_CTX_new()` is first invoked from the worker thread. 

 _Originally posted by @fundawang in [#319](https://github.com/latchset/kryoptic/issues/319#issuecomment-5678399944)_

simo5:
#### Description

Introduce `new_lib_ctx_checked()` to explicitly verify that `OSSL_LIB_CTX_new()` succeeds and does not return a null pointer. Previously, `new_lib_ctx()` silently retained null pointers, which OpenSSL treats as the global default context. Update internal context initialization to use the checked constructor and add documentation explaining null context behavior.

Even though I think the analysis on the bug is mostly bogus AI hallucinated non-sense it is a good thing to catch this case which may happen when the openssl configuration file is broken in some cases.

Fixes #509 

#### Checklist

- [ ] Test suite updated

#### Reviewer's checklist:


ossl/src/lib.rs
```diff
@@ -260,6 +260,10 @@ impl OsslContext {
     /// For most of the applications, its recommented to load the default
     /// configuration file after this call to make sure the application
     /// behaves as expected using `load_default_configuration()`.
+    ///
+    /// Deprecation notice:
+    /// On OSSL_LIB_CTX_new() failure this function silently returns the
+    /// default context (NULL ptr), use new_lib_ctx_checked() instead.
     pub fn new_lib_ctx() -> OsslContext {
```
Jakuje, line 270: is there some semantics to mark the API deprecated other than just writing it in the doc text?

simo5: yes added deprecation attribute as documented in https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

Jakuje (approved):

simo5:
Thanks for the review!

PR #513, opened 2026-09-16
Make handling slices of structures from FFI safer

simo5:
#### Description

According to the from_raw_parts[_mut] documentation pointers received
from OpenSSL or callbacks may not always be properly aligned, which can
lead to undefined behavior when creating Rust slices or dereferencing
pointers.

Add helper functions that makes it safe to use slices, copying them when
they are not properly aligned.

#### Checklist

- [x] Test suite updated

#### Reviewer's checklist:


ossl/src/pkey.rs
```diff
@@ -1232,40 +1232,43 @@ unsafe extern "C" fn export_params_callback(
...
+            let p = std::ptr::read_unaligned(counter);
+            if p.key.is_null() {
+                break;
             }
+            total_size += std::mem::size_of::<OSSL_PARAM>();
+            total_size += p.data_size;
+            counter = counter.offset(1);
+        }
+        let max_ptr =
+            (params as *const u8).wrapping_add(total_size) as *const c_void;
```
Jakuje, line 1257: why wrapping_add here?

src/misc.rs
```diff
@@ -56,14 +58,64 @@ macro_rules! sizeof {
...
 /// Uses unsafe functions:
 /// - std::slice::from_raw_parts()
 ///
-/// If len is 0 and empty slice reference is returned
-pub(crate) unsafe fn bytes_to_slice<'a, T>(
+/// If len is 0 an empty slice reference is returned
+pub(crate) fn bytes_to_slice<'a, T: Byte>(
     ptr: *const T,
     len: usize,
 ) -> &'a [T] {
```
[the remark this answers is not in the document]
simo5: ah I was sure I had fixed this, good catch ...

src/fns/objmgmt.rs
```diff
@@ -270,14 +263,7 @@ fn get_attribute_value(
...
     let cnt = usize::try_from(count).map_err(|_| CKR_ARGUMENTS_BAD)?;
-    if cnt > 0 && template.is_null() {
-        return Err(CKR_ARGUMENTS_BAD)?;
-    }
-    let mut tmpl: &mut [CK_ATTRIBUTE] = if cnt > 0 {
-        unsafe { std::slice::from_raw_parts_mut(template, cnt) }
-    } else {
-        &mut []
-    };
+    let mut tmpl = struct_to_slice_mut(template, cnt)?;
```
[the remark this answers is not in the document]
simo5: this is bogus usize implicitly gets SliceLen trait in misc.rs

src/pkcs11/mod.rs
```diff
@@ -1084,18 +1081,13 @@ impl CK_OTP_PARAMS {
     /// Returns the internal data memory buffer as a slice
     ///
     /// Errors out if the internal data pointer is null
-    pub fn to_slice(&self) -> Result<&[CK_OTP_PARAM]> {
+    pub fn to_slice(&self) -> Result<Cow<'_, [CK_OTP_PARAM]>> {
```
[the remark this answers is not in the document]
simo5: this is not a problem, only the FFI interface is supported as stable

ossl/src/pkey.rs
simo5: uhmm, yeah this is backwards ... I'll fix

src/fns/keymgmt.rs
```diff
@@ -305,13 +300,8 @@ fn wrap_key(
...
     let pwraplen = unsafe { *pul_wrapped_key_len as CK_ULONG };
-    let wrapped: &mut [u8] = if wrapped_key.is_null() {
-        &mut [] /* empty buffer will be always too small */
-    } else {
-        let wraplen =
-            usize::try_from(pwraplen).map_err(|_| CKR_ARGUMENTS_BAD)?;
-        unsafe { std::slice::from_raw_parts_mut(wrapped_key, wraplen) }
-    };
+    let wraplen = ZeroOk(parse_len(pwraplen)?);
+    let wrapped = bytes_to_slice_mut(wrapped_key, wraplen)?;
```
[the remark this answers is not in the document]
simo5: I have actually changed this code at least three times before opening the PR, switching from NullOk on the pointer to ZerOk ... but I forgot already why I was flipping.

But thinking again a zero length array is generally fine, the only case that is problematic is null ptr with non-zero length

And sounds like handling it as part of the function is not going to work universally anyway, as it is only valid in src/fns as a signal that we want the actual length returned ...

I'll revisit all uses of ZeroOk and probably eliminate it and have an explicit check in src/fns functions for the output length behavior

ossl/src/pkey.rs
simo5: in my defense this is not a new thing, it was already like that before, and I need to figure out why now ...

simo5: Ok so I asked Gemini to summarize wihy wrapping_add here is appropriate:

## Why it is appropriate for `export_params_callback`

In `export_params_callback`, the function calculates `total_size` by summing the size of the `OSSL_PARAM` array structures plus the reported `p.data_size` of each parameter, aiming to establish an upper address boundary `max_ptr`:                                                                      
                                                                                                    
```rust 
let max_ptr = (params as *const u8).wrapping_add(total_size) as *const c_void;                      
```

`wrapping_add` is appropriate here for several key reasons:                                         
                                                                                                    
1. **Uncertainty of OpenSSL's memory allocation:**                                                  
   The function is guessing whether OpenSSL allocated the parameter array and parameter payload buffers in a single contiguous memory block. If OpenSSL allocated `params` in its own buffer while `p.data` buffers point to separate heap allocations (or if `total_size` exceeds the actual allocation of `params`), calculating `(params as *const u8).add(total_size)` would step outside the bounds of `params`'s allocated object, causing **immediate Undefined Behavior**.                                  

2. **The pointer is only used for bounds comparison, never dereferenced:**                          
   `max_ptr` is used strictly as a sentinel upper bound in the comparison:                          
   ```rust
   if pdata > base_ptr && pdata < max_ptr { 
       OPENSSL_cleanse(p.data, p.data_size);
   }
   ``` 
   Because `max_ptr` is never read from or written to, computing an address beyond the allocated object with `wrapping_add` is safe and sound.

3. **Prevents miscompilation from LLVM optimizer assumptions:**
   Using `add()` would permit LLVM to assume that any pointer comparison involving `max_ptr` only makes sense within the single allocation of `params`, potentially optimizing out or miscompiling the `pdata < max_ptr` range check. `wrapping_add()` disables this `inbounds` assumption.

simo5: sounds like this is not backwards at all as I initially thought, there fore nothing to fix.

simo5:
@Jakuje I think this is ok for re-review now

Jakuje (approved):

